PR #3, opened 2019-04-12
EN-520 : Address linting errors

antictor:
This PR not only fixes outstanding lint errors but also updates the scripts and pre-commit hook to better ensure they are caught before they are even committed.


packages/html-storybook/src/components/Button/Button.stories.js
```diff
@@ -1,64 +1,62 @@
...
-import { withKnobs, text } from '@storybook/addon-knobs'
 
 const stories = storiesOf('Button', module)
 
 stories.addDecorator(withKnobs)
 
 stories.add('Primary', () => (
   <div>
-    <button class="rn-btn--primary neutral large " type="button">
+    <button className="rn-btn--primary neutral large" type="button">
```
98dfsj: So I purposefully used `class` and not `className` here as these are examples of how to use the code in HTML not React. I get that it fails linting but we are either going to have to ignore this rule for this package OR add a transformer to allow for the use of `class` without issue.

antictor: Storybook accepts jsx, not html, so what do you propose?

antictor: Would it be better to put the examples in the docs markdown instead of having a htmlstorybook at all?

98dfsj (approved):

PR #4, opened 2019-04-12
EN-500 : Enable Jest tests in Circle

antictor:
A few tweeks and added Jest testing to Circle CI

Screenshot 2019-04-12 at 15 44 28

Makes use of workflows to only test things is they lint ok


98dfsj (approved):
Nice :)

PR #9, opened 2019-04-17
EN-501 : Auto publish NPM packages

antictor:
Publishes npm packages to npmjs.com upon merge to master.

Adds instructions to CircleCI config to publish npm packages only if the commit is made to master.
2 Environment settings are added to the CircleCI job to store the hostname that the npm registry sits at, so a private registry can be used for testing, and an authentication key with the ability to publish packages.

Tested with a forked copy of the repo and published to npm.royalnavy.dev

Screenshot 2019-04-17 at 19 03 46
Screenshot 2019-04-17 at 19 03 30

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -250,7 +250,11 @@ workflows:
             - test_react-component-library
             - test_storybook-react-input-state
             - test_vue-component-library
-      - publish_documentation
+      - publish_documentation:
+          filters:
+            branches:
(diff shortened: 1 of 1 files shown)


antictor:
Bug due to missing filter and account requiring 2 factor auth so will have to use an account with that turned off which is a bit of a pain

98dfsj (approved):

PR #10, opened 2019-04-29
EN-293 Feature/vue component transfer

98dfsj:
# Closes issue: en-293, en-291 and en-292

## Overview

Built transfer component.

## Work carried out

- [x] Built transfer component
- [x] Got tests to pass
- [x] Built storybook examples
- [x] Styled component
- [x] Wrote docs

## Screenshot

transfer


packages/css-framework/index.scss
```diff
@@ -59,7 +59,7 @@
 @import "src/components/progress";
 @import "src/components/sidenav";
 @import "src/components/table";
-@import "src/components/toggle";
+@import "src/components/transfer";
```
antictor, line 63: You killed toggle

packages/css-framework/src/components/_transfer.scss
```diff
@@ -0,0 +1,59 @@
+.rn-transfer {
+  .column-left,
+  .column-right {
+    width: 40%;
+    float: left;
```
antictor: flexbox?

packages/css-framework/src/components/_transfer.scss
```diff
@@ -0,0 +1,59 @@
...
+    border: 1px solid #e0e8ed;
+    color: #012159;
```
antictor: Colours should use the colour helpers to pick from the palette

packages/css-framework/src/components/_transfer.scss
```diff
@@ -0,0 +1,59 @@
...
+    label {
+      span {
+        display: block;
+        padding: 10px;
```
antictor: Are there common variables for spacing?

packages/css-framework/src/components/_transfer.scss
```diff
@@ -0,0 +1,59 @@
...
+    }
+  }
+  ul {
+    margin: 0;
+    list-style: none;
+    padding-left: 0;
+    height: 250px;
+    overflow-y: auto;
+  }
+  h1 {
```
antictor: Isnt this in the lib already? Also H1 is for a page, not a component, and there can be only 1.

packages/css-framework/src/components/_transfer.scss
```diff
@@ -0,0 +1,59 @@
...
+      }
+      input {
+        display: none;
+        &:checked + span {
+          background: #e9f8ff;
+        }
+      }
+    }
+  }
+  .transfer-buttons {
```
antictor, line 46: dont we have styles for buttons already

packages/vue-component-library/src/components/RnTransfer/scripts.js
```diff
@@ -0,0 +1,39 @@
...
+
+    rightList () {
+      return this.listData.filter(item => item.list === 'right')
+    }
+  },
+
+  methods: {
+    move (item, side) {
+      if (item.checked === true) {
+        // eslint-disable-next-line no-param-reassign
```
antictor: Didn't we talk about this before? I thought the agreement was no mutation of params

antictor (changes requested):

packages/css-framework/src/components/_transfer.scss
98dfsj: Have we documented that decision somewhere? I don't recall it being made.

packages/css-framework/src/components/_transfer.scss
98dfsj: These are not regular buttons.

packages/vue-component-library/src/components/RnTransfer/scripts.js
98dfsj: It feels like the most efficient way to do it though, the list data comes in like this:

```
    {
        "id": "stfaith",
        "location": 'St Faith', 
        "list" : 'left',
      },
```
and just changing `item.list` to `right` is a quick way that works well. ultimately what is returned is a mutated JSON object anyway so surely it doesn't really matter.

packages/css-framework/src/components/_transfer.scss
98dfsj: I seem to recall @hxltrhuxze telling me not to bother with variables as they are all about to change.

packages/css-framework/index.scss
98dfsj: Whoops! Sorry, I'll re-add it.

packages/css-framework/src/components/_transfer.scss
98dfsj: Float works perfectly well and is supported further back in time. I see no benefit to using flexbox here.

antictor: We have used Flexbox elsewhere I think, I'll let our designer css guru make the call :)

packages/css-framework/src/components/_transfer.scss
antictor: All the more reason, you use the built in colours and you get the update for free

packages/css-framework/src/components/_transfer.scss
antictor: It's normal (not gonna say best practice) to have a single H1 in a page to help screen reader navigate it and its role is to state the pages purpose. Also why would you put a H1 in a component?

packages/css-framework/src/components/_transfer.scss
antictor: Are the M&S buttons?

packages/vue-component-library/src/components/RnTransfer/scripts.js
antictor: Mutating data can have weird side effects, like causing infinite re-render. Can't we just return a new value?

packages/css-framework/src/components/_transfer.scss
98dfsj: Cool. I'll go with the flow.

packages/css-framework/src/components/_transfer.scss
```diff
@@ -0,0 +1,59 @@
...
+    width: 10%;
+    float: left;
+    margin-top: 90px;
+
+    .left, .right {
```
hxltrhuxze: Are there any individual styles for these classes? If not, we can refactor them into a generic `button` class.

packages/css-framework/src/components/_transfer.scss
98dfsj: Because html elements such as `article` and `section` (which I realise now, these are not - my bad, thought I'd made it a `section`) reset the document flow. I've always personally been of the opinion that every time you use one of these elements, you should start again with a `h1`, especially in components as you have no idea where in the document flow the component is going to be used and using a `h3` may not make sense. 

Again though, I'm not in charge so I'm happy to go with the flow here. I'd be interested in know what @hxltrhuxze and @xdobtqletrmn 's opinion on this is though.

packages/css-framework/src/components/_transfer.scss
```diff
@@ -0,0 +1,59 @@
...
+        &:checked + span {
+          background: #e9f8ff;
+        }
+      }
+    }
+  }
+  .transfer-buttons {
+    width: 10%;
+    float: left;
+    margin-top: 90px;
```
hxltrhuxze: Magic numbers get a bit funky when we go responsive. Can this be replaced by:

```
display: flex;
flex-direction: column;
align-items: center;
justify-content: center;
```
This'll keep the buttons centred no matter the component size

packages/css-framework/src/components/_transfer.scss
98dfsj: Damn right they are.

packages/css-framework/src/components/_transfer.scss
```diff
@@ -0,0 +1,59 @@
+.rn-transfer {
+  .column-left,
```
hxltrhuxze: Can `column-left` and `column-right` be replaced with one generic class?

packages/css-framework/src/components/_transfer.scss
```diff
@@ -0,0 +1,59 @@
...
+      input {
+        display: none;
+        &:checked + span {
+          background: #e9f8ff;
+        }
+      }
+    }
+  }
+  .transfer-buttons {
+    width: 10%;
```
hxltrhuxze: Can we make this a fixed width when we change the transfer to flexbox? Will mean that on smaller screen sizes the buttons don't get crushed

packages/css-framework/src/components/_transfer.scss
```diff
@@ -0,0 +1,59 @@
...
+        cursor: pointer;
+        background-color: #324E68;
+      }
+    }
+  }
+  ul {
+    margin: 0;
+    list-style: none;
+    padding-left: 0;
+    height: 250px;
```
hxltrhuxze: What dictates the `250px`? Is there a way we can make this more generic, so it's not tied directly to a pixel height?

hxltrhuxze (changes requested):

packages/css-framework/src/components/_transfer.scss
98dfsj: See, this is what comes of having a 3 week gap in between development. This used to be needed but I refactored it out in the previous sprint and totally forgot about it by the time I picked it up again in this one :-/

packages/css-framework/src/components/_transfer.scss
98dfsj: Whoops. Sorry this was left in from testing. Height has now been moved to storybook.

packages/css-framework/src/components/_transfer.scss
98dfsj: Fair. Done.

antictor (approved):

hxltrhuxze (approved):

PR #13, opened 2019-05-02
Add support for links to Vue Dropdown

antictor:
The previous version of dropdown only allowed the component to act like a field and emitted a value in an 'input' event. This updated version introduces the ability to pass in a list of options with a url for each one, so the dropdown is used as a navigation aid.

Also adds progressive enhancement so regular HTML pages can use dropdowns with links.

droplink


98dfsj (approved):
Really nice work on the docs.

PR #15, opened 2019-05-07
EN-299 Sortable table

98dfsj:
# Closes issue: EN-299, EN-298, EN-297

## Overview

Added sorting capabilities to the dynamic table component.

## Work carried out

- [x] Added sort_column method
- [x] Added styles
- [x] Added docs


packages/css-framework/src/components/_table.scss
```diff
@@ -22,4 +22,28 @@
     padding: spacing(3) spacing(2);
     color: color(neutral, 600);
   }
+
+  &__arrow {
+    margin-left: 10px;
```
antictor: This a standard margin unit ?

packages/vue-component-library/src/components/RnTable/scripts.js
```diff
@@ -4,5 +4,40 @@ export default {
...
+  }),
+  created () { 
+    this.tableDataMutable = this.tableData 
+  },
+  methods: {
+    sort_type (order, next) {
+      this.sortOrder = order
+      this.sortNext = next
+    },
+    sort_column (index) {
```
antictor, line 36: Maybe try and use a computed property instead of a function to return sorted, the computed property would do the same thing but follows the rule to compute derived state.

packages/css-framework/src/components/_table.scss
98dfsj: Whoops, well spotted.

antictor (approved):

PR #17, opened 2019-05-10
EN-583 Docsite to monorepo

98dfsj:
# Closes issue: EN-583

## Overview

Importes the current `royalnavy.io` repo into the monorepo as a package

## Reason

We wanted a version of the docs site which could be instantly tested without the need for adding symlinks or doing test deploys.

## Work carried out

- [x] Imported codebase into monorepo
- [x] Added references to the package.json file
- [x] Fixed broken Swatch component
- [x] Fixed failing tests
- [x] Updated docs to correct incorrectly referenced `swatch` component.


m7kvqbe1 (approved):
Looks good to me - npm scripts all playing nice and build successful.

PR #19, opened 2019-05-14
Docsite deployment automation (Prod & Staging)

m7kvqbe1:
## Overview

Automates the deployment of the docs site for master and develop branches to production and staging Firebase environments respectively.

## Link to preview

I emulated a staging deployment to test the CircleCI config by temporarily adding the development branch name to the workflow filters:

https://circleci.com/workflow-run/8f3c5821-3a28-435b-a790-ef55ea5b5da9

## Reason

>When new content is pushed to the `standards-toolkit` repo, trigger a branch deployment of docs.royalnavy.io so the new documentation can be tested in situ.

## Work carried out

- [x] Generate a firebase auth token for CI and expose
- [x] Add FIREBASE_AUTH_TOKEN env var to CircleCI via admin console
- [x] Automate production deployment for master
- [x] Automate staging deployment for develop
- [x] Manual deployments can be made to pre-defined environments via project aliases 

## Screenshot

Screenshot 2019-05-14 at 14 00 38

## Developer Notes

Further implementation has been researched that will enable branch deployments. This will allow us to carry out UAT on feature branches and bug fixes with ease.

See Jira ticket for detailed information about upcoming implementation:

https://rn-nelson.atlassian.net/browse/EN-576


packages/docs-site/.firebaserc
```diff
@@ -1,21 +1,22 @@
 {
   "projects": {
-    "docs-site": "standards"
+    "production": "standards",
+    "staging": "standards-staging"
   },
   "targets": {
-    "docs-site": {
+    "standards": {
```
98dfsj, line 7: Not sure `standards` is actually the right name here. I know it's picky but we use `standards` as the name of the project in its entirety. `docs-site` probably is the right name, plus it matches the package name in the repo.

m7kvqbe1: This is because i'm targeting the `standards` project within Firebase. Under each of these projects (`standards` and `standards-staging`) there are multiple sites deployed:

docs-site
react-storybook
vue-storybook
...

98dfsj: ah, ignore me then :)

98dfsj (approved):

m7kvqbe1:
You can access the docs-site staging deployment here for now:

https://standards-staging.firebaseapp.com

We could use something else if we update the royalnavy.io DNS.

.circleci/config.yml
```diff
@@ -226,6 +229,46 @@ jobs:
...
+      - dependencies
+      - attach_workspace:
+          at: ~/standards-toolkit
+      - run:
+          name: Build site
+          working_directory: packages/docs-site
+          command: npm run build
+      - persist_to_workspace:
+          root: .
+          paths:
```
antictor, line 246: Will this approach also cause all the node_modules to get persisted?

antictor (approved):
All looks fine to me. My only question is about the use of workspaces. Wondering if there is a way to further reduce how much is persisted to a workspace to speed up the time taken to re-attach those files, and would there ever be any issue with the workspace root being the same as the project root?

.circleci/config.yml
m7kvqbe1: It will only persist the paths specified below:

```
- packages/docs-site/public
- packages/docs-site/firebase.json
- packages/docs-site/.firebaserc
```

I'm only carrying over the build artefacts for the two deployment jobs.

I guess you could use it to persist `node_modules` if you wanted but there is cacheing for that?

m7kvqbe1: It's my understand that it only persists the paths specified:

https://circleci.com/docs/2.0/configuration-reference/#example-for-root-key

m7kvqbe1: re: workspace root question

When learning about workspaces, I used an official CircleCI example of workspace-forwarding and referenced that for best practice:

https://github.com/CircleCI-Public/circleci-demo-workflows/blob/workspace-forwarding/.circleci/config.yml

PR #23, opened 2019-05-15
Update pull_request_template.md

m7kvqbe1:
Fixes minor typo in the pull request template.


98dfsj (commented):
Since we are are doing this, the first line should be amended too as it still mentions GH issues which we don't use anymore, plus we no longer 'close' issues with PR's we're merely putting them into ready for release so we're probably better off making the like something like:

`## Related Jira ticket:` or something.

98dfsj (approved):

PR #24, opened 2019-05-15
EN-598 Update colour palette

antictor:
This change updates the colour palette in the css framework to the new colours agreed by design.

In addition some changes have been made to components that used shades no longer supported.

While working on testing the components to ensure they render correctly I discovered that buttons do not use the correct colours currently and have a follow up branch being worked on that corrects this by changing how their colours are calculated.

Some changes are also as a result of running prettier over code as part of auto save.

Finally, addressed a warning from the sass compiler with a change to how globals are defined.

### DIFF
--- a/packages/css-framework/src/contexts/_light.scss
+++ b/packages/css-framework/src/contexts/_light.scss
@@ -26,7 +26,7 @@
       breakpoint: 1600px,
       baseFontSize: 118%,
     ),
-  ) !global;
+  );
 
   // Set the context colours
   // Todo - Review colour uses and remove neutral 00
@@ -129,7 +129,7 @@
       800: #245c40,
       900: #1f4a35,
     ),
-  ) !global;
+  );
 
   // State Mapping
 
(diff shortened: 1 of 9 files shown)


packages/css-framework/src/contexts/_light.scss
```diff
@@ -1,190 +1,237 @@
...
+      200: #fefbb8,
+      300: #faed7e,
+      400: #f5db54,
+      500: #e8c242,
+      600: #cf9328,
+      700: #ae6d1d,
+      800: #8c4f17,
+      900: #693a12,
+    ),
+    altwarning: (
```
hxltrhuxze, line 95: Is `altwarning` the best use for purple? I don't have a good suggestion for another name though I'm afraid...

hxltrhuxze (approved):

packages/css-framework/src/contexts/_light.scss
antictor: I know. no idea 🤷‍♂

antictor: info maybe.. or Deep !

PR #26, opened 2019-05-16
Docs consolidation

98dfsj:
## Overview

Adds a documentation consolidation script to combine docs from all over the monorepo together. 

## Reason

As a developer I want to document my React JS component by creating a README.md file in the component source folder, whilst following a convention for the document structure.

## Screenshot

Screenshot 2019-05-17 15 42 33

### Exclusions
If a components documentation should have a package excluded for any reason, it can be done by adding a parameter to 'framework-tabs`. 

e.g.

```html
<framework-tabs exclude="react"></framework-tabs>
```
Multiple packages can be excluded by turning the exclude parameter into a comma separated list.

## Conventions

In order for the consolidator to process a file, the title field in the front matter should exactly match in both the documentation page and the component readme. It is case insensitive though.


antictor (commented):
Why are there files in documentation/library/pages/components/* ?
I thought this would now come from folders like packages/react-component-library/src/components/alert/README.md?

packages/documentation/library/pages/components/nav.md
```diff
@@ -0,0 +1,12 @@
+---
+title: Nav
+description: Documentation and examples of how to use NELSON navs
+---
+
+# Overview
+
+<framework-tabs></framework-tabs>
+
+#### Storybook
```
antictor, line 10: also we wont be using storybook in the docs, it will be up to the developer to create the documentation and storybook will purely be used for developers to test the code.

packages/vue-component-library/src/components/RnTransfer/README.md
```diff
@@ -0,0 +1,42 @@
+---
```
antictor, line 1: Just noticed that you do have docs here, but why do we have docs here And the document folder?

antictor:
Another thought, outside the scope here, how are we going to include html examples in a page? We'd have to load the css but then thats no good for a github readme as it can't pull in css. So we'll have to allow images, can we enhance this to pull them in too?

packages/documentation/library/pages/components/nav.md
98dfsj: The files left in component are the generic component docs. The docs from each individual component are merged in and injected into the 'framework-tabs' tag. This allows for extra information that isn't framework specific.

The content of the generic component docs is just what was there before. It should be updated in a different ticket/pr.

98dfsj:
> Another thought, outside the scope here, how are we going to include html examples in a page? We'd have to load the css but then thats no good for a github readme as it can't pull in css. So we'll have to allow images, can we enhance this to pull them in too?

I think that's probably a good point and definitely something we should do but that feels like it's an upgrade to this and should be a separate ticket.

antictor (commented):
Does this work for React or just Vue?

98dfsj:
Everything. You specify the packages you want including in package.json. If you read the main README.md file you'll see instructions for how it is used.

antictor (approved):

PR #27, opened 2019-05-17
Feature/add nesting and index sidebar

m7kvqbe1:
## Related Jira issue: 

https://rn-nelson.atlassian.net/browse/EN-585

## Overview

Implements nested navigation and indexed pages based on user defined values.

## Reason

>As a user I want to be able to view the documentation site and navigate a skeletal structure that describes the navigation and key pages described in the documentation site design

## Work carried out

- [x] Nest the flat page structure returned by GraphQL based on URL structure
- [x] Index pages based on user defined values specified in documentation frontmatter
- [x] Recursively render a nested list of hyperlinks allowing end user to navigate sitemap
- [x] Refactor code base to utilise containers (data), enhancers (HoCs) and presenters (view)
- [x] Add some snapshot tests
- [x] General refactoring of code base

## Screenshot

Screenshot 2019-05-17 at 14 28 43

## Developer notes

Medium to long term there has got to be a simpler way of doing this. This seems like something Gatsby should just be handling out of the box?


packages/documentation/library/pages/about-the-design-system/index.md
```diff
@@ -4,6 +4,7 @@ description: ''
 tags: public
 pageClass: ''
 template: default
+index: 6
```
98dfsj, line 7: It would be better if we could remove the dependencies on adding stuff to the markdown files - Ideally we'd actually limit the markdown frontmatter to `title`, `description` and `template`. If this absolutely has to be here then I suggest you at least name it something that's non-developer friendly like `navPosition` or something a bit more descriptive.

packages/docs-site/package.json
```diff
@@ -57,14 +57,15 @@
...
+    "sass": "^1.20.1",
     "source-code-pro": "^2.30.1",
     "styled-components": "^4.1.3",
     "stylelint": "^9.10.1",
     "stylelint-junit-formatter": "^0.2.1",
     "uuid": "^3.3.2"
   },
   "resolutions": {
-    "firebase-tools/**/lodash" : "4.17.11"
+    "firebase-tools/**/lodash": "4.17.11"
```
98dfsj: Did adding lodash explicitly actually fix the resolution issue?

m7kvqbe1: This change happened automatically when I did a `yarn add lodash`.

I'm using lodash in that sidebar component - I use the sortBy helper.

packages/documentation/library/pages/about-the-design-system/index.md
m7kvqbe1: Without some form of draggable UI that lets you order pages (like in a CMS) i'm not sure how else you would define this? 

Open to suggestions.

The index is relative to other pages within the same directory.

packages/docs-site/src/components/sidebar/index.js
```diff
@@ -28,27 +105,12 @@ const Sidebar = () => (
...
-                  <a href={slug}>
-                    <span>{title}</span>
-                  </a>
-                </li>
-              )
-            })}
-          </ul>
-          <pre>{JSON.stringify(data, undefined, 2)}</pre>
+          {renderList(nested)}
+          <pre>{JSON.stringify(nested, undefined, 2)}</pre>
```
m7kvqbe1:
```suggestion
          <pre>{JSON.stringify(nested, null, 2)}</pre>
```

packages/documentation/library/pages/about-the-design-system/index.md
m7kvqbe1: An alternative might be to add a numerical index to the start of the filename:

```
  ├── directory
  │   └── 1_example-page.md
  ├── another-directory
  │   ├── 1_example-page.md
  │   └── 2_another-example-page.md
```

m7kvqbe1:
Based on my research there appears to be two common ways of generating site navigation in Gatsby:

1. Implementation along the same lines as this pull request, where we query for all of the pages and then generate a representative data structure from the flat collection returned by GraphQL:

https://bilianavaleva.com/generate-navigation-with-gatsby/

2. Hard coding menu links as meta data within the `gatsby-config.js` and querying for this:

```
module.exports = {
  siteMetadata: {
    title: 'Gatsby Default Starter',
+    menuLinks:[
+      {
+         name:'home',
+         link:'/'
+      },
+      {
+         name:'page2',
+         link:'/page-2'
+      }
+    ]
  },
  plugins: []
}
```

https://www.gatsbyjs.org/docs/centralizing-your-sites-navigation/

---

My preference is option one - I think hard coding navigation and it needing to be updated whenever a change is made to the structure of the corresponding files in the documentation package is an unnecessary dependency and asking for trouble (issues with dead links come to mind).

If we are all in agreement that option one is indeed the way to go - I propose refactoring this initial implementation into a HoC that generates the nested data structure and then sidebar and navigation presentational components that consume and recursively render said data structure.

98dfsj:
I agree, option one all the way.

packages/documentation/library/pages/about-the-design-system/index.md
antictor: I like the meta tag approach. Don't like depending on file names or folders to get the order right. Limits us when we want to pull in component documentation from the component folder

98dfsj (approved):

PR #29, opened 2019-05-20
EN-599 Buttons

antictor:
## Overview

The new generation button, faster, better, more springy. A new React JS button based on current designs and those used in the docs website design. There is a choice of different button styles and basic icons can be applied.

button-variants

button-size

button-icon

Documentation for using the buttons can be found in the README.md file in the component folder.


hxltrhuxze:
I’ve just had a glance at this button PR quickly. Couple of points:
- The primary/secondary/tertiary buttons are ways of distinguishing hierarchy between buttons. Generally there should only be one primary or “main action” per piece of UI. The Primary button can appear blue (as normal), as red (danger), or as Green (Success). They don’t exist alongside the colour variations. You can (and should) have a primary button that is red, signifying it is the main action for a piece of UI and clicking it has destructive effects.
- The spacing and font size seem off on the buttons. Particularly with the regular and large buttons (see attached image), the regular button text is too close to the top, whilst the large button text is too close to the bottom.
- Can we also reduce the font-weight of the text? They should be set to medium weight, but it looks semibold from here
- We don’t have outline buttons, particularly with a 2px border. Can these be removed?

antictor:
The outline buttons idea originated from the designs for the docs site, plus I noticed them in bootstrap so thought them useful and would adapt it with an icon to get the desired effect.

Screenshot 2019-05-20 at 15 47 16

antictor:
With the button colours, agree that you'd only have a red or green. The examples are just to demonstrate the colours. The guide on how to use them would appear in the intro and design part of the docs. What do you think I should change within this code?

With button spacing, would appreciate your eye on the code and suggest how to get it right, using line heights or padding

antictor:
Had a long chat with designers, changing how buttons are structured so will close this until the next version is ready

antictor:
Updated button styles after discussion with design.

hxltrhuxze (approved):

PR #30, opened 2019-05-21
EN-592: React component phase banner

98dfsj:
## Overview

Created phase banner

## Work carried out

- [x] Built phase banner
- [x] Created stories
- [x] Added tests
- [x] Added docs

## Screenshot

Screenshot 2019-05-21 12 26 10


packages/css-framework/src/components/_phase-banner.scss
```diff
@@ -0,0 +1,25 @@
+.rn-phase-banner {
+  display: block;
+  background-color: color(primary, 100);
+  padding: spacing(2) spacing(6);
+  &--alpha {
```
hxltrhuxze: are these needed? There's `display: block;` already on the `.rn-phase-banner`

packages/css-framework/src/components/_phase-banner.scss
```diff
@@ -0,0 +1,25 @@
...
+  &__text {
+    display: inline-block;
+    margin-left: spacing(1);
+    font-size: font-size(xs);
+    a {
+      color: color(neutral, black);
+    }
+  }
+
+  .rn-badge--solid {
```
hxltrhuxze: Can we vertically align the text with the badge? At the moment the "This is a new service..." text is higher than the alpha button. Ideally the text and badge are both vertically centred in the banner

packages/css-framework/src/components/_phase-banner.scss
```diff
@@ -0,0 +1,25 @@
+.rn-phase-banner {
+  display: block;
+  background-color: color(primary, 100);
+  padding: spacing(2) spacing(6);
```
hxltrhuxze: Do we want to place the content of the banner inside a `.container` class? This way it'll scale and stay horizontally in line with the rest of the content

packages/react-component-library/src/components/PhaseBanner/PhaseBanner.test.js
```diff
@@ -0,0 +1,27 @@
...
+import { shallow } from 'enzyme'
+
+import PhaseBanner from '.'
+
+describe('PhaseBanner', () => {
+  let onChangeStub
+  let wrapper
+
+  beforeAll(() => {
+    onChangeStub = jest.fn()
```
m7kvqbe1: Is this being used?

98dfsj: Ah... no it isn't. Well spotted :)

packages/css-framework/src/components/_phase-banner.scss
98dfsj: Whoops. That was left over from earlier. That's what happens when you start something one day and finish it the other.

packages/css-framework/src/components/_phase-banner.scss
98dfsj: Can you explain what you mean? Where should the `.container` class be? Does it require an extra DOM element and can you point me to the file I should be looking it which describes what the `container` class does so I know when to use it in the future?

packages/css-framework/src/components/_phase-banner.scss
```diff
@@ -0,0 +1,17 @@
+.rn-phase-banner {
+  background-color: color(primary, 100);
+  padding: spacing(2) spacing(6);
+  
+  &__text {
```
m7kvqbe1: I personally wouldn't nest selectors like this - my main reason is searching for selectors in an IDE.

I like to write out the full selector where possible (different for pseudo selectors).

Have we agreed some best practices around writing CSS and are they documented somewhere? Are we linting SASS?

packages/css-framework/src/components/_phase-banner.scss
hxltrhuxze: The container class sits around the content, so in this case it would be:
```
<div className="rn-phase-banner">
    <div class="container">
        content here
    </div>
</div>
```

There should be a site-wide container element. If it doesn't exist, it needs to be created. It will provide horizontal padding for all the elements in the site, and will assert a max width on the content if the browser window is opened to a larger size.

packages/css-framework/src/components/_phase-banner.scss
antictor: I agree. When using BEM it's generally a good idea to spell out the whole selector and I'd even go as far as not nesting  (see my Button PR). BEM does away with the need to nest and that then makes it easier to search for a selector in your IDE. You often realise you don't need the whole selector all and maybe just need `.thing__text` instead of `.thing.thing__text` which makes it easier for the browser to figure out the rule to use. We are linting sass but not looked at the config in a while

98dfsj: I'm totally fine with that but I'd like to get @hxltrhuxze 's opinion on this matter before I change it.

hxltrhuxze: Totally fine with this naming convention. Whilst I like scoping stuff to use the `&__class` helper, I can see the advantages of spelling out the whole class name each time.

98dfsj: Works for me.

98dfsj:
Don't approve this. I forgot to write the docs.

98dfsj:
Docs added. Can be reviewed again.

packages/css-framework/src/components/_phase-banner.scss
```diff
@@ -0,0 +1,20 @@
...
+  padding: spacing(2) 0;
+}
+
+.rn-phase-banner__text {
+  display: inline-block;
+  margin-left: spacing(1);
+  font-size: font-size(xs);
+}
+
+.rn-phase-banner__text a {
```
antictor, line 12: do we need to style the 'a' or can it be a class?

packages/css-framework/src/objects/_wrapper.scss
```diff
@@ -1 +1,9 @@
-// Wrapper Object
\ No newline at end of file
+.container {
+  max-width: 1440px;
```
antictor: Can this be a variable?

packages/css-framework/src/objects/_wrapper.scss
```diff
@@ -1 +1,9 @@
...
+  padding: 0 spacing(4);
+  margin: 0 auto;
+  @media only screen and (min-width: 768px) {
```
antictor, line 4: Isn't there a mixing or something for this?

packages/css-framework/src/objects/_wrapper.scss
```diff
@@ -1 +1,9 @@
...
+    padding: 0 spacing(8)
```
antictor, line 5: standard spacing unit?

packages/react-component-library/src/components/PhaseBanner/index.js
```diff
@@ -0,0 +1,33 @@
...
+PhaseBanner.propTypes = {
+  /** The phase the banner is in */
+  phase: PropTypes.string,
+  /** The html to display with the banner */
+  html: PropTypes.string,
+}
+
+PhaseBanner.defaultProps = {
+  phase: 'alpha',
+  html:
```
antictor: How about using 'children' instead of a html property?

packages/react-component-library/src/components/PhaseBanner/index.js
```diff
@@ -0,0 +1,33 @@
...
+
+const PhaseBanner = ({ phase, html }) => (
+  <div className={`rn-phase-banner rn-phase-banner--${phase}`}>
+    <div className="container">
+      <Badge size="small" state="primary" type="solid">
+        {phase}
+      </Badge>
+      <span
+        className="rn-phase-banner__text"
+        dangerouslySetInnerHTML={{ __html: html }}
```
antictor: dangerous by name..... This is why I'd rather use children, you can still default it.

packages/react-component-library/src/components/PhaseBanner/README.md
```diff
@@ -0,0 +1,30 @@
...
+description: A simple banner to indicate the phase of the project.
+---
+
+## Usage
+
+The banner should be used directly underneath the main site header, before the content begins. If used with no props then it will look exactly like the example image below, however it can be customised to with any text that is required.
+
+## Examples
+
+<img src="images/example.gif" width="664" />
```
antictor: I wasn't sure how we'd copy the images in the gatsby scripts so I always prefixed my images with the component name just in case.

packages/css-framework/src/objects/_wrapper.scss
98dfsj: I've actually just removed this as the max width is for the component lib contain and isn't needed in the docs site apparently.

packages/css-framework/src/objects/_wrapper.scss
98dfsj: It doesn't currently work. @hxltrhuxze is planning to recreate it soon.

packages/css-framework/src/objects/_wrapper.scss
98dfsj: Please clarify what you mean.

packages/react-component-library/src/components/PhaseBanner/README.md
98dfsj: good call.

packages/css-framework/src/components/_phase-banner.scss
98dfsj: We need to style the a really. It's potentially user generated content.

packages/css-framework/src/objects/_wrapper.scss
antictor: hmm. ok, though I've used it for Nav. What doesn't work? Maybe leave a Todo in here so we don't forget

packages/css-framework/src/objects/_wrapper.scss
antictor: nvm.. mis-read

antictor (approved):

PR #33, opened 2019-05-25
React Typescript component library

antictor:
Hear me out.....

So we are reviewing the entire React component library to meet the new designs and as part of that we need to go through each and every component to update along with the associated styles to follow BEM.

Typescript makes the most sense for libraries as it allows the production of type definitions for developers building apps using Typescript and provides way better autocomplete & type checking for regular JS projects edited with Visual Studio code. Fuel would benefit as it uses Typescript. Add to this Typescript is really easy for components, it doesn't get hard until you get into application logic.

If we want to move to Typescript then now is the best time to do this, otherwise if we did it later it would mean re-writing code again.

I propose we create a new package in which we create all the new components and use Typescript. Once the list is complete we delete the old one and rename this one to replace it.

I've migrated the components we've updated so far plus the `Badge` component as it used used by  `PhaseBanner`.

Just like before the package uses StoryBook, Jest and ESLint (TSLint is being dropped for ESLint by MS). The build scripts use Webpack in combination with Dart Sass loader and Babel. Babel is faster and easier then AwesomeTypescriptLoader and the native Typescript compiler. Babel can determine the syntax and polyfills in the final JS as it uses babel-preset-env to define the target browsers and understand their limitations.

I've also created a repo with the base project I created to use for this package so it can be used for other projects if required.

I can talk everyone through how the build process works if need be.


98dfsj (approved):

98dfsj:
Definitely agree with maintaining this separately for the time being. It's something that needs to be done but I think it would be useful for us all to have a bit of a crash-course in TS before we go 'all in'

m7kvqbe1:
It looks like it's mainly just the addition of defining interfaces for functional components (within the context of a component library like this).

Are there any other key bits I'm missing?

m7kvqbe1 (approved):

antictor:
It's mostly defining types with interfaces, not prop types, putting defaults in function declarations and taking advantage of being able to define the valid strings for things like size. The side effect is you also get warnings on things that raw ES doesn't always see. I used to mock TS when it was impossible to use with Redux but with the component library and GraphQL libraries it shines.

PR #40, opened 2019-05-29
EN-603 Footer

antictor:
Footer component taken from designs in the docs site and the Jira ticket.

Uses the `Links` component just added. By default produces the footer example shown in the ticket but also allows links to be overridden the same way as `Nav` and the main footer message can be overridden by placing content within the tag `<Footer>This is some content</Footer>`

Did cross browser testing on Browserstack to check Chrome 38, only issue I saw was that the copyright message falls below the main message on IE 10 due to issues with flexbox but I figured this is acceptable.

**Desktop**
footer

**Mobile**
footermobile

**Chrome 38**
Screenshot 2019-05-28 at 17 47 51


packages/css-framework/src/components/_footer.scss
```diff
@@ -0,0 +1,44 @@
...
+  line-height: 1.5;
+  margin: spacing(1) 0;
+}
+
+.rn-footer__copyright {
+  white-space: nowrap;
+}
+
+.rn-footer__divider {
+  border:0;
```
m7kvqbe1, line 42:
```suggestion
  border: 0;
```

m7kvqbe1 (dismissed):
Looks lovely to me - looking forward to implementing some components with TS.

Found one tiny bit of missing whitespace (see comment with suggestion).

m7kvqbe1:
The only other thing i'll say is maybe it could do with a couple of simple tests? Just a snapshot and then something to assert that when given certain props you get the expected output.

antictor:
Gonna withdraw the PR after all as I need to move this to the doc site and add tests

PR #42, opened 2019-05-29
WIP: Feature/restructure css framework

m7kvqbe1:
## Related Jira issue:

https://rn-nelson.atlassian.net/browse/EN-624

## Overview

Restructure the CSS Framework in order to allow for different types of exports and the bundling of styles into their respective React & Vue implementations.

## Reason

> Currently component library styles can be imported into an application in an 'all or nothing' fashion. This creates large builds because the styles for every component need to be downloaded regardless of what components are being utilised.

## Work carried out

- [x] An application can now include individual components from their component library of choice, with styles bundled for only the components being utilised.

- [x] An application can also include the CSS library and helpers independent of component styles (allowing the utilisation of the CSS framework in the development of app centric components and styling).

- [x] An application can include the framework in its entirety without any JavaScript component implementation.

- [x] Make Jest stub file and style imports in both the React and Vue component libraries.

## Developer notes

See new structure definition below:

- /

  - index.scss - everything [including component styles]

- /src/library    

  - index.scss - variables, functions, resets...

- /src/helpers

  - index.scss - helpers and objects

- /src/components (these each depend on /src/library/index.scss)

  - component_name.scss

  - …

---

I think the Normalize.css resets might be being duplicated within each components bundled styles. Maybe Normalize should be an app dependency or at least excluded from the component SASS stack?


98dfsj:
Have you checked that the `library.scss` file is only actually added to the bundle once?

packages/css-framework/index.scss
```diff
@@ -1,90 +1,8 @@
...
-@import "src/abstracts/mixins/cover";
-@import "src/abstracts/mixins/hover";
-@import "src/abstracts/mixins/pseudo";
-@import "src/abstracts/mixins/shadows";
-@import "src/abstracts/mixins/truncate";
-@import "src/abstracts/mixins/typography";
-@import "src/abstracts/mixins/zindex";
-
-// Base
-@import "src/base/normalize";
```
hxltrhuxze, line 44: Agree re: normalize being imported into every instance. Every other file in `library` doesn't output any CSS. `normalize`, `box-sizing`, and `fonts` are the only files in this instance that output CSS, therefore meaning they'll be duplicated with every import. I don't know how best to tackle this unfortunately.

98dfsj: It really feels like there should be something in webpack that can handle this.

m7kvqbe1:
> Have you checked that the `library.scss` file is only actually added to the bundle once?

Yeah - see developer notes. The problem is if there is output from library (which is Normalize.css).

m7kvqbe1:
We should be careful to ensure that when you consume the components you don't get a bleed of styles into the rest of the project (from things like resets).

hxltrhuxze:
We could potentially wrap each component in a parent component that contains all global styles

https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/CssBaseline/CssBaseline.js

PR #44, opened 2019-05-30
Feature/nested navigation

m7kvqbe1:
## Related Jira issue: 

https://rn-nelson.atlassian.net/browse/EN-596

## Overview

Upgrade the Nav component within the React component library to support nested items.

## Reason

The docs-site project requires nested vertical navigation.

## Work carried out

- [x] Upgrade implementation to support nesting
- [x] Add component tests

## Screenshot

Screenshot 2019-05-30 at 15 11 51

## Developer notes

- Kept styling to a minimum for nested items (no component library designs to inform this yet).
- Added classes like `has-children` in order to allow custom styles to be easily implemented/extended where the component is consumed (for example the docs-site).
- Supports nesting as deep as the collection provided via recursion.
- Only vertical navigation will display nested items at this point in time.


packages/react-component-library/src/components/Nav/Nav.test.tsx
```diff
@@ -0,0 +1,122 @@
...
+        },
+        {
+          href: 'http://testurl.test',
+          label: 'Child 2',
+        },
+      ]
+    },
+  ]
+
+  describe('Given the nav is supplied navItems prop', () => {
```
antictor, line 54: We can talk more about agreed test methods later. But I'd be tempted to not just test the render but inspect what is rendered in individual tests

```
describe('Nav' => () ={
  describe('Given the nav is supplied a flat navItems collection', () => {
    let navItems

    beforeEach(() => {
      navItems = ....
    })

    describe('and is asked to render vertically', () => {
      let wrapper

      beforeEach(() => {
        wrapper =shallow(<Nav navItems={navItems} orientation="vertical"></Nav>)
      })

     it('should indicate that the navigation is vertical', () => {
       expect(wrapper.class).toInclude('vertical')';
     })

    it('should render 5 list items', () => {
...
})
```

That idea being that you nest tests and build up contexts and then each it test just 1 thing, or one aspect.

packages/react-component-library/package.json
```diff
@@ -48,6 +48,7 @@
     "@types/node": "^12.0.2",
     "@types/react": "^16.8.18",
     "@types/react-dom": "^16.8.4",
+    "@types/react-test-renderer": "^16.8.1",
```
antictor: what about enzyme?

m7kvqbe1: Changed to enzyme-to-json for snapshot serialisation.

packages/react-component-library/src/components/Nav/Nav.test.tsx
m7kvqbe1: Sounds good to me I like this style.

m7kvqbe1: I'm currently checking the appropriate change to the rendered output happens for the different potential variants of each prop.

Then i'd do the same for user input (if it was an interactive component) and ensure that appropriate events are emitted or any expected side effects are initiated.

antictor (approved):

PR #45, opened 2019-05-30
Exclude tests from Code Climate checks

m7kvqbe1:
Need to exclude tests from the Code Climate checks.

Had to manually add an exclusion for ours as the default doesn't catch them.

Also excluded vendor stuff and release artefacts (would have been handled by defaults previously).

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,0 +1,4 @@
+exclude_patterns:
+- "dist/"
+- "**/node_modules/"
+- "*.test.tsx"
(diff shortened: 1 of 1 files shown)


.codeclimate.yml
```diff
@@ -0,0 +1,4 @@
+exclude_patterns:
+- "dist/"
+- "**/node_modules/"
+- "*.test.tsx"
```
98dfsj: For now, it's probably worth also including .js files as the docs-site isn't ts yet.

98dfsj (approved):

antictor (approved):

PR #57, opened 2019-06-03
EN-619 TextInput

antictor:
A TextInput component that lets you create form fields
Screenshot 2019-06-03 at 18 38 52

Simulator Screen Shot - iPhone Xʀ - 2019-06-03 at 18 41 43

The TextInput is a regular text input field that also shows the field label and handles adornments.

There are also some small improvements to buttons to handle spacing when used in forms and to allow a form to be submitted.


antictor:
Chrome 38 :)

Screenshot 2019-06-04 at 08 49 11

hxltrhuxze:
Can we check the spacing of the inputs please? They're a bit off compared to the designs
Screenshot 2019-06-04 at 10 35 00

antictor:
Screenshot 2019-06-04 at 13 36 28

Updated styles

hxltrhuxze:
can we move the label down `2px`?

antictor:
I can, but I wanted to keep to the standard spacing units

hxltrhuxze (approved):

09txzS1qs:
Is this alignment right? @hxltrhuxze @antictor to me it should be further left and comfortably above the magnifying glass icon. 
text-input

hxltrhuxze:
There shouldn't be a left aligned icon inside the input imo. If we need a left aligned icon, it should be in a pre-fix.

PR #66, opened 2019-06-10
Adjust the PhaseBanner to go full screen width with content in container

antictor:
The phase banner was designed to sit inside a container but wouldn't go full screen properly.

Design want the phase to go full screen width but the content is kept inside the content area.

Screenshot 2019-06-10 at 14 48 58

Simulator Screen Shot - iPhone Xʀ - 2019-06-10 at 14 50 29

Simulator Screen Shot - iPad Air (3rd generation) - 2019-06-10 at 14 53 53

PR #67, opened 2019-06-10
EN-588 Masthead

antictor:
Masthead component designed to work on mobile and desktop.

Note that links are hard coded until we start to get content in.

Input component is embedded into the docs site until we figure out why we can't bring in components that use hooks.

Screenshot 2019-06-10 at 17 05 27
Screenshot 2019-06-10 at 17 06 01

Screenshot 2019-06-10 at 17 05 33

Screenshot 2019-06-10 at 17 06 48


hxltrhuxze:
can we give the NELSON logo a bit more vertical padding on mobile? It looks quite cramped and the input is overpowering it.

antictor:
Screenshot 2019-06-11 at 09 18 08

packages/docs-site/src/components/presenters/Masthead/index.js
```diff
@@ -0,0 +1,93 @@
...
+
+  const onSubmit = term => {
+    console.log('Do a thing:', term)
+  }
+
+  const toggle = () => {
+    setOpen(!open)
+  }
+
+  const navItems = [
```
m7kvqbe1, line 30: There is an enhancer that will give you this data. Something for later though - next sprint.

packages/docs-site/package.json
```diff
@@ -115,5 +115,8 @@
       "name": "HTML",
       "componentPath": "/src/components"
     }
-  ]
+  ],
+  "devDependencies": {
+    "formik": "^1.5.7"
```
m7kvqbe1: I'm yet to use Formik - any recommended resources? I'll have a quick look through the docs today.

packages/docs-site/src/components/presenters/Masthead/masthead.scss
```diff
@@ -0,0 +1,70 @@
...
+    margin: 0;
+  }
+
+  .masthead__search .rn-btn {
+    display: none;
+  }
+
+  .masthead .rn-nav {
+    display: block;
+    margin-left: -1rem;
```
m7kvqbe1, line 63: Do we definitely need to use a negative margin?

m7kvqbe1 (approved):
Looks great - just a few tiny comments / questions.

packages/docs-site/package.json
antictor: TBH their docs are good enough to learn it. Be aware there's a new version coming soon with Hook support

packages/docs-site/src/components/presenters/Masthead/masthead.scss
antictor: yes, otherwise the buttons look mis-aligned. Was trying to use the standard content width variable and to do that you need to do this. I'm not 100% happy and may revisit later.

09txzS1qs (approved):
We still need to update design of phase banner as it's displaying with errors highlighted in #66.

Aside from that looks good.

PR #70, opened 2019-06-11
EN-655 Doc site layout

antictor:
Updates the docs site layout to correctly position the masthead, body and footer. Allows for the inclusion of a left and right navs. Currently they are simple included, as the site progresses the left nav component will return null when there is no nav and the layout will adjust for that.

Left nav is not shown in mobile view as it's an enhancement and a user can still achieve their goal without it.

Screenshot 2019-06-11 at 15 28 33

Mobile

<hr/>

Screenshot 2019-06-11 at 15 29 32

iPad with nav 

<hr/>

Screenshot 2019-06-11 at 15 29 41
iPad no left nav

<hr/>

Screenshot 2019-06-11 at 15 33 26
Desktop nav

<hr/>

Screenshot 2019-06-11 at 15 33 36
Desktop no nav

### DIFF
--- a/packages/css-framework/src/contexts/_light.scss
+++ b/packages/css-framework/src/contexts/_light.scss
@@ -14,6 +14,10 @@
       breakpoint: 768px,
       baseFontSize: 106%,
     ),
+    l: (
+      breakpoint: 1024px,
+      baseFontSize: 106%,
+    ),
     xl: (
       breakpoint: 1200px,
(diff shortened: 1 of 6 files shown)


m7kvqbe1 (approved):
Looks good to me. 

The Sidebar styles look like they could do with further tweaking so that it plays nice on tablet (unrelated to this PR obviously).

PR #74, opened 2019-06-14
EN-665 Populate primary nav

antictor:
Adds a hook to the docs-site to fetch navigation data and return the top level navigation items, and mark the current active one.

Implements the new hook into masthead and the default layout so the current navigation options are shown.

Added the ability to add custom props to buttons to help with testing.

Screenshot 2019-06-14 at 11 42 49


m7kvqbe1 (approved):
👍

PR #76, opened 2019-06-17
Feature/en 558 tab set component

m7kvqbe1:
## Related Jira issue:

https://rn-nelson.atlassian.net/browse/EN-558

## Overview

Implement the base tab set component within the react-component-library.

## Reason

>We have a need for tabsets in the docsite. These should follow the same default stying as the component library.

## Work carried out

- [x] Implement tab set component
- [x] Add automated tests
- [x] Perform UAT down to Chrome 38 (and various form factors)

## Screenshot

Screenshot 2019-06-17 at 12 47 57
Screenshot 2019-06-17 at 13 38 50

## Developer notes

This initial implementation meets the docs-site requirements but will need to be expanded upon as component library requirements evolve (there is still an open design ticket at this point in time).


antictor:
How does the focus/active state compare to button and input?

packages/css-framework/src/components/_tab-set.scss
```diff
@@ -0,0 +1,49 @@
+$active-border-color: #2a77C7;
+
+.rn-tab-set {
+  display: flex;
+  flex-flow: column;
+}
+
+.rn-tab-set__head {
+  border-bottom: 1px solid #e2e9ee;
```
antictor: is this in the colour palette?

packages/css-framework/src/components/_tab-set.scss
```diff
@@ -0,0 +1,49 @@
...
+    border-color: $active-border-color;
+  }
+}
+
+.rn-tab-set__tab {
+  margin: 0 spacing(1);
+  padding: spacing(4) spacing(8);
+  border: none;
+  font-size: font-size(xs);
+  font-weight: 700;
```
antictor, line 35: Starting to think we need vars for weight

m7kvqbe1:
> How does the focus/active state compare to button and input?

I'll update to match the button component.

packages/react-component-library/src/components/TabSet/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+              return (
+                <li
+                  key={uuid()}
+                  className={`rn-tab-set__tab-item ${
+                    index === activeTab ? 'is-active' : ''
+                  }`}
+                  data-testid="tab"
+                >
+                  <button
```
antictor, line 41: so.. when you click on a tab it's a button that marks itself as active?

Where's the link... where's the thing that says when I click on a tab, do the thing that tab has been told to do?

m7kvqbe1: When you click on a tab it sets some internal state (`activeTab`) to the index of the tab that was clicked. An `is-active` CSS class is then conditionally applied to the appropriate content pane based on this state.

Are you suggesting an expansion of this functionality in order to add custom actions when tabs are clicked? Invoking a custom callback or something.

packages/css-framework/src/components/_tab-set.scss
m7kvqbe1: Sounds good to me - here i've used the value that most closely reflected the screenshots in the Jira ticket.

m7kvqbe1: Lets discuss this on Wednesday - Designers have already mentioned the need to do a proper review the CSS framework.

packages/react-component-library/src/components/TabSet/index.tsx
antictor: I thought that was the point of the tab, not only that it updates itself but it calls a callback to notify that it's state has changes, so you'd use it to show the correct chunk of content in a page

m7kvqbe1: The current implementation isn't just the tabs, it also includes hidden panes for the content. I could decouple the tabs from the showing / hiding of the content panes and just make clicking the tabs invoke a supplied callback instead.

m7kvqbe1: This functionality has now been implemented.

09txzS1qs (commented):
Is there still work to be done on this? Will approve once @hxltrhuxze has review too.

m7kvqbe1:
> Is there still work to be done on this? Will approve once @hxltrhuxze has review too.

This is now complete and is ready to be reviewed / merged.

antictor (approved):

hxltrhuxze (approved):

PR #78, opened 2019-06-18
Feature/EN-647 Component Markdown Template for Docs Site

hxltrhuxze:
## Overview

Each component in the new docs site now has a space for design documentation. This PR expands the original documentation template to be more in depth in general, and adds a design section.

## Reason

Currently the component template is heavily geared towards the development (implementation) part of the component, but doesn't have any information regarding design choices and usage.

## Work carried out

- [x] Split the title and introduction into their own section
- [x] Added a design section. This includes:
    - [x] An anatomy section, breaking down the component into its individual parts. This allows us to show which parts are required and which are optional (e.g. labels or icons).
    - [x] Sizing and Spacing. This includes an image of the component with red guidelines placed over it. It will show the distance between each individual component and the margin/padding/font sizes that are used to construct the component.
    - [x] States. If applicable to the component, the state section shows all the different state variations (e.g. default, hover, active).
    - [x] Hierarchy & Placement. Where the user should place this component. It includes information regarding usage and where to use the component. An example of this would be the primary button, where you shouldn't place two of them side by side.
- [x] A repeatable block of the section above that targets specific component variations (e.g. week view of the calendar, instead of the month view) and notes any differences to appearance/behaviour.
- [x] An adornments section. This shows the extra modifiers/adornments you can add to a component, and how to use them. An example of this is the pre and post-fixed input blocks.

## Developer notes

I've included basic JSX to outline the difference between the design/develop section. I imagine this won't be the final syntax of the tab component, so that will need updating.


09txzS1qs:
Seems comprehensive and logical, one question:

* Is the template based off existing design systems? If not what's the thought process behind it? 

It would be useful to test the template with a component we've already been built as my feeling is it will be harder to find gaps when it's this generic. Can we do a separate pull request documenting a component we've already built (e.g button) to validate the template?

hxltrhuxze:
So the template is heavily inspired by material.io (e.g. https://material.io/design/components/buttons.html) and bootstrap. The design section in particular closely follows Google's approach, as they cover in depth how to construct a component, and all its relevant usage. By having this, it shows the user _why_ we've made certain design decisions too.

antictor (approved):

09txzS1qs (approved):
Agreed in planning to test current format and understand if it's appropriate level of detail for users and for us to maintain.

PR #80, opened 2019-06-19
Bugfix EN-671 fix react build

antictor:
Fixes the issue not being able to use the react components with hooks in them in the docs site or from other react applications. The root cause was the fact that react was not being stripped from the bundle and the bundle format was incorrect.

Checked it running the Storybook instance for the react components, generated the docs site using the common TextInput and created a standalone react application that used TextInput.


antictor:
Investigating the failing test, which passes locally.. :/

packages/react-component-library/package.json
```diff
@@ -17,6 +17,8 @@
   "license": "MIT",
   "scripts": {
     "build": "concurrently \"webpack -p --config=webpack/prod.js\" \"yarn types\"",
+    "build:dev": "concurrently \"webpack -p --config=webpack/dev.js\" \"yarn types\"",
+    "build:analyze": "concurrently \"webpack -p --config=webpack/analyze.js\" \"yarn types\"",
     "dev": "concurrently \"webpack --watch --config=webpack/dev.js\" \"yarn types --watch\"",
```
m7kvqbe1, line 22: Has this `dev` task been superseded by `build:dev`?

m7kvqbe1 (approved):
Looks good - just left a comment about that old dev task.

packages/react-component-library/package.json
antictor: no. Dev watches for changes, the build dev builds a bundle that is not super compressed and means sensible error messages when tests fail in docs

PR #91, opened 2019-06-26
Fix alignment of docs site (EN-689) and introduce common container style (EN-659)

antictor:
Fixes an issue with alignment in the docs site layout by introducing some common container styles.

Before
Screenshot 2019-06-25 at 10 30 12 (1)

After
Screenshot 2019-06-26 at 12 40 25
Simulator Screen Shot - iPhone Xʀ - 2019-06-26 at 12 30 39


packages/css-framework/src/components/_container.scss
```diff
@@ -1,24 +1,39 @@
 .rn-container {
-  padding-left: 40px;
-  padding-right: 40px;
+  max-width: $navy-content-width;
+  margin: 0 auto;
+  padding: spacing(3);
```
hxltrhuxze: We should avoid vertical padding on the container class. It's purely there to contain content within a defined horizontal range. Having padding by default means we would have to add and remove padding on a case by case basis.

It is much better to have 0 vertical padding on the container and add in the necessary padding on the component. e.g.:
```
<div class="phase-banner">
    <div class="container">
        <h1>Phase Banner Content</h1>
    </div>
</div>
```
In the above example, if you needed vertical padding, it should be added to either the `.phase-banner` class, or the `h1`  - if the entire component needs padding, then that should be added to the `.phase-banner`.

The container should remain component agnostic. If we are centring content, we can safely assume all components will sit within the same `.container` class. This will ensure they line up vertically (much like the problem this pull request is solving). By adding vertical padding, the container is no longer component agnostic and has very prescribed vertical padding that may not apply in every situation.

packages/css-framework/src/components/_container.scss
```diff
@@ -1,24 +1,39 @@
...
-    padding-left: 80px;
-    padding-right: 80px;
+  &.rn-container--small {
+    padding: spacing(2);
   }
 
-  &.centred {
-    max-width: 1600px;
-    margin: 0 auto;
+  &.rn-container--large {
```
hxltrhuxze, line 13: Much like the comments above, having a specified padding on the container now means (in this example), all components that use this class will have an equal vertical and horizontal padding. I feel that has too many assumptions about the components and we will therefore spend a lot of time undoing these rules.

packages/css-framework/src/components/_container.scss
antictor: Have updated the padding for the container to only be on left and right, but have not gone so far as to have a separate div to set the vertical padding otherwise we start to get tags purely to apply extra styles. I think the new code is a good balance.

antictor:
Screenshot 2019-06-26 at 13 56 40

Adjusted the styles to display correctly on iPad.  Just realised that previously the masthead was a mess on ipad

packages/css-framework/src/components/_container.scss
```diff
@@ -1,39 +1,25 @@
...
   }
 
   &.rn-container--large {
-    padding: spacing(5);
+    padding-left: spacing(5);
+    padding-right: spacing(5);
   }
 }
 
 .rn-pagecontainer {
```
hxltrhuxze: what's the difference between `.rn-container` and `.rn-pagecontainer`? Can we just use `.rn-container`?

packages/docs-site/src/components/presenters/footer/footer.scss
```diff
@@ -1,9 +1,15 @@
 @import "@royalnavy/css-framework";
 
 .rn-footer {
+  margin-top: spacing(5);
   background: color(primary, 800);
 }
 
+.rn-footer__container {
```
hxltrhuxze: We don't need to specify a container class for the `footer` or a component. This padding can sit on `.rn-footer`.

antictor: agreed, simplified

packages/css-framework/src/components/_container.scss
antictor: well spotted

hxltrhuxze (approved):

PR #98, opened 2019-06-28
Refactor the TabSet implementation

m7kvqbe1:
## Related Jira issue:

https://rn-nelson.atlassian.net/browse/EN-699

## Overview

Refactor the TabSet implementation to be more JSX centric and use <Tab  /> children.

## Reason

> We need to consume a TabSet in a nested-JSX style rather than programmatic in docs-site MDX.

## Work carried out

- [x] Update implementation
- [x] Update automated tests


packages/react-component-library/src/components/TabSet/TabSet.stories.tsx
```diff
@@ -2,25 +2,21 @@ import React from 'react'
...
-  {
-    title: 'Example Tab 1',
-    content: 'This is some example tab 1 content',
-  },
-  {
-    title: 'Example Tab 2',
-    content: 'This is some example tab 2 content',
-  },
-]
+import Tab from './tab'
```
antictor: why not call it ./Tab

I know it's petty, just reminding myself we need to get a consistent naming convention.

antictor (approved):

PR #102, opened 2019-07-01
Feature/css helper breakpoints

hxltrhuxze:
## Overview
This PR adds support for breakpoints to the CSS framework helper classes.

## Reason

Before this, the helper classes in the CSS framework were very much all-or-nothing. There was no way to set them depending on breakpoint sizes. For example, this meant if you used the helper classes to put content inside a flex container on a single line, there was no way to stack these items on mobile. This PR solves this by adding in breakpoint prefixes to all helper classes:

`md:h_mr-4` would add the `margin-right: spacing(4)` class at the `md` and above breakpoint.

## Work carried out

- [x] Spacing Helpers
- [x] Flex Helpers
- [x] Position Helpers
- [x] Shadow Helpers
- [x] visibility helpers
- [x] typography helpers
- [x] Sizing Helpers
- [ ] Purge CSS

## Developer notes

This PR *does not* contain the ability to purge unused styles from the docs site/react component library. More investigation is needed here.

### DIFF
--- a/packages/css-framework/src/helpers/_flex.scss
+++ b/packages/css-framework/src/helpers/_flex.scss
@@ -1,5 +1,5 @@
 /* stylelint-disable */
-@mixin sharedStyles {
+@mixin flexSharedStyles {
   display: flex !important;
 
   // Direction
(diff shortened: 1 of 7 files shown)


antictor (approved):

m7kvqbe1 (approved):
Looks good to me - does the README need updating?

PR #116, opened 2019-07-03
Feature/EN-679 getting started develop docs

m7kvqbe1:
## Related Jira issue: 

https://rn-nelson.atlassian.net/browse/EN-679

## Overview

Create getting started documentation for developers.

## Reason

> We need a ‘get started’ guide that developers can use to understand what NELSON standards are, how to start coding.

## Work carried out

- [x] Add initial draft
- [x] Expand based on feedback


antictor:
I think we need to write the getting started so it's written on 2 levels. A brief TL:DR type section saying roughly what is said here, but say that it's written to work with NPM and you npm install then the import statements, then I think we need a longer tutorial for people who may have never used node or react before and keep the language aimed at a non-technical level.

When talking with the PM he liked the idea of having something written that he as a non-react developer could follow and make something.

The reason I say this is that many of the people we have spoken too have been non-react or even non-technical users who want to put together a quick app.

antictor:
Maybe give some intro to the components available and the helper classes. Talk about using container classes

m7kvqbe1:
> I think we need to write the getting started so it's written on 2 levels. A brief TL:DR type section saying roughly what is said here, but say that it's written to work with NPM and you npm install then the import statements, then I think we need a longer tutorial for people who may have never used node or react before and keep the language aimed at a non-technical level.
> 
> When talking with the PM he liked the idea of having something written that he as a non-react developer could follow and make something.
> 
> The reason I say this is that many of the people we have spoken too have been non-react or even non-technical users who want to put together a quick app.

Isn't this kind of duping MDN or React docs? My concern here is where does it start and where does it end? How long is a piece of string? There are other pre-existing resources that already provide access to this knowledge.

If you want to consume the component library I think we should expect a base level of knowledge and need to agree what that is.

antictor:
There are many of tutorials out there, maybe too many. People who haven't used React (which seems to be everyone who has wanted to use this so far) wouldn't know where to start.

If I was using this document I'd expect more to give me some idea of how to use this toolkit, whats different or what does it provide?

09txzS1qs:
Agree it could be useful to link to tutorials for React. If we structure the content so experiences devs can get started quickly but then have a separate section for if you're new to React that could be valuable.

antictor:
LGTM as a starter, I think the getting started and learning resources will grow. Pass it by the Product Owner first to get his buy in on Monday and we can get it tested with some real users once we release.

antictor (approved):

09txzS1qs (commented):
@m7kvqbe1 edited content for you to review.

PR #120, opened 2019-07-05
EN-663 Sidebar component

antictor:
A generic sidebar that sits static on the left of the page and expands when the user moves over the top nav section of it.

The notification section is optional and is responsible for revealing a popover, though the application using it is responsible for the content of that popover as this will differ between applications with different notification data.

The user avatar is also optional but will navigate the user to the provided url.

Jul-08-2019 10-54-37


packages/css-framework/src/fragments/_sidebar.scss
```diff
@@ -0,0 +1,137 @@
...
+  text-align: center;
+  display: inline-block;
+  width: $bar-width - $border;
+}
+
+.rn-iconfill {
+  fill: color(neutral, 200);
+}
+
+
```
m7kvqbe1, line 48: Do we need the extra line breaks?

packages/docs-site/src/library/pages/components/sidebar.md
```diff
@@ -0,0 +1,281 @@
...
+
+If the application wishes to display the users name, in the form of their initials then an object must be provided to the `user` property, this must include an `initials` property plus a url property expected by the `LinkComponent`, e.g. 'to'.
+
+The Notification section is responsible for displaying a notification icon if notifications are supported by the application, indicating if there are any unread notifications  and toggling the display of a notification popover containing content passed into it. The host application is responsible for generating the markup and styles used within the notification popover.
+
+### Sidebar without notifications
+A simple sidebar with no notification section, simply showing the current user and using the default Link component so a regular anchor tag is used. 
+
+<CodeHighlighter source={`import { Link } from 'react-router-dom'\n
+const user = { initials: 'ZT', href: '/userprofile' }\n
```
m7kvqbe1: Maybe this is pedantic but should you be using ZT here?

packages/react-component-library/src/fragments/NotificationSidePanel/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+const BAR_WIDTH = 60
+const POPOVER_WIDTH = 335
+
+const NotificationSidePanel: React.FC<NotificationSidePanelProps> = ({
+  children,
+  className = '',
+  unreadNotification,
+}) => {
+  const hasNotifications = children !== null
+  const [showNotifications, setShowNotifications] = useState(false)
```
m7kvqbe1, line 20: Does it make sense to bundle a lot of this into a custom hook?

packages/react-component-library/src/fragments/Sidebar/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+  navItems,
+  NotificationsPopoverContent,
+  unreadNotification = false,
+  user,
+}) => {
+  const [open, setOpen] = useState()
+
+  const renderUserLink = ({ initials, ...rest }: UserLinkType): JSX.Element => (
+    <LinkComponent className="rn-sidebar__nav-link" {...rest}>
+      <Avatar
```
m7kvqbe1, line 29: Is `Avatar` independently documented and can it be used elsewhere or is it tightly coupled to the Sidebar at the moment?

packages/react-component-library/src/icons/index.ts
```diff
@@ -1,5 +1,9 @@
+import Bell from './Bell'
+import Graph from './Graph'
+import House from './House'
 import Search from './Search'
+import Tools from './Tools'
 import TriangleUp from './TriangleUp'
 import TriangleDown from './TriangleDown'
 
-export { Search, TriangleDown, TriangleUp }
+export { Bell, Graph, House, Search, Tools, TriangleDown, TriangleUp }
```
m7kvqbe1, line 9: Are you lifting these from prototype mocks? Are the designers creating these themselves or are we using an off the shelf icon library?

m7kvqbe1 (approved):
Looks good - few queries but nothing blocking.

packages/css-framework/src/fragments/_sidebar.scss
antictor: I put them in just to help organise the file to group styles together

packages/react-component-library/src/icons/index.ts
antictor: The icon library question still seems to be up in the air

packages/react-component-library/src/fragments/Sidebar/index.tsx
antictor: Makes sense to document it on it's own as it's a component. I'll do that in a follow up

packages/react-component-library/src/fragments/NotificationSidePanel/index.tsx
antictor: Not sure how I would

packages/css-framework/src/fragments/_sidebar.scss
antictor: Helps group the styles to understand the structure

packages/react-component-library/src/fragments/NotificationSidePanel/index.tsx
m7kvqbe1: You could export `documentClick` and `toggleNotifications` from the hook and just consume like so:

```const [state, setter, somethingElse, somethingElse] = useCustomHook(arg1, arg2, …)```

Maybe it isn’t warranted where there isn’t going to be code re-use?

Just trying to start a discussion around custom hooks usage - I did this in the Pagination implementation in order to encapsulate related logic, even though it was tightly coupled to the component.

antictor: Will take a look and do it as a follow up

hxltrhuxze:
Couple of points regarding design.

Can we take a look at the notification sheet? The spacing & colours look a tad off:
Screenshot 2019-07-09 at 12 19 26

Also can we check the spacing between the icons and the text in the sidebar? In the designs, the text label is closer to the icon when the menu is expanded:
Screenshot 2019-07-09 at 12 21 27

antictor:
Updated label spacing and discussed reason for notification popover

PR #156, opened 2019-07-24
Fixes to Badge spacing, Sidebar and make Breadcrumb consistent.

antictor:
Changes the Breadcrumb component to use LinkComponent in the parent component like the other Nav components and renamed the list of links to be the same also. Added some more type information to make the component more tightly typed.

Badge spacing was causing layout issues when used in app.

Sidebar had a weird z-index issue


m7kvqbe1 (approved):
Looks good to me - just a snapshot that needs updating.

PR #159, opened 2019-07-24
Bugfix/duplicate svg images for component pages

m7kvqbe1:
## Related Jira issue: 

https://rn-nelson.atlassian.net/browse/EN-732

## Overview

Fix issues with markdown documentation.

## Reason

> Snag the site and fix all the things.

## Work carried out

- [x] Namespace SVG image components
- [x] Add line breaks between PropTable instances
- [x] Tweak issues with some markdown content
- [x] Add SketchWidget to component pages

## Screenshots

localhost_8000_components_pagination


packages/docs-site/src/components/presenters/hero-banner/hero-banner.scss
```diff
@@ -1,35 +1,39 @@
-@import "@royalnavy/css-framework/index.scss";
+@import "@royalnavy/css-framework";
 
 .hero-banner {
-  background-color: color(primary, 800);
+  background: color(primary, 800);
```
hxltrhuxze: Any reason why we're generalising the background, rather than explicitly setting `background-color`?

hxltrhuxze (approved):

packages/docs-site/src/components/presenters/hero-banner/hero-banner.scss
m7kvqbe1: No reason this was changed and changed back after realising both footer and HeroBanner were already picking the same background colour. Might have been an issue using a system colour picker maybe there was some window shadow overlaying when he person who wrote the ticket colour picked.

m7kvqbe1: Changed back to `background-color`.

antictor (approved):

PR #167, opened 2019-07-24
Bugfix/logo width

hxltrhuxze:
## Overview
Fixing the main NELSON logo on mobile so that it doesn't cause horizontal scrolling.

## Reason
The logo was fixed width, so was extending beyond its wrapper on mobile

## Work carried out

- [x] Fixed logo


packages/docs-site/src/components/presenters/Masthead/masthead.scss
```diff
@@ -127,8 +127,8 @@
   }
 
   .masthead__logo {
-    width: 405px;
-    height: 52px;
+    width: 300px;
+    // height: 52px;
```
m7kvqbe1:
```suggestion
```

m7kvqbe1 (approved):
Do we need to keep that comment?

PR #183, opened 2019-07-29
Feature/EN-741 radio component

m7kvqbe1:
## Related Jira issue: 

https://rn-nelson.atlassian.net/browse/EN-741

## Overview

Implement a Formik friendly Radio component.

## Reason

>The Radio Button component should be a toggle that can either be in an off or on state. It should also be able to accept a label to describe the component.

## Work carried out

- [x] Implement base component and style (m7kvqbe1)
- [x] Add to Storybook (m7kvqbe1)
- [x] Integrate with Formik library (m7kvqbe1)
- [x] Add automated tests (m7kvqbe1)
- [x] Create developer documentation (m7kvqbe1)
- [x] Perform UAT down to Chrome 38 (and across a variety of form factors)
- [x] Adjusted the CSS to match the design comps more closely (hxltrhuxze)
- [x] Added the Design Docs (hxltrhuxze)
- [x] Added related SVG imagery (hxltrhuxze)
- [x] Decouple from Formik (m7kvqbe1)
- [x] Test on IE11

## Screenshot

Screenshot 2019-07-29 at 14 14 30


antictor:
Have you tested it on IE 11?

m7kvqbe1:
> Have you tested it on IE 11?

No - i'll do this

09txzS1qs (commented):
on style - looks like when radio is selected it's significantly smaller and aligned right than the two above. Is there a reason for this @hxltrhuxze ? I don't think the size of the radios should change with their state, and the change should be indicated with contrast / layers - see https://design-system.service.gov.uk/components/radios/ as an example. 

Also are they smaller alongside the text to save on screen real estate? I know I already signed off design so apologies for late questions but on second pass I want to iron out these concerns.

hxltrhuxze:
@09txzS1qs All the radio buttons above are the same size. The item that is throwing off the perspective is the Hover/Active state in the middle. If you check out the radio buttons in the example you linked, the blue background in the Hover/Active state is exactly the same as the yellow background that appears on the GDS radios when you click them.

Regarding size: They're sized correctly according to the label font size. GDS generally has oversized elements to aid with usability, however they're not suitable for producing the types of interfaces we're designing.

Here is the Bootstrap radio buttons with labels:
Screenshot 2019-08-07 at 12 28 54

And here are the Material.io radios:
Screenshot 2019-08-07 at 12 30 36

antictor:
Did this get tested on IE11 yet?

m7kvqbe1:
> Did this get tested on IE11 yet?

Not yet it was still being worked on. I will do this though. Is that our minimum target environment now?

09txzS1qs:
> @09txzS1qs All the radio buttons above are the same size. The item that is throwing off the perspective is the Hover/Active state in the middle. If you check out the radio buttons in the example you linked, the blue background in the Hover/Active state is exactly the same as the yellow background that appears on the GDS radios when you click them.
> 
> Regarding size: They're sized correctly according to the label font size. GDS generally has oversized elements to aid with usability, however they're not suitable for producing the types of interfaces we're designing.
> 
> Here is the Bootstrap radio buttons with labels:
> Screenshot 2019-08-07 at 12 28 54
> 
> And here are the Material.io radios:
> Screenshot 2019-08-07 at 12 30 36

Thanks for clarifying, makes sense. All good from my perspective then.

m7kvqbe1:
This has now been tested on IE11.

hxltrhuxze (approved):

PR #198, opened 2019-08-06
Feature/EN-762 decouple formik from textinput

m7kvqbe1:
## Related Jira issue:

https://rn-nelson.atlassian.net/browse/EN-762

## Overview

Decouple the `TextInput` component from Formik and add an enhancer to allow easy Formik usage.

## Reason

> Don't be prescriptive about technology - ensure compatibility with multiple form libraries or vanilla usage. An abstraction layer is required to enable this.

## Work carried out

- [x] Decouple component implementation from Formik
- [x] Create a `withFormik` enhancer
- [x] Update automated tests accordingly
- [x] Fix issue where modifiers were used for stateful classes
- [x] Make error styles generic
- [x] Add vanilla usage examples to storybook

## Developer notes

The enhancer can be used by all future form components.


packages/react-component-library/src/enhancers/withFormik.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+import FieldProps from '../types/FieldProps'
+import FormProps from '../types/FormProps'
+
+export interface FormikProps {
+  field: FieldProps
+  form: FormProps
+}
+
+const withFormik = (FormComponent: React.FC<any>) => ({
```
antictor, line 11: Don't need &lt;any&gt;

packages/react-component-library/src/components/index.ts
```diff
@@ -1,3 +1,4 @@
+// Components
 import Avatar from './Avatar'
```
antictor, line 2: One thing I want to change to. Don't have a default export in a component and then in the index for components use:

export * from './TextInput'

antictor (approved):

packages/react-component-library/src/enhancers/withFormik.tsx
m7kvqbe1: Hmm surely I do need any because the component props are arbitrary? Not sure how else I would define a contract for arbitrary props?

packages/react-component-library/src/components/index.ts
m7kvqbe1: Sounds good to me.

Issue #187, opened 2019-08-01
Add Masthead component


antictor:
Applications require a top bar that communicates the application/service name and optionally provides features for: search, user profile info, notifications and navigation.

The Masthead may or may not be used in conjunction with the Sidebar.

The Masthead needs to be modular to the application in question is able to use any combination of its features, including all or none of them.

Currently we don't have any applications that require the Masthead to support mobile, but this will come in time.

As this is potentially complex it may be worth creating the Masthead in stages, though the design should include all options to ensure that it will evolve without issue.

Initially the primary project for this simply requires the ability to display a logo and service name.  

Another project, re-starting soon will also require the user profile section to be added.  

Finally there is a project being worked on elsewhere that requires a Masthead including search and navigation.

<img width="1038" alt="Screenshot 2019-08-01 at 14 13 44" src="https://user-images.githubusercontent.com/48056118/62296409-e9948780-b466-11e9-8dcb-2c3a3cd56929.png">

<img width="1097" alt="Screenshot 2019-08-01 at 14 14 17" src="https://user-images.githubusercontent.com/48056118/62296347-cc5fb900-b466-11e9-8e6d-ad1078135eb1.png">

<img width="501" alt="Screenshot 2019-08-01 at 14 13 56" src="https://user-images.githubusercontent.com/48056118/62296348-cc5fb900-b466-11e9-843b-62eacbec05e1.png">

<img width="516" alt="Screenshot 2019-08-01 at 13 29 25" src="https://user-images.githubusercontent.com/48056118/62295964-1a27f180-b466-11e9-8323-f0e50519844f.png">

antictor:
Creating a basic version first that just has the logo and service name for current project.

hxltrhuxze:
I've updated both the mobile and desktop comps to include placement for:
- Search
- Notifications
- Profile

---

## Desktop Masthead
The Desktop version of the Masthead keeps the placement of the Notification Badge and the Profile Avatar, however adds the ability to have a Search Button next to them.

When the Search Button is clicked, the new Search Bar replaces the the nav with a dark background input. The input takes up the full height of the nav.

<img width="790" alt="Screenshot 2019-08-05 at 16 30 58" src="https://user-images.githubusercontent.com/48090803/62476226-7ceaf700-b79e-11e9-8852-156ec067da40.png">

---

## Mobile Masthead
To maintain usage consistency, not a lot has changed between the desktop and mobile versions of the Masthead.

[...]

PR #240, opened 2019-08-27
Feature/masthead #187

antictor:
Linked to issue #187 

Finally....

A masthead component that can display the service name and icon at the top of the page. In addition to the basic elements the component can display  search bar, notifications (with indicator), a link to the users profile and a primary navigation that looks similar to tab nav but does not switch to vertical layout on mobile, instead it scrolls horizontally.

Tested on IE11, Chrome and mobile safari.

Some notes:

Getting horizontal scroll to work was a challenge. One library proved useful to allow scroll to work on desktop browsers but didn't work at all with mobile safari. The end result was a mix of the library and some webkit specific css rules. There is a Slight bug with this in that animation of opacity over a css element using touch scroll can incorrectly briefly show the scroll element on top of the fading element. This is a bug with IOS 12 that has been fixed in IOS 13.

In all this ended up being a really complicated feature, even though the end result looks simple to the user.

iphone

ipad

ie11

chrome


hxltrhuxze:
Some feedback:
- Can we reduce the weight of the links in the nav? On first inspection, these appear to be too heavy.
- The Search bar and the background of the search icon need to animate together - At the moment there is a slight delay between the search bar fading in and the background around the search icon inverting in colour.
- Can we make the nav the same height as the search bar? Currently the spacing is off for the nav links causing them to appear closer to the top of their container than the bottom (and also making the search bar taller than it).
- Notification sheet:
    - Can we build the original design for the notification sheet?
    - Can we add the icon from the design for the "View all notifications" button?

Apart from that, awesome job :) Really like what you've done here! 🎉

antictor:
The icon and searchbar animate different because one is a opacity change the other is a colour cross fade.. trust me I've tried to get them more in sync. Open to changes and suggestions.

The font weights are set exactly like the design but I'm not totally sure what the browser is rendering them as. As far as I can see the font definitions are correct and the weight but I'll lighten them.

I'll double check the searchbar height and nav spacing. I had the height matching exactly but then in some screen sizes and browsers it would be a pixel or two too small and the underline would show and be distracting.

With the notification sheet that is a separate component that needs to be built and this is just here right now to act as a a placeholder and container for the notification content so it's out of scope for this component. I'll go look at the design for the view all icon and see if I can get that in.

antictor:
Heights adjusted to match and vertical alignment sorted.

antictor:
Screenshot 2019-09-03 at 13 58 06
Screenshot 2019-09-03 at 13 58 00

packages/css-framework/src/components/_searchbar.scss
```diff
@@ -0,0 +1,95 @@
...
+}
+
+.rn-searchbar-enter {
+  opacity: 0.5;
+}
+
+.rn-searchbar-enter.rn-searchbar-enter-active {
+  opacity: 1;
+  transition: opacity 300ms;
+  z-index:999;
```
m7kvqbe1: There is a z-index helper in the CSS framework I think:

https://github.com/Royal-Navy/standards-toolkit/tree/develop/packages/css-framework#z-index

packages/docs-site/src/library/pages/components/masthead.md
```diff
@@ -0,0 +1,182 @@
...
+The Masthead is a simple section at the top of the page that allows the developer to communicate the service logo, name and offer a search bar.
+
+## Usage
+
+<TabSet>
+
+<Tab title="Design">
+
+  ## Design
+  TBD.
```
m7kvqbe1: We need to make sure design docs are added before merge to develop or it'll block releases.

packages/react-component-library/jest/setupTests.ts
```diff
@@ -1,4 +1,5 @@
 import { configure } from 'enzyme'
 import EnzymeAdapter from 'enzyme-adapter-react-16'
```
m7kvqbe1, line 2: Do we still have enzyme tests that need porting over to RTL?

packages/react-component-library/src/components/Searchbar/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+  useEffect(() => {
+    document.addEventListener('mousedown', documentClick)
+
+    return () => {
+      document.removeEventListener('mousedown', documentClick)
+    }
+  }, [])
+
+  return (
+    <Formik initialValues={initialValues} onSubmit={onSubmit}>
```
m7kvqbe1: It is being tightly coupled to Formik here. Could we avoid this maybe?

packages/react-component-library/src/fragments/Masthead/Masthead.tsx
```diff
@@ -0,0 +1,150 @@
...
+  navItems?: NavItemTypes[]
+  NotificationsPopoverContent?: JSX.Element
+  onSearch?: (term: string) => void
+  searchPlaceholder?: string
+  title: string
+  unreadNotification?: boolean
+  user?: UserType
+}
+
+export const Masthead: React.FC<MastheadProps> = ({
```
m7kvqbe1, line 23: Might be nice to build up the Masthead using composition instead of using props to initialise it. Could have broken up the code a bit more. Notably the search and notifications. This is a really chunky component. Maybe just something for us to consider in the future.

packages/react-component-library/src/fragments/Masthead/Masthead.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+          {NotificationsPopoverContent && (
+            <NotificationPanel
+              buttonClassName="rn-masthead__option"
+              className="rn-masthead__notification"
+              data-testid="masthead-notifications"
+              notificationPlacement="below"
+              onHide={() => setShowNotifications(false)}
+              onShow={() => setShowNotifications(true)}
+              scheme="light"
```
m7kvqbe1, line 104: Should this prop value be hard coded?

m7kvqbe1 (commented):
A few comments - nothing major.

That was a beast of a PR - I think we should probably have separated that work out into more discrete chunks. It was difficult to get my head around how everything in there fit together.

packages/react-component-library/src/fragments/Masthead/Masthead.tsx
antictor: I broke out the search bar, navbar and notification but will have one last look through.

packages/css-framework/src/components/_searchbar.scss
antictor: There is indeed and I did use it at one point, will take another look

packages/docs-site/src/library/pages/components/masthead.md
antictor: Yeah, and thats going to be a thing going forward for all components that we need to discuss as a team

packages/react-component-library/jest/setupTests.ts
antictor: No idea, can look after the PR

packages/react-component-library/src/components/Searchbar/index.tsx
antictor: its coupled internally so the outside world is not aware, or care. I thought it would be ok. All that the consumer is bothered about is that they get a call to a function with the term to search for

packages/react-component-library/src/fragments/Masthead/Masthead.tsx
antictor: Its telling it to use the light version of notifications, compared to sidebar using the dark so I think it's ok

packages/react-component-library/src/components/Searchbar/index.tsx
antictor: Though I may have a look and see if I could strip Formik out as it's a single field

antictor:
I know it was a big PR, I had already broken down alot into PR's before this one but was torn about submitting Searchbar etc as separate PR's. The reason I pushed ahead was pragmatism and worry that it would drag on even longer. But I know if was not a nice PR to review. Next time :)

m7kvqbe1 (approved):

PR #242, opened 2019-08-29
Hotfix/Remove standards-toolkit.sketch

m7kvqbe1:
Remove incorrect Sketch toolkit - will be replaced later once correct file is sourced.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -385,8 +385,8 @@ workflows:
                 - develop
       - deploy_docs_site_production:
           requires:
-            - publish_css-framework
-            - publish_react-component-library
+            # - publish_css-framework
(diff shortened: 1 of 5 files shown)


ghost (dismissed):

antictor (dismissed):

antictor (dismissed):

m7kvqbe1:
- Yarn resolutions only work from the root package.json
- You have to generate a brand new lock file for new resolutions to be respected (means lots of other deps end up bumped)
- TypeScript compiler throws an error from a union type that was previously fine (temp fix with an any)
- Had to explicitly add `to-style` dep for Gatsby
- Yarn sucks, NPM sucks

Issue #241, opened 2019-08-29
Unable to use form components with Formik in 1.1.0 and above


antictor:
The `withFormik` enhancer is not exported with the react component library so users are unable to use the fields in a formik form.

Can we either export `withFormik` or possibly already enhance them and have them as `Formik.TextInput`.

m7kvqbe1:
Good idea - lets export enhanced versions in addition to the enhancer itself.

It appears I exported the enhancer in the wrong index file.

m7kvqbe1:
https://github.com/Royal-Navy/standards-toolkit/pull/245

PR #245, opened 2019-08-30
Bugfix/add formik enhancer exports

m7kvqbe1:
## Related issue

https://github.com/Royal-Navy/standards-toolkit/issues/241

## Overview

Fix issues with Formik enhancer exports and export pre-enhanced components.

## Reason

Exports were previously added to the incorrect index.

## Work carried out

- [x] Tidy up stale component exports
- [x] Export `withFormik` enhancer correctly
- [x] Pre-enhance components and export via `Formik` object
- [x] Add `suppressImplicitAnyIndexErrors` to TS config

## Developer notes

`suppressImplicitAnyIndexErrors` has been added in order to allow developer access to dictionary objects that aren't under the developers control going forward.

### DIFF
--- a/packages/react-component-library/src/components/index.ts
+++ b/packages/react-component-library/src/components/index.ts
@@ -3,6 +3,7 @@ import Badge from './Badge'
 import Breadcrumbs from './Breadcrumbs'
 import Button from './Button'
 import Checkbox from './Checkbox'
+import Link from './Link'
 import Nav from './Nav'
 import Pagination from './Pagination'
 import PhaseBanner from './PhaseBanner'
@@ -11,14 +12,16 @@ import Sidebar from '../fragments/Sidebar'
 import Radio from './Radio'
 import Tab from './TabSet/Tab'
 import TabSet from './TabSet'
+import TabNav from './TabNav'
 import TextInput from './TextInput'
 
-export default {
+export {
   Avatar,
   Badge,
   Breadcrumbs,
   Button,
   Checkbox,
+  Link,
   Nav,
   Pagination,
   PhaseBanner,
@@ -27,5 +30,6 @@ export default {
   Sidebar,
   Tab,
   TabSet,
+  TabNav,
   TextInput,
 }
--- a/packages/react-component-library/src/components/index.ts
+++ b/packages/react-component-library/src/components/index.ts
@@ -3,27 +3,29 @@ import Badge from './Badge'
 import Breadcrumbs from './Breadcrumbs'
 import Button from './Button'
 import Checkbox from './Checkbox'
-import Link from './Link'
 import Nav from './Nav'
+import Pagination from './Pagination'
 import PhaseBanner from './PhaseBanner'
 import Popover from './Popover'
-import TabSet from './TabSet'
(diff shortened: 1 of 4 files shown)


packages/react-component-library/src/enhancers/index.tsx
```diff
@@ -0,0 +1,13 @@
+import * as Components from '../components'
+import withFormik from './withFormik'
+
+const Formik: object = {
+  // Enhanced components
+  // ...
+}
+
+Object.keys(Components).forEach(key => {
```
antictor: You are enhancing every component, not just form ones?

I'd have just picked ones that can be enhanced

packages/react-component-library/src/components/index.ts
```diff
@@ -1,21 +1,20 @@
-// Components
 import Avatar from './Avatar'
 import Badge from './Badge'
 import Breadcrumbs from './Breadcrumbs'
 import Button from './Button'
 import Checkbox from './Checkbox'
 import Link from './Link'
 import Nav from './Nav'
+import Pagination from './Pagination'
```
antictor, line 8: Going forward, not here, it seems Typescript favours not using `export default` and allows something like:

`
Button.tsx

export const Button = () => {}

---

index.ts

export * from './Button'

`

This gets around the having to import and then export and also makes it easier to include things like type definitions in a component and have them automatically exported too.

packages/react-component-library/src/index.ts
```diff
@@ -30,4 +35,9 @@ export {
   Tab,
   TabSet,
   TextInput,
+
+  Icons,
```
antictor, line 39: Why have spaces and can we just do it alphabetically to make it easier to find things and avoid possible duplication?

packages/react-component-library/src/enhancers/index.tsx
m7kvqbe1: Haha yes good point! I'll change!

packages/react-component-library/src/index.ts
m7kvqbe1: The exports are organised in a way that corresponds with the imports at the top of the file.

Happy to make it all alphabetical though.

antictor: I think when we move to the alternative export method I mention we do it then

packages/react-component-library/src/enhancers/index.tsx
```diff
@@ -0,0 +1,19 @@
+import TextInput from '../components/TextInput'
+import Checkbox from '../components/Checkbox'
+import Radio from '../components/Radio'
+
+import withFormik from './withFormik'
+
+const Formik: object = {
```
antictor, line 4: import { Checkbox, Radio, TextInput } from '../components'

const Formik = [Checkbox, Radio, TextInput].map(component => withFormik(component))

antictor (approved):

packages/react-component-library/src/enhancers/index.tsx
m7kvqbe1: Changed import statements but want Formik to be an object so not using map

PR #272, opened 2019-09-12
Select component

antictor:
**TODO - Design docs**

The original Select component was meant to look and act like a simple dropdown with the ability to be used in a form but after more investigation it has to act more like a mix between a select and an autocomplete form in order to be able to provide the functionality required.

The component has been constructed by putting a wrapper around (https://react-select.com/home)[React Select].

The user is able to pass a list of options, a label and a field name to a the component and it will provide field that can be typed into with suggested matches shown or clicked on to reveal options and the user picks the one they want.

Select

In order to review this PR I would suggest looking at the example in the docs site or run storybook to get a feel for how it works.


m7kvqbe1 (approved):
LGTM 

Are we publishing the `reactapp` package or just leaving it in there as a resource to aid testing?

m7kvqbe1:
Couple of codeclimate bits need looking at - looks like they are non-issues though.

antictor:
The react test app is not for publication

Issue #222, opened 2019-08-16
Dropdown component


ghost:
## Description

Similar in appearance to a select component, a dropdown can contain a variety of different forms of content such as navigation items, filters etc. It is accessed via a button.

## Component classification

Is this component an atom, molecule or organism?
- [x] Atom
- [ ] Molecule
- [ ] Organism

## Details / user needs

Dropdowns are a familiar interaction element. They are used to wrap extended functionality into an interface under the wrapper of a single description.

Dropdown's should be designed with flexibility in mind. They can contain a number of types of information and can require different actions from users.

Some design considerations:
- Dimensions of dropdown - ideally this should be left to the App designer
- Content of the dropdown should be divided into relevant sections
- How will the dropdown be labelled?
- How will the content of the dropdown be accessible?
- The title of a dropdown can contain more than just text: an icon or label might be present

_Used in 2x exemplar projects_

## Definition of done
- [x] The Sketch file is updated with the component
- [x] There are screenshots of all component variations in this ticket
- [x] All states are included if applicable (default, hover, active, focus)
- [x] A design documentation PR is linked from this ticket
- [ ] All questions laid out in this ticket have been resolved 
- [x] A pull request for a component must contain a fully tested, documented and finished component in a state it can be shared with partner developers

hxltrhuxze:
# Dropdown Component

<img width="797" alt="Screenshot 2019-09-12 at 11 46 43" src="https://user-images.githubusercontent.com/48090803/64778140-65e4b500-d553-11e9-9a18-ddcb5c5bcc83.png">

The Dropdown component provides a way to group/hide multiple actions under one clear heading. It looks similar to the select component, but with a few differences:

- The Dropdown label is fixed. When using a select, the item clicked on in the list will sit beneath the label in the main select input area. The Dropdown doesn't do this, as all the items in the dropdown are independent actions that are not being "set" by the component.
- The main Dropdown bar can accept an icon on the left of the Dropdown Label.
- The Dropdown sheet that appears on click is visually separated from the main Dropdown bar. This is to help distinguish it from the Select component.


### Dropdown item
The Dropdown item has multiple different variations:
<img width="692" alt="Screenshot 2019-09-12 at 12 11 16" src="https://user-images.githubusercontent.com/48090803/64779702-8f531000-d556-11e9-82ec-fd0c627db26b.png">

1. Default Items
[...]

PR #276, opened 2019-09-19
Dropdown Component

antictor:
## Related issue
#222

## Overview
New Dropdown component designed to show a link that when clicked on shows a list of actions that can be issued.

### How
The component uses the same 3rd party tool as the `Select` component, to provide a select like experience that can be used by mouse of keyboard navigation. The different between the two is that this component has no text input, cannot be used in a form and does not show the selected value in the main label area.

### DIFF
--- a/packages/react-component-library/src/components/Dropdown/index.tsx
+++ b/packages/react-component-library/src/components/Dropdown/index.tsx
@@ -1,0 +1,2 @@
+export * from './Dropdown'
+export * from './DropdownOption'
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,14 +5,14 @@ orbs:
 
 commands:
   dependencies:
-    description: 'Load dependency cache, get dependencies and update cache'
+    description: "Load dependency cache, get dependencies and update cache"
     steps:
       - restore_cache:
           keys:
             - yarn-packages-v1-{{ .Branch }}-{{ checksum "yarn.lock" }}
             - yarn-packages-v1-{{ .Branch }}-
             - yarn-packages-v1-
-      - run: yarn install --frozen-lockfile
+      - run: yarn install
       - save_cache:
           paths:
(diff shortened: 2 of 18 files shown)


packages/react-component-library/src/components/Dropdown/index.tsx
```diff
@@ -0,0 +1,2 @@
+export * from './Dropdown'
```
m7kvqbe1, line 1: Can you explain the thinking behind this new way of handling the component exports?

m7kvqbe1:
@antictor 

I notice you've squashed your commits ahead of this PR. Do you think we should make a habit of doing this going forward to try and keep the history a bit nicer to look at?

Squash at the point of reaching base implementation / at the point the PR is raised

or maybe...

Squash at merge (less keen on this)

packages/react-component-library/src/components/Dropdown/index.tsx
antictor: It's the recommended  way to do exports in TS. It re-exports all the things exported from a thing and also advises not to use default exports.

So you have

Thing.ts

const export type ThingProps {
  name: string
}

export const Thing:React.FC<ThingProps> Thing = ({ name }) => (<p>{name}</p>)


------
Components 
export * from './Thing' 

-------

import { Thing, ThingProps } from './components'

or even 

import { Thing, ThingPros} from './components/Thing'



I started using the pattern and find it easier for things like what we are doing.

antictor:
Chatted about the squash thing. Like to keep a tidy commit history that is to the point and doesn't contain distractions.

packages/react-component-library/src/components/Dropdown/index.tsx
m7kvqbe1: Ok cool - i'll adopt this pattern going forward.

PR #283, opened 2019-09-21
Bug/icon dependency

antictor:
Previously the build of packages was done in parallel with no consideration for cross dependencies because this wasn't an issue.

Whilst working on the Dropdown component, which uses icons in storybook and tests, it would fail due to the fact there wasn't a built copy of the icons before it reached building the react code.

The fix is to tell the build script to only build 1 thing at a time and it will analyse their dependencies to figure out the order.

### DIFF
--- a/packages/icon-library/package.json
+++ b/packages/icon-library/package.json
@@ -10,6 +10,7 @@
   "files": [
     "dist"
   ],
+  "types": "dist/index.d.ts",
   "scripts": {
     "build": "run-s svgr exports build:ts types",
     "build:ts": "webpack -p --config=webpack/prod.js",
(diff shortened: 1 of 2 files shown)


packages/icon-library/package.json
```diff
@@ -10,6 +10,7 @@
   "files": [
     "dist"
   ],
+  "types": "dist/index.d.ts",
```
m7kvqbe1, line 13: There is another PR that already adds this missing line.

m7kvqbe1 (approved):

packages/icon-library/package.json
m7kvqbe1: Duplicate here: https://github.com/Royal-Navy/standards-toolkit/pull/280

antictor: Ah, must have cherry picked 2 commits by mistake

PR #285, opened 2019-09-23
Dropdown Component

antictor:
- [x] Develop component code
- [x] Developer documentation
- [x] Design documentation

## Related issue
#222 

Introduction of a new Dropdown component. This component is similar to the Select except it is used to provide a dropdown of actions (a bit like a menu) rather than a form component. As such this component doesn't have a field to show the currently selected item and only response to a callback property to indicate a selected action.

### Standard dropdown
standard

### Dropdown with left icons
lefticons

### Dropdown with scrolling options
scroll

### Internet Explorer 11
window


hxltrhuxze:
Can we have the label the same colour as the value state would be? so `color(neutral, 500)`?

Also, couple of notes on the Dropdown arrow:
1. When the Dropdown becomes active, can we transition the arrow to blue and rotate it 90deg?
2. Following that, can we also move the arrow `translateY(-1px)` to re-centre the icon in the middle of the bar?

EDIT: Sorry, another point. Can we make the default width of the dropdown and the sheet smaller? It's using a lot of horizontal space at the moment

antictor:
The field fills the container width so it is up to the developer how wide they want to make it. I didn't want the developer to be limited in width.

I'll take a look at the transition.

antictor:
spin

hxltrhuxze:
Sorry, one more minor tweak - the dropdown arrow needs to be in its original place when the dropdown is in its default state. It should only shift up a pixel when it's active.

antictor:
it only applies the 1px correction when it's open

antictor:
Screenshot 2019-09-23 at 16 28 22
Screenshot 2019-09-23 at 16 28 27

packages/css-framework/src/components/_dropdown.scss
```diff
@@ -1,124 +1,135 @@
+$borderThick: 2px;
```
m7kvqbe1: Elsewhere the convention seems to be hyphenated variable names (`$border-thick`).

packages/react-component-library/src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,38 @@
...
+  label: string
+}
+
+export const Dropdown: React.FC<DropdownProps> = ({
+  onSelect,
+  options,
+  label,
+}) => {
+  const onChange = (option: ValueType<DropdownOption>) => {
+    const dropdownOption = option as DropdownOption
```
m7kvqbe1, line 21: The as keyword in TS is new to me. So it is basically just casting the object with that type?

packages/react-component-library/src/components/Dropdown/DropdownLabel.tsx
```diff
@@ -0,0 +1,50 @@
+import React from 'react'
+
+import { Visibility, VisibilityOff } from '../../icons'
```
m7kvqbe1, line 3: If we can delay merging this i'll hopefully have a fix for the tree shaking issue today.

m7kvqbe1 (approved):
Looks good to me - might be good to hold off merging to see if we can get the tree shaking fix in pre-release.

packages/react-component-library/src/components/Dropdown/DropdownLabel.tsx
antictor: I'd rather merge it and then retrofit the improvements then it wont get held up if the other ticket does.

packages/react-component-library/src/components/Dropdown/Dropdown.tsx
antictor: Yeah it just forces a cast. In this case it was needed a two types are the same but it is not possible to say that.

packages/css-framework/src/components/_dropdown.scss
antictor: Good spot

hxltrhuxze (approved):

Issue #289, opened 2019-09-23
Using library icons increases bundle size and breaks build


antictor:
If I include an icon from `@royal-navy/icon-library` in the react component library the compile time goes from 6 seconds to 30+ seconds and the bundle size increases by almost a megabyte.

If I include an icon from `@royal-navy/icon-library` in a markdown file int the documentation site the build time goes from a minute to about 8 minutes and this also causes Netlify to fail as the build times out.

For both options we need to find why it is in effect trying to re-compile the icons and how to tree shake so that in the case of the component library it only includes the icons needed which should barely affect the bundle size.

m7kvqbe1:
So, what we've learned is that in order to take advantage of tree shaking, you must...

- Use ES2015 module syntax (i.e. import and export).
- Ensure no compilers transform your ES2015 module syntax into CommonJS modules (this is the default behavior of the popular Babel preset @babel/preset-env.
- Add a "sideEffects" property to your project's package.json file.
- Use the production mode configuration option to enable various optimizations including minification and tree shaking.

PR #291, opened 2019-09-24
Bugfix/Enable tree shaking

m7kvqbe1:
## Related issue

https://github.com/Royal-Navy/standards-toolkit/issues/289

## Overview

Get tree shaking working by updating some build config.

## Reason

> The resulting bundles from the build were massive as entire dependencies were being included as opposed to just the modules that were being consumed. This helps to ensure that developers downstream consuming our packages can utilise tree shaking as part of their own builds.

## Work carried out

- [x] Stop transforming modules to commonJS
- [x] Specify sideEffects false
- [x] Tested against docs-site (consumes the component library)
- [x] Update Babel config to transform ESM to CommonJS for NODE_ENV `test`

## Screenshots

Before:

Screenshot 2019-09-24 at 15 23 28

After:

Screenshot 2019-09-24 at 15 26 59

## Developer notes

- ~~There are still issue with the time it takes to generate build since we disabled concurrency. Currently investigating this as part of a separate issue.~~

- Have had to specify an environment in the Babel config instead of using the Jest `transformIgnoreModules` option as the latter does not play nice with Yarn workspaces (https://github.com/facebook/jest/issues/5325).


m7kvqbe1:
Tests currently failing because Jest (node) doesn't know how to handle ES6 modules out the box. Updating accordingly.

antictor (approved):

Issue #208, opened 2019-08-15
Number clicker component


ghost:
## Description

A numerical input component that allows users to increment values by a set amount.

## Component classification

Is this component an atom, molecule or organism?
- [x] Atom
- [ ] Molecule
- [ ] Organism

## Details / user needs

Users often need to be guided as to:
- The fact that an input is numeric
- It can only be incremented as a specific value (e.g. multiple of 10)
- It has a specific unit, for example percentage, hours or £££

We need a component that allows this input to be made in a quick, accessible and easy to comprehend way.

We should ensure that when using a mobile keypad, the numeric keypad is selected when the element is in focus.

![number-clicker](https://user-images.githubusercontent.com/52033457/63095431-1feffd80-bf63-11e9-979f-de5265b0fadf.png)

_Used in 3x exemplar projects_

## Variations

- Pre and post fixed fixed functionality
- Enabled, disabled, in focus, hover states

## Questions

- There has been some discussion about how to treat error scenarios when the user inputs a value that is not a valid increment. This needs further discussion with the Product Owner.

## Definition of done
- [ ] The Sketch file is updated with the component
- [ ] There are screenshots of all component variations in this ticket
- [ ] All states are included if applicable (default, hover, active, focus)
- [ ] A design documentation PR is linked from this ticket
- [ ] All questions laid out in this ticket have been resolved 
- [ ] A pull request for a component must contain a fully tested, documented and finished component in a state it can be shared with partner developers

hxltrhuxze:
# Number Clicker
The simple number clicker can either have a value entered in the same manner as a text input, or the user can use the incremental up/down arrows.

<img width="468" alt="Screenshot 2019-08-27 at 17 05 33" src="https://user-images.githubusercontent.com/48090803/63788183-e9b06780-c8ec-11e9-9b81-c555be2f1b2a.png">


### Abstract link
https://share.goabstract.com/e13d7d86-745f-4c19-bc04-d57542754099?sha=c7226d3e551d7d83470dbb287d92cbc868ae406d

PR #295, opened 2019-09-24
Number Input

antictor:
## Related issue
#208 

**Adds a new component `NumberInput` also known as 'number clicker'. I changed the name to NumberInput to be consistent with TextInput.**

Introduces a form element to allow users to enter numeric information either directly into the field or using arrow buttons to increase/decrease the number.

If the user tries to enter in text it is ignored. If the user tries to enter a number that is too big or too small it is ignored.

The developer can specify the increments the number should increase or decrease by. If the user does tries to enter a number between those gaps they are able to do so, so this should be enforced through form validation. Formik and Yup can capture this so when the user moves to another field it highlights the error.

number input


packages/css-framework/src/components/_numberinput.scss
```diff
@@ -0,0 +1,146 @@
...
+}
+
+.rn-numberinput__increase,
+.rn-numberinput__decrease {
+  display: flex;
+  justify-content: center;
+  flex-direction: column;
+  width: 42px;
+  align-items: center;
+
```
m7kvqbe1, line 115: Do we need this extra whitespace?

packages/docs-site/src/library/pages/components/forms/input.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Input
+title: Text Input
```
m7kvqbe1, line 2: Nit picking - should this be another PR? I guess it's a non-consequential change.

packages/reactapp/src/NumberForm.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react'
```
m7kvqbe1, line 1: I see you are using this app for testing? Because working with state in storybook is a real pain?

m7kvqbe1 (approved):
LGTM - just a couple of comments.

packages/css-framework/src/components/_numberinput.scss
antictor: don't need it it just helped me read the code and understand what was going on

packages/docs-site/src/library/pages/components/forms/input.md
antictor: I changed it in this PR to differentiate it from the Number input. I though it was part of this pr. Good to see though that you are catching this :)

packages/reactapp/src/NumberForm.js
antictor: I was getting an issue in Storybook in which it wasn't setting the field default value which seemed to be something to do with Storybook and not formik or my code so it was easiest to put an example in the test app.

PR #297, opened 2019-09-26
Fix button issues

m7kvqbe1:
## Overview

Identified and resolved numerous issues with Button component.

## Reason

While developing the Dialog and Modal components I identified changes fixes that needed to be made to the Button component.

## Work carried out

- [x] Upgrade imports and exports to be inline with new style (+ export interface)
- [x] Fix an issue that would prevent button icons being vertically aligned correctly
- [x] Fix issues with dynamic classes (`rn-btn--undefined` no longer appears)
- [x] Extend the ComponentWithClass base type
- [x] Fix issue where incorrect colour shade was used for `--danger` modifier
- [x] Conditionally render icon markup

## Developer notes

I've tested in storybook in order try to ensure there are no breaking changes as a result of these works.


packages/react-component-library/src/components/Modal/Modal.stories.tsx
```diff
@@ -0,0 +1,36 @@
+import { storiesOf } from '@storybook/react'
```
antictor: Why is this here? Doesn't belong in this commit

packages/react-component-library/src/components/Button/Button.tsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react'
```
antictor, line 1: Doesn't belong in this commit

packages/css-framework/src/components/_button.scss
```diff
@@ -74,6 +74,8 @@ $btn-focus-width: 0.2rem;
 }
 
 .rn-btn__icon {
+  display: inline-block;
```
antictor: I've also used flexbox to do the same in places as seems better

packages/react-component-library/src/components/Button/Button.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react'
 
-export interface ButtonProps {
+export interface ButtonProps extends ComponentWithClass {
   children?: string
   className?: string
```
antictor: Remove children and classname from here as you get them from ComponentWithClass

packages/react-component-library/src/components/Button/index.tsx
```diff
@@ -1,43 +1 @@
-import React from 'react'
```
antictor, line 1: merge this commit with earlier commit for button export change and the button.tsx should be in there too

antictor (commented):
The intention to have a commit history that breaks down the changes into small parts (telling the story) is good but there is still a mix up in some commits so the commit comment doesn't match the commit content

packages/react-component-library/src/components/Modal/Modal.stories.tsx
m7kvqbe1: It's marked outdated - I accidentally included that file in that commit and then it was cherry picked over to this branch. I have since removed the offending change with an additional commit on this branch (https://github.com/Royal-Navy/standards-toolkit/pull/297/commits/db2b576f8f4a08ef5670c123862bc47a8e2d0f27)

m7kvqbe1:
@antictor I've now interactive rebased this. I've left commits for changes associated with feedback.

antictor (approved):

PR #311, opened 2019-10-01
Forms documentation

antictor:
Adds a new section to the docs site dealing with forms. Describes how to use forms with Formik or standalone.


packages/docs-site/src/library/pages/forms.md
```diff
@@ -0,0 +1,126 @@
...
+    >
+      <Form>
+        <TextInput label="Colour" name="colour" />
+      </Form>
+    </form>
+  )
+}
+```
+
+Note: because the name of the object exposed by the component library conflicts with the name of the object exported by Formik, you must alias it to another name.
```
m7kvqbe1, line 120: Hmm yeah - you are right, this is annoying

m7kvqbe1 (approved):

packages/docs-site/src/library/pages/forms.md
antictor: So do we want to change that or just live with it? No one is using it yet... though it would be a breaking change

Issue #302, opened 2019-09-27
Fixup ESLint warnings


m7kvqbe1:
There appear to be a large number of ESLint warnings:

<img width="1422" alt="Screenshot 2019-09-27 at 12 47 55" src="https://user-images.githubusercontent.com/48086589/65767210-1a223600-e125-11e9-9c7a-d69f51d60702.png">

It may be that our lint config needs a little bit of tweaking.

Most of them appear to be regarding console.log (Storybook stories) and explicit any type usage.

antictor:
The thing is we want to be warned about 'any' usage as it encourages us to fix it, but there are times we need it

PR #317, opened 2019-10-03
Remove console.log lint warnings

thyhjwb6:
## Related issue

#302 

## Overview
Removing unnecessary use of `console.log` in Storybook stories and tests.

## Reason
To reduce the number of lint warnings.

## Work carried out
- [x] Removed `console.log` from Storybook stories
- [x] Removed `console.log` from tests

## Screenshot
Screenshot 2019-10-03 at 12 24 18

### DIFF
--- a/packages/react-component-library/src/components/Dialog/Dialog.stories.tsx
+++ b/packages/react-component-library/src/components/Dialog/Dialog.stories.tsx
@@ -1,3 +1,4 @@
+import { action } from '@storybook/addon-actions'
 import { storiesOf } from '@storybook/react'
 import React from 'react'
 
@@ -10,12 +11,8 @@ stories.add('Default', () => {
     <Dialog
       title="Dialog Title"
       description="Dialog description."
-      onConfirm={event => {
-        console.log('onConfirm')
-      }}
-      onCancel={event => {
-        console.log('onCancel')
-      }}
+      onConfirm={action('onConfirm')}
+      onCancel={action('onCancel')}
       isOpen
     />
   )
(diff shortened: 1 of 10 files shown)


packages/react-component-library/src/components/Dialog/Dialog.stories.tsx
```diff
@@ -1,5 +1,6 @@
 import { storiesOf } from '@storybook/react'
 import React from 'react'
+import { action } from '@storybook/addon-actions'
```
antictor: sort order of imports :p

packages/react-component-library/src/components/Dialog/Dialog.stories.tsx
```diff
@@ -26,12 +23,8 @@ stories.add('Danger', () => {
     <Dialog
       title="Dialog Title"
       description="Dialog description."
-      onConfirm={event => {
-        console.log('onConfirm')
```
antictor, line 30: I started off using action then switched to log because actions didn't show up but that is likely because it wasn't configured right and I was lazy :) 👍

packages/react-component-library/src/components/NumberInput/NumberInput.stories.tsx
```diff
@@ -18,24 +19,16 @@ const initialValues: Data = {
...
-const onCancel = () => {
-  console.log('Cancel')
-}
-
 const FormikNumberInput = withFormik(NumberInput)
 
 stories.add('Vanilla', () => (
   <div style={{ margin: 50 }}>
-    <Formik initialValues={initialValues} onSubmit={onSubmit}>
+    <Formik initialValues={initialValues} onSubmit={action('Form submit')}>
```
antictor: maybe

`{ (data) => action(data) )` ?

antictor (commented):
Just a few minor comments, nothing really.

packages/react-component-library/src/components/NumberInput/NumberInput.stories.tsx
thyhjwb6: This outputs `data` as expected:

Screenshot 2019-10-03 at 13 49 57

antictor (approved):

PR #322, opened 2019-10-04
Declare jest spy for blur assertion

thyhjwb6:
## Related issue
#302 

## Overview
Added specific type for Jest spy, rather than using `any` for event.

## Reason
Addresses Lint warning for `any` type used.

## Work carried out
- [x] Use Jest spy type

### DIFF
--- a/packages/react-component-library/src/components/Button/Button.test.tsx
+++ b/packages/react-component-library/src/components/Button/Button.test.tsx
@@ -14,16 +14,15 @@ describe('Button', () => {
     })
 
     describe('and a user clicks on the button', () => {
-      let event: any
+      let blurSpy: jest.SpyInstance
 
       beforeEach(() => {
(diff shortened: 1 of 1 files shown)


m7kvqbe1 (approved):

packages/react-component-library/src/components/Button/Button.test.tsx
```diff
@@ -14,24 +14,23 @@ describe('Button', () => {
...
-      let event: any
+      let blurSpy: jest.SpyInstance
 
       beforeEach(() => {
-        event = {
+        blurSpy = jest.fn()
+        component.simulate('click', {
           currentTarget: {
-            blur: jest.fn(),
+            blur: blurSpy,
```
antictor, line 23: That new?

antictor (approved):

PR #325, opened 2019-10-08
Add support for errors to native form components

antictor:
`withFormik` displayed errors outside of a form component, but that means it stripped away the ability to pass an error message into a component and have it displayed. Native form components need to be able to display errors so users can use them in the upcoming redux form components can also take advantage of it without duplicating code.

### DIFF
--- a/packages/react-component-library/src/enhancers/withFormik.tsx
+++ b/packages/react-component-library/src/enhancers/withFormik.tsx
@@ -20,6 +20,7 @@ const withFormik = (FormComponent: React.FC<any>) => ({
   const formComponentClassNames = classNames(className, {
     'is-invalid': hasError,
   })
+  const errorMessage = hasError ? errors[field.name] : undefined
 
   return (
     <>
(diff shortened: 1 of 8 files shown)


packages/react-component-library/src/enhancers/withFormik.tsx
```diff
@@ -20,19 +20,16 @@ const withFormik = (FormComponent: React.FC<any>) => ({
   const formComponentClassNames = classNames(className, {
     'is-invalid': hasError,
   })
+  const errorMessage = hasError ? errors[field.name] : undefined
```
thyhjwb6, line 23: Some Lodash `get` could potentially tidy this up. Have we made a decision not to use Lodash?

thyhjwb6 (commented):
This looks good. I cannot see corresponding tests for the component updates ie scenarios with and without errors. Is this okay?

packages/react-component-library/src/enhancers/withFormik.tsx
antictor: Not at all, lodash is a good idea.

antictor:
The point about the tests is a valid one. I'll make an effort to give proper test coverage

Issue #301, opened 2019-09-27
Convert old Enzyme tests to RTL


m7kvqbe1:
There appear to be some legacy Enzyme component tests leftover in the react-component-library:

- Button
- Nav

These need converting to react-testing-library (https://github.com/testing-library/react-testing-library) tests.

thyhjwb6:
There is a PR to remove the final dependencies and then this will be done once merged.

PR #326, opened 2019-10-08
Refactor Nav tests to use RTL

thyhjwb6:
## Related issue

#301 

## Overview

Refactored the Nav tests to use React Testing Library rather than Enzyme.

## Reason

The team has been migrating away from Enzyme and this is one of two remaining usages of Enzyme.

## Work carried out

- [x] Refactored tests
- [x] Use of test cases with `it.each`

## Developer notes

Note the usage of `it.each` to quickly test multiple variations.
Enzyme and related dependencies will be removed in another PR.


packages/react-component-library/src/components/Nav/Nav.test.tsx
```diff
@@ -1,132 +1,172 @@
...
-        },
-        {
-          href: 'http://testurl.test',
-          label: 'Components',
-        },
+      wrapper = render(<Nav navItems={navItemsMock} />)
+    })
+
+    it('should default the orientation to vertical', () => {
+      expect(wrapper.getByTestId('nav').classList).toContain('rn-nav--vertical')
```
antictor, line 30: I prefer this way of checking class names compared to the other pr.

antictor (approved):

PR #327, opened 2019-10-08
CSS Framework Spacing Rework

hxltrhuxze:
## Overview

This PR aims to rework the CSS Framework's Spacing values. The NELSON Toolkit supplies pre-determined spacing values to maintain consistency across all our applications and to ensure the designs from the Sketch library correlate with our code output. At the lower end of this Spacing curve, the gaps between the values were too large to allow for finesse when designing/building components.

| Updated Scale | Current Scale | Computed Value* |
|----------------|---------------|-------------------|
| 0                         | 0                        | `0`                         |
| 1                         | -                         | `2px`                     |
| 2                         | 1                        | `4px`                     |
| 3                        | -                         | `6px`                     |
| 4                        | 2                        | `8px`                     |
| 5                        | -                         | `10px`                   |
| 6                        | 3                        | `12px`                   |
| 7                        | -                         | `14px`                   |
| 8                        | 4                        | `16px`                   |
| 9                        | -                         | `18px`                   |
| 10                      | 5                        | `20px`                   |
| 11                       | 6                        |  `24px`                 |
| 12                      | 8                         | `32px`                  |
| 13                      | 10                       | `40px`                  |
| 14                      | 12                       | `48px`                 |
| 15                      | 16                       | `64px`                 |
| 16                      | 20                      | `80px`                 |
| 17                      | 24                       | `96px`                 |
| 18                      | -                         | `112px`                |
| 19                      | 18                       | `128px`                |
| 20                     | -                         | `144px`                |
| px                     | px                       | `1px`                     |
| full                    | full                      | `100%`                 |

\* Note: As the Spacing Scale is sized in `rems`, this is the assumed value based on an html `font-size` of `100%` (`16px` default `font-size` for Browsers).

The original Spacing Scale had a value range of `0 - 32`. As you approach the higher end of the scale, the gap between each Scale Value increased. This was used to demonstrate that the Computed Values between each of the Scale Items was increasing in size, however this led to confusion:
```
// Valid
spacing(6);

// Invalid
spacing(7);

// Valid
spacing(8);
```
To combat this, the new Spacing Scale is entirely linear in its design - every value between 1 & 20 is assigned to a spacing size. This should make it more intuitive to use, without having to second guess the implementation.

## To-do List

- [x] Update the Spacing Scale
- [x] Update Sketch Library Spacing Symbols to include the new values
- [x] Rewrite Docs for Spacing
- [x] Update **all** instances of the `spacing()` mixin
    - [x] Component Library
    - [x] Docs Site
    - [x] (Misc. areas not covered by the above)
- [x] Update **all** instances of the `spacing` helpers (e.g. `h_mt-4`)

## Developer notes

This PR should be considered a breaking change as we are changing the Spacing API for the CSS Framework.


m7kvqbe1 (approved):
We should try and write the commit messages in an imperative style.

So instead of "Fixed something that does this" it should be "Fix something that does this".

Looks like there is a conflict that needs resolving too (now that the content changes have been merged).

Issue #221, opened 2019-08-15
Tooltip component


ghost:
## Description

A small floating window that provides supplemental information to the user about a particular page element. It is activated by hovering over, or clicking on a separate trigger element.

## Component classification

Is this component an atom, molecule or organism?
- [x] Atom
- [ ] Molecule
- [ ] Organism

## Details / user needs

Users sometimes need additional explanation or information about an element on a page, such as a technical term or legend.

Tooltips should be activated by clicking on or hovering over a trigger element. Be aware that hover is not an option for touch screen devices.

Some design considerations: 
- The tooltip window should include an arrow or indicator that points to the trigger element.
- The dimensions of the tooltip window should be flexible to suit the content. However it should still be 'small', so appropriate guidance needs to be provided.
- We may want to have a transparent screen overlay when a tooltip is active.

## Definition of done
- [ ] The Sketch file is updated with the component
- [ ] There are screenshots of all component variations in this ticket
- [ ] All states are included if applicable (default, hover, active, focus)
- [ ] A design documentation PR is linked from this ticket
- [ ] All questions laid out in this ticket have been resolved 
- [ ] A pull request for a component must contain a fully tested, documented and finished component in a state it can be shared with partner developers

hxltrhuxze:
# Tooltip Component
The tooltip component is a simple component that can either have a single paragraph of text or a title and paragraph.
<img width="436" alt="Screenshot 2019-08-27 at 16 17 43" src="https://user-images.githubusercontent.com/48090803/63784218-380e3800-c8e6-11e9-9fef-99cf806d6d82.png">

<img width="467" alt="Screenshot 2019-09-12 at 12 25 31" src="https://user-images.githubusercontent.com/48090803/64780498-6fbce700-d558-11e9-9303-affa060a90e3.png">

### Interactions
When hovering on a Tooltip Trigger, the tooltip will fade in towards the trigger. For example, if the tooltip is situated above the Trigger, hovering on the trigger will cause the tooltip to fade from `0%` opacity to `100%` opacity, and transition downwards, towards the trigger.

### Abstract link
https://share.goabstract.com/b8bdd9f7-b588-4a3f-9c09-a3fa996963aa?sha=88831705e7ef8db1cb8119e01d2bcabfdc03bd24

ghost:
These probably don't affect sign-off, but:
- Will it ever be a different colour (or only if we do dark mode)?
[...]

PR #328, opened 2019-10-08
Tooltip component

antictor:
## Related issue
#221 

New React Tooltip component.

A simple balloon like component that can be used to display a small piece of information associated with something.

Screenshot 2019-10-08 at 17 29 52

Screenshot 2019-10-08 at 17 29 56

Simulator Screen Shot - iPhone 11 Pro Max - 2019-10-08 at 17 29 23


antictor:
Code ready for review, waiting for updated design documentation

packages/css-framework/src/components/_tooltip.scss
```diff
@@ -0,0 +1,117 @@
...
+  position: absolute;
+}
+
+.rn-tooltip__message {
+  font-size: font-size(s);
+  color: color(neutral, 100);
+}
+
+.rn-tooltip__content {
+  @include z-index(modal);
```
m7kvqbe1, line 18: Can you have tooltips within a dialog / modal still with this z-index?

packages/react-component-library/src/components/Tooltip/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+    left,
+    right,
+    top,
+    width,
+  }
+
+  return (
+    <div
+      className={`rn-tooltip
+      ${position ? `rn-tooltip--${position}` : ''}
```
m7kvqbe1: Is it cleaner to stick to using the classNames utility for these conditional classes?

m7kvqbe1 (approved):
Looks good - just a couple of comments.

packages/react-component-library/src/components/Tooltip/index.tsx
antictor: I think it would be even better to set a default for position and use classNames to build this to be consistent.

PR #329, opened 2019-10-10
Refactor Button tests to not use Enzyme

thyhjwb6:
## Related issue

#301

## Overview

Refactored the Button tests to use React Testing Library rather than Enzyme.

## Reason

The team has been migrating away from Enzyme and this is one of two remaining usages of Enzyme.

## Work carried out

- [x] Refactored tests
- [x] Added test cases for variant, type and size

## Developer notes

Note the usage of `it.each` to quickly test multiple variations.
Enzyme and related dependencies will be removed in another PR.

### DIFF
--- a/packages/react-component-library/src/components/Button/Button.test.tsx
+++ b/packages/react-component-library/src/components/Button/Button.test.tsx
@@ -1,37 +1,93 @@
+import '@testing-library/jest-dom/extend-expect'
 import React, { FormEvent } from 'react'
-import { shallow, ShallowWrapper } from 'enzyme'
+import { render, RenderResult, fireEvent } from '@testing-library/react'
 
 import { Button } from './index'
 
 describe('Button', () => {
-  describe('Given a button is created with a listener for onClick', () => {
-    let component: ShallowWrapper
-    let onClick: (event: FormEvent<HTMLButtonElement>) => void
+  let wrapper: RenderResult
+  let onClickSpy: (event: FormEvent<HTMLButtonElement>) => void
+  let blurSpy: jest.SpyInstance
+  let button: HTMLElement
 
+  beforeEach(() => {
+    onClickSpy = jest.fn()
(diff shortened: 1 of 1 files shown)


packages/react-component-library/src/components/Button/Button.test.tsx
```diff
@@ -1,37 +1,92 @@
...
+      wrapper = render(<Button onClick={onClickSpy} type={type}>Click me</Button>)
+      button = wrapper.getByText('Click me').parentElement
+
+      expect(button).toHaveAttribute('type', expected)
+    })
+  })
+
+  describe('when the variant is specified', () => {
+    it.each`
+      variant        | expected
```
antictor, line 82: I like it because it follows the Gherkin idea of data for tests in tables. I would say it's less readable, partly because it's new

packages/react-component-library/src/components/Button/Button.test.tsx
```diff
@@ -1,37 +1,92 @@
...
+      ${'primary'}   | ${'rn-btn rn-btn--primary rn-btn--regular'}
+      ${'secondary'} | ${'rn-btn rn-btn--secondary rn-btn--regular'}
+      ${'tertiary'}  | ${'rn-btn rn-btn--tertiary rn-btn--regular'}
+    `('styles the button when the variant is $variant', ({ variant, expected }) => {
+      wrapper = render(<Button onClick={onClickSpy} variant={variant}>Click me</Button>)
+      button = wrapper.getByText('Click me').parentElement
+
+      expect(button.className).toEqual(expected)
```
antictor: Not sure if I'd do this or use a assertion to see if the component contains a class name specified instead of looking for an exact match. This could break if the order changes or additional class names are added. This is just supposed to test that the type affects a particular class name

antictor (approved):

m7kvqbe1 (approved):

PR #341, opened 2019-10-11
Refactor and cleanup storybook plugin

antictor:
When I tried to delete the Vue Component package is caused the Storybook State plugin it broke the lint script which was caused by an issue with cross dependencies and configuration in the package.

To address this the following was done:

- [x] Simplified the Webpack config and removed until config code
- [x] Simplified the lint config to the minimum required and re-implemented the associated dependencies
- [x] Removed some unused elements of the storybook plugin code, such as it's local circle scripts
- [x] Re-wrote the tests to use React test tools instead of enzyme
- [x] Fixed some outstanding lint errors


thyhjwb6 (approved):
Approved

I noticed some of the commit messages do not read as instructions, ie starting with a verb

m7kvqbe1 (approved):

PR #342, opened 2019-10-11
Remove the deprecated react component package

antictor:
The old React JS component library was moved to a deprecated folder in case we wanted to look at the old code. Time has passed and no one has needed to access it and the majority of the components have been re-written so it is time to remove it.


m7kvqbe1 (approved):

thyhjwb6 (approved):
👍

PR #350, opened 2019-10-14
Remove Vue Components

antictor:
The Vue components are now very stale and not in sync with the css styles so need to be removed and will be re-created when the need arises.


m7kvqbe1 (approved):
Shall we request to have the package removed from the registry to prevent accidental usage / confusion? It's not being and shouldn't be being consumed by anyone.

antictor:
You can't get it removed from the registry.. so it just sits there I'm afraid, like so many others

m7kvqbe1:
> You can't get it removed from the registry.. so it just sits there I'm afraid, like so many others

You can - i've done it before with other packages. You just have to send them a support ticket.

m7kvqbe1:
https://docs.npmjs.com/unpublishing-packages-from-the-registry

PR #363, opened 2019-10-16
[WIP] Standards Toolkit 2.0.0

hxltrhuxze:
## Overview

This PR contains all updates for `v2.0.0` of the Standards Toolkit.

## Reason

As we're bumping the Toolkit to `v2.0.0`, this PR will contain all breaking changes that cannot be merged into `v1.x.x`.

#### SCSS Framework changes
- [x] Changed the Spacing API to be linear in scale
- [x] Changed the Typography API to reduce the base font size
- [x] Adjusted the Body font-size scale in order to make the Sketch designs & Code match closer.
- [x] Changed all references of Helper to Utility to be more consistent with the function of these classes
- [x] Converted all SCSS variables to strings
- [x] Added more colours to the tookit
- [x] Ported the CSS Framework to CSS modules
- [x] Added v2 migration docs
- [x] Port over new components to v2


packages/css-framework/src/abstracts/functions/_index.scss
```diff
@@ -0,0 +1,6 @@
+@forward "color";
+@forward "map-deep-get";
+@forward "non-destructive-map-merge";
+@forward "spacing";
+@forward "typography";
+@forward "map-get-default";
```
m7kvqbe1: Can we have a newline saved to the end of this file? There are a few others like this I think.

There should be a plugin for your IDE that respects the .editorconfig (https://editorconfig.org/) file at the repo root.

packages/css-framework/src/components/_carditem.scss
```diff
@@ -1,6 +1,6 @@
 // .rn-carditem {
```
m7kvqbe1: Do we need to keep these commented out styles? Maybe we could remove these?

packages/css-framework/src/components/_icon.scss
```diff
@@ -1,16 +1,16 @@
 // .svg-icon {
```
m7kvqbe1: Redundant styles?

packages/css-framework/src/components/_listitem.scss
```diff
@@ -2,7 +2,7 @@
 //   // border-bottom: 1px solid color(40);
```
m7kvqbe1: Redundant styles?

packages/docs-site/src/components/presenters/Masthead/index.js
```diff
@@ -26,6 +25,15 @@ const MastHead = ({ navItems }) => {
 
   return (
     <div className="masthead">
+      <Link
```
m7kvqbe1, line 28: Looks good for now - maybe in the future we could handle this with a flag or some config?

packages/css-framework/src/components/_listitem.scss
hxltrhuxze: No idea - the whole of `_listitem.scss` seems to be commented out. Presuming it's safe to drop?

packages/css-framework/src/components/_icon.scss
hxltrhuxze: Good spot - looks like this has been sat here for ~10 months so I'll go ahead and delete it.

packages/css-framework/src/components/_carditem.scss
hxltrhuxze: Yeah, I'll delete the card styles. We've only released the frame, so I reckon this is legacy stuff.

packages/css-framework/src/abstracts/functions/_index.scss
hxltrhuxze: Only installed the plugin halfway through this PR, so I've probably missed a few spots. Thanks though!

packages/docs-site/src/components/presenters/Masthead/index.js
hxltrhuxze: Yeah, probably a smart idea. I guess after a period of time of `v2` being live, it might be an idea to use this banner to point to the legacy `v1` docs? Just a thought

packages/css-framework/src/components/_listitem.scss
m7kvqbe1: Yeah - I think these are maybe styles leftover from the legacy library.

m7kvqbe1 (approved):
LGTM - I will admit I didn't read every single line of SASS for the sake of my own sanity.

Issue #338, opened 2019-10-11
NumberInput has no ability to display prefix (currency)


m7kvqbe1:
NumberInput has no ability to display prefix (currency).

The Coffee app mocks show a '£' symbol to the left of the input.

This is a new requirement.

hxltrhuxze:
I probably should've used a `pre-fix` with the £ symbol inside it:

<img width="745" alt="Screenshot 2019-10-17 at 17 19 48" src="https://user-images.githubusercontent.com/48090803/67027933-5cc2a700-f102-11e9-9fbb-dd8b02051a45.png">

Having said that, it might be useful to add the ability to have characters to the left of the input value

Issue #337, opened 2019-10-11
Switch styling needs changing to match other input fields


m7kvqbe1:
Switch styling needs changing to match other input fields.

The current height and width of this component means it looks out of place when used in conjunction with inputs in a form.

hxltrhuxze:
This is fixed in #366

Issue #336, opened 2019-10-11
Modal centre screen positioning needs fixing


m7kvqbe1:
Modal centre screen positioning needs fixing.

It isn't currently dynamically positioning so that it's always centre - no matter the viewport width and height.

Issue #334, opened 2019-10-11
Masthead has `text-decoration` underline on anchors


m7kvqbe1:
Masthead has `text-decoration` underline on anchors?

Maybe this base style is coming from the CSS framework itself?

I had to override the property within the Coffee app styles.

<img width="286" alt="Screenshot 2019-10-11 at 10 34 11" src="https://user-images.githubusercontent.com/48086589/66641481-c6c6e200-ec12-11e9-908c-b7cf9f645287.png">
<img width="59" alt="Screenshot 2019-10-11 at 10 34 03" src="https://user-images.githubusercontent.com/48086589/66641482-c6c6e200-ec12-11e9-94ce-c94f00c0f1d2.png">

PR #366, opened 2019-10-17
Demo app component tech-debt

hxltrhuxze:
#331 
#332 
#333 
#334 
#336 
#337 
#338 

## Overview

This PR addresses a number of issues with the component library. 

## Reason

Whilst building the demo application, a number of usage issues arose when using the components from the Standards Toolkit. This PR aims to address these issues

## Work carried out

- [x] Badge styles now match the Sketch Toolkit
- [x] Modal now sits in the centre of the screen at any resolution
- [x] Links now no longer underline themselves in the Masthead
- [x] The Masthead now has a box shadow on it
- [x] Changed look of the Switch component to the old style to improve readability
- [x] Changed the height of the large Switch component to match the height of Form Buttons.
- [x] Made buttons match the input height

#### TODO
- [ ] NumberInput has no ability to display prefix (currency) #338 
- [ ] You can’t place icons to the left of the button text #333


packages/css-framework/src/components/_scrollable-nav.scss
```diff
@@ -28,8 +28,13 @@ $btn-focus-width: 0.2rem;
   border-top: 4px solid transparent;
   border-bottom: 4px solid transparent;
   cursor: pointer;
+  position: relative;
```
antictor: Whats the relative for?

packages/css-framework/src/components/_scrollable-nav.scss
```diff
@@ -28,8 +28,13 @@ $btn-focus-width: 0.2rem;
...
 
-  &:focus {
+  &:hover {
+    border-bottom: 4px solid color(primary, 600);
```
antictor, line 33: There wasn't a border in the design as far as was aware but I did think it needed 'something'.

packages/css-framework/src/components/_scrollable-nav.scss
antictor: Also, make sure you merge from develop. There will be a conflict on this file I think.

packages/css-framework/src/components/_modal.scss
```diff
@@ -20,13 +20,16 @@
...
-  background: color(neutral, white);
-
-  top: calc(50% - 200px);
-  left: calc(50% - 300px);
-  position: fixed;
+  background: color(neutral, white);  
+  position: fixed; 
+  top: 50%;
+  left: 50%;
+  transform: translate3d(-50%, -50%, 0);
```
antictor, line 27: Nice

packages/css-framework/src/components/_badge.scss
```diff
@@ -1,28 +1,28 @@
 $badge-border-radius: 3px;
-$badge-font-weight: 700;
+$badge-font-weight: 600;
```
antictor, line 2: Do we have vars for weights?

packages/css-framework/src/components/_switch.scss
```diff
@@ -54,68 +45,44 @@ $elements: 5;
...
+  display: inline-block;
+  width: auto;
+  height: auto;
+  padding: spacing(2) spacing(4);
+  font-size: inherit;
+  color: color(neutral, 400);
+  border: 1px solid $border-color;
+  background-color: $background-color;
+  position: relative;
+  z-index: 0;
```
antictor, line 73: vars for z-index?

packages/css-framework/src/components/_switch.scss
```diff
@@ -1,7 +1,7 @@
 $border-radius: 2px;
 $background-color: color(neutral, 0);
 $border-color: color(neutral, 200);
-$active-background-color: color(neutral, white);
```
antictor, line 4: while you are here... we talked about namespacing vars the other day. As it stands vars get put into a global scope so vars from one file can break another.

packages/css-framework/src/components/_switch.scss
hxltrhuxze: Good spot.

For this I've made the `.rn-switch` component `position: relative`. This will automatically set the `z-index` to `auto` for this class. By doing this, it allows us to scope the stacking order of the `.rn-switch__option` to within the `.rn-switch`. If we'd used variables here, instead of `positon: relative` on the `.rn-switch`, then the `.rn-switch__option` would have displayed higher than the switch itself.

packages/css-framework/src/components/_badge.scss
hxltrhuxze: We have font sizes, but not weights. Could potentially add this in the future, but feels like it's a bit overkill, considering the range is only `100 - 900`.

antictor (approved):

packages/css-framework/src/components/_switch.scss
hxltrhuxze: Yeah this was me just modifying the pre-existing variables here.

I'm not entirely sold on the idea of having variables _within_ a particular component file. There's definitely times where using a variable in this situation is beneficial, however I've seen a number of instances where the variable is only used once. This means you'd actually be changing the value _outside_ of its context within the class definition. I think we should avoid this as much as possible, and I definitely agree with you about namespacing these variables where we use them.

packages/css-framework/src/components/_scrollable-nav.scss
hxltrhuxze: good spot, removed.

packages/css-framework/src/components/_scrollable-nav.scss
hxltrhuxze: Yeah, when fiddling with it, that interaction felt the best. Having a fresh look at it in context helped.

Issue #286, opened 2019-09-23
Table Component


hxltrhuxze:
# Table Component

The Table component is used to display tabular data.

# Base Tables
From looking at the examples and use cases we have, I’ve split the following section into 2 distinct types of components. As an overview, the Data Table should be used when you want to display a large amount of data that have no to little data visualisation or interactivity associated. The Interactive Table however is much more versatile due to its ability to accept interact-able sub-components.

## Data Table
The Data Table is used to display large amounts of data at once - it’s most appropriately used for feeds and raw data. Each row can be hovered and clicked on, however because of this the table cells containing sub-components are read-only. The Data Table cells can only have a single line of text for each table cell - if a value overflows it should be truncated.

### Condensed Variation
<img width="809" alt="Screenshot 2019-09-23 at 16 31 38" src="https://user-images.githubusercontent.com/48090803/65446596-99b1bb80-de2c-11e9-929f-affa3d11d91c.png">
The condensed variation of the Data Table is used for large amounts of data that need to be displaced at once. It has the narrowest rows of any of the table layouts, ensuring as much information can fit into the available space as possible. 

### Expanded Variation
<img width="901" alt="Screenshot 2019-09-23 at 17 14 32" src="https://user-images.githubusercontent.com/48090803/65446617-a2a28d00-de2c-11e9-9c0f-c6074ac49054.png">
The expanded version of the Data Table has the ability to add richer single-line sub-components to table cells. One or more rows can also be selected with the left-aligned checkboxes.

## Interactive Table
<img width="852" alt="Screenshot 2019-09-23 at 17 21 45" src="https://user-images.githubusercontent.com/48090803/65446642-afbf7c00-de2c-11e9-890b-4b641dff5e9b.png">
The Interactive Table is an expanded table that can contain rich sub-components and interactivity. Each cell can be individually interacted with, allowing for a large range of behaviours to exist within the table. However, due to this the table rows cannot be hovered/clicked as a whole. Excluding visual appearance, this is probably the largest difference between the two Base Table designs. Content inside the table cells can also wrap multiple lines, allowing things like long form text to be displayed fully. The option to concatenate the content is also available for this table type.

## Backgrounds
<img width="873" alt="Screenshot 2019-09-23 at 17 02 11" src="https://user-images.githubusercontent.com/48090803/65446675-c5cd3c80-de2c-11e9-8175-0dd9709f07c2.png">
[...]

PR #368, opened 2019-10-18
Add sorting to table component

thyhjwb6:
## Related issue

#286

## Overview
Adds the ability to sort a table column by clicking the header cell.

## Reason
When analysing data it will be necessary to change the order to compare.

## Work carried out
- [x] Add sort
- [x] Update develop documentation
- [x] Update design documentation
- [x] Add sort icons

### Desktop
Screenshot 2019-10-31 at 07 49 35

### Mobile
Screenshot 2019-10-31 at 07 49 47

## Developer notes
- Take a close look at the SVG implementation


hxltrhuxze:
Really nice :) Having played with the live demo, I think there's possibly a couple of things we should fix to make it more usable.

1. Firstly, it's difficult to know that a column is sortable without clicking on the column header. Perhaps an icon like this could be used?

Screenshot 2019-10-29 at 15 46 14

2. Secondly, is there a way to deselect a table column _after_ it's been sorted? This may be fixed by point 1, as after two clicks on the table header we could revert the column back to its default and swap the icon back to the icon I've linked above.

packages/css-framework/src/components/_table.scss
```diff
@@ -1,13 +1,25 @@
...
+  table-layout: fixed;
   width: 100%;
   th {
     font-weight: bold;
     text-align: left;
     @include font-size(xs);
     text-transform: uppercase;
     padding: spacing(4) spacing(2);
+
+    &.sortable {
```
m7kvqbe1: Could we change this stateful class to `is-sortable`? Just for consistency.

packages/icon-library/src/assets/table/sort-ascending.svg
```diff
@@ -0,0 +1,5 @@
+<svg width="16" height="16" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 9 14">
```
m7kvqbe1, line 1: Could we not use these ones already in the library?

https://material.io/resources/icons/?search=down&icon=arrow_drop_down&style=round
https://material.io/resources/icons/?search=up&icon=arrow_drop_up&style=round

Maybe they aren't fit for purpose?

packages/react-component-library/src/components/Table/Column.tsx
```diff
@@ -1,10 +1,53 @@
...
+  IconSortDescending,
+  IconSortUnsorted,
+} from '@royalnavy/icon-library'
+
+import { ASCENDING, DESCENDING } from './constants'
 
 export interface ColumnProps {
   field: string
+  sortable?: boolean
+  onSortClick?: Function
```
m7kvqbe1: Is it worth defining the signature and being more explicit with this type?

m7kvqbe1 (approved):
Looks good to me - just a couple of small comments

packages/icon-library/src/assets/table/sort-ascending.svg
thyhjwb6: I originally used the icons from the existing library and through discussion with @hxltrhuxze we are using a different icon for `ascending`, `descending` and `unsorted`.

https://github.com/Royal-Navy/standards-toolkit/pull/368#issuecomment-547488735

PR #378, opened 2019-10-18
Remove Enzyme dependencies

thyhjwb6:
## Related issue

#301 

## Overview
Final step is to remove Enzyme related dependencies.

## Reason
These are no longer required.

## Work carried out
- [x] Remove `enzyme`
- [x] Remove `enzyme-to-json`
- [x] Remove `enzyme-adapter-react-16`


packages/react-component-library/jest/setupTests.ts
```diff
@@ -1,5 +1 @@
-import { configure } from 'enzyme'
-import EnzymeAdapter from 'enzyme-adapter-react-16'
 import 'babel-polyfill'
```
antictor, line 1: Do we even need this?

antictor (approved):
Just wondering if we can delete the test setup as do we need babel polyfill?

Might be worth a try before merging.

Don't forget to squash that fixup

packages/react-component-library/jest/setupTests.ts
antictor: Appears we do.

Issue #278, opened 2019-09-20
Cannot build on Windows


antictor:
Some of the commands used in the npm scripts for the project are bash commands. These don't work in CMD or Powershell

antictor:
PR Raised but blocked from getting approval and merging until Lerna fixes the vulnerability in one of it's dependencies.

PR #379, opened 2019-10-21
Support Windows developers

antictor:
## Related issue
#278 

Some small changes to the build processes and scripts to allow them to work on Windows. Most the changes involved replacing rm -rf with rimraf and run-s instead of &&


packages/icon-library/generate-exports.js
```diff
@@ -0,0 +1,32 @@
+const fs = require('fs')
```
m7kvqbe1: This is way less readable than my bash script...

https://github.com/Royal-Navy/standards-toolkit/pull/379/files#diff-d975a43e2deef42acab44c11f3da4fde

You can run the shell script on windows still??

Not sure I agree with this?

packages/react-component-library/package.json
```diff
@@ -32,8 +31,8 @@
     "prepare": "yarn build",
     "storybook:static": "build-storybook -c .storybook -o .static_storybook",
     "storybook": "start-storybook -p 6006",
-    "test:watch": "NODE_ENV=test jest --watch",
```
m7kvqbe1, line 35: Why have you removed the `NODE_ENV=test` setting this environment variable ensures that the correct transforms are applied as specified within the .babelrc file:

https://github.com/Royal-Navy/standards-toolkit/blob/develop/packages/react-component-library/.babelrc

```json
"env": {
    "test": {
      "plugins": [
        "@babel/plugin-transform-modules-commonjs"
      ]
    }
  }
```

Without this the tests will fail because Node cannot interpret ESM modules it needs them to be transformed to CommonJS.

packages/icon-library/generate-exports.js
m7kvqbe1, line 1: This is way less readable than my bash script...

https://github.com/Royal-Navy/standards-toolkit/pull/379/files#diff-d975a43e2deef42acab44c11f3da4fde

You can run the shell script on windows still??

Not sure I agree with this?

m7kvqbe1 (changes requested):

packages/icon-library/generate-exports.js
m7kvqbe1: Just use WSL?

packages/react-component-library/package.json
antictor: They still pass, though this may be because the babel transform is in the jest config.

I wonder if there is another way to do this without the NODE_ENV bit in the command line as Powershell and CMD don't support that syntax.. though I'll see if there is a way

packages/icon-library/generate-exports.js
antictor: WSL is overkill just for 1 script. Plus I used WSL and came away with the impression it's an all or nothing thing with it. The reason is that your ide or git tool will be running using native windows git but in wsl you are using the vm git. The upshot was that if I used vscode the githooks had to work on windows. Why not have a script that is OS agnostic instead of one that is specific to Bash?

Issue #219, opened 2019-08-15
Button group component


ghost:
## Description

A number of buttons that are physically joined together.

## Details / user needs

A series of grouped buttons, indicating to the user they perform related or similar actions. For example: zoom in | zoom out. The buttons should be physically connected.

It should be possible to label the button group.

This should follow the standard button conventions closely.

**Note that this is different to a switch were one or more buttons can be active.**

_Used in 2 exemplar projects_

## Variations

- Enabled, disabled, in focus, hover states 

## Questions

Outstanding questions for the Product Owner, Designers or Developers.

## Definition of done
- [x] The Sketch file is updated with the component
- [x] There are screenshots of all component variations in this ticket
- [x] All states are included if applicable (default, hover, active, focus)
- [x] A design documentation PR is linked from this ticket
- [x] All questions laid out in this ticket have been resolved 
- [ ] A pull request for a component must contain a fully tested, documented and finished component in a state it can be shared with partner developers

hxltrhuxze:
# Button Groups
Button groups connect secondary buttons together into a single, unseparated block. They can consist of a group of icons or labels, however these should not be mixed.

### States

<img width="349" alt="Screenshot 2019-09-20 at 12 48 12" src="https://user-images.githubusercontent.com/48090803/65324714-ed11d880-dba4-11e9-9d3b-bb5efab893cb.png">

The Button Group has 3 states: Default, Hover/Focused, and Disabled. There is no Active state for Button Groups as they function the same as regular buttons - you click for an action and it reverts to its default state. If it's required to maintain state for the selected option, the Switch component should be used instead.


### Abstract Link
https://share.goabstract.com/f14a16f5-12fa-43cb-86fb-405a21cb432b?sha=9cb86ee7d25366567fde55baef1330d320279a9a

ghost:
Is there a maximum number of buttons we would have in a button group? It feels like it would be helpful to others (specifically teams using this w/o a designer on the team) to guide them on that? Enforcing feels a bit strong but perhaps it’s appropriate if we feel strongly about it. Sounds like general agreement on 5 and that we'll capture that in the docs.

Would you ever use icons in this sort of button group?

ghost:
I think this is 'signed off' from my perspective provided thought has gone into the comments above (which it sounds like there has). It'd be good to document some answers/thoughts @hxltrhuxze

hxltrhuxze:
@xdobtqletrmn 

[...]

PR #399, opened 2019-10-29
ButtonGroup Component

antictor:
## Related issue
#219

The button group accepts a list of button item properties and renders a group of like styled buttons as a unit.

Oct-30-2019 10-10-31

### DIFF
--- a/packages/docs-site/src/library/pages/components/buttongroup.md
+++ b/packages/docs-site/src/library/pages/components/buttongroup.md
@@ -9,15 +9,18 @@ import DataTable from '../../../components/presenters/data-table'
 import CodeHighlighter from '../../../components/presenters/code-highlighter'
 import SketchWidget from '../../../components/presenters/sketch-widget'
 
+import ButtonGroupComponent from '../../images/components/button-group/component.svg'
+import ButtonGroupAnatomy from '../../images/components/button-group/anatomy.svg'
+
 
 # Overview
 
 A Button group is a useful component for grouping together buttons to allow a user to select an option.
 
-
+<ButtonGroupComponent />
 
 ## Usage
-The usage section is a bullet pointed list of scenarios the component should be used in.
+The Button Group acts as a wrapper for multiple buttons. It removes the spacing and border radius between adjacent buttons, making them sit flush next to each other. It is used to indicate a relationship between similar actions - pagination and toolbars are great examples of this.
 
 <TabSet>
   <Tab title="Design">
@@ -25,12 +28,18 @@ The usage section is a bullet pointed list of scenarios the component should be 
 <SketchWidget name="Button group" href="/standards-toolkit.sketch" />
 
 ### Anatomy
+The Button Group by itself doesn't have any visual elements. The Group simply removes the spacing between buttons and makes them sit flush beside each other.
 
+<ButtonGroupAnatomy />
+
+1. **Container** - wraps the component and removes border-radius and spacing between buttons.
+2. **Buttons** - the Button Components added to the Group.
 
 ### Sizing & Spacing
-
+The Button Group should always use the same size Button components. This allows the buttons to be positioned flush against each other. The Buttons themselves though can be any of the [available sizes](/components/buttons) outlined on the Button Docs page.
 
 ### Hierarchy & Placement
+The Button Group is used to collate similar actions together under one consistent grouping. It should only feature Secondary Buttons, as Primary Buttons should be standalone actions.
 
 </Tab>
 
--- a/packages/docs-site/src/library/pages/components/buttongroup.md
+++ b/packages/docs-site/src/library/pages/components/buttongroup.md
@@ -4,7 +4,7 @@ description: A collection of buttons commonly used to select an option
 header: true
 ---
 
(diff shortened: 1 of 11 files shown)


packages/react-component-library/src/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -0,0 +1,5 @@
+import React from 'react'
+
+export const ButtonGroup: React.FC = ({ children }) => (
```
thyhjwb6: Is there an opportunity to tighten this up with an interface as `children` will always be a collection of `Button`?

packages/docs-site/src/library/pages/components/buttongroup.md
```diff
@@ -0,0 +1,303 @@
+---
+title: Button Group
+description: A collection of buttons commonly used to select an option
+header: true
+---
+
+import { Button,ButtonGroup,  Icons, Tab, TabSet } from '@royalnavy/react-component-library'
```
thyhjwb6: Spacing could be tweaked here

packages/docs-site/src/library/pages/components/buttongroup.md
```diff
@@ -0,0 +1,303 @@
...
+### Sizing & Spacing
+
+
+### Hierarchy & Placement
+
+</Tab>
+
+<Tab title="Develop">
+
+A ButtonGroup is a simple wrapper component that is designed to group buttons together into a common component. The ButtonGroup component itself accept no properties, only content which should consist of a collection of buttons.
```
thyhjwb6: `accept` 👉 `accepts`

thyhjwb6 (commented):
A few small comments

hxltrhuxze:
Several feedback points:
1. The button border needs to be from the button itself. This PR is adding a neutral border to the first child for some reason.
2. I'd prefer it if all the buttons inside the group were forced to be the same type. Having a mix of primary/danger/neutral is confusing.
3. what's going on with this button? We have dropdowns which != buttons, so I'm not sure why this button has a downwards icon here. Also, why is the icon blue?
Screenshot 2019-10-29 at 15 23 20

packages/docs-site/src/library/pages/components/buttongroup.md
antictor: This file is still being worked on by Ed.

antictor: Waiting for design documentation to be added

packages/react-component-library/src/components/ButtonGroup/ButtonGroup.tsx
antictor: Could do. Whats the syntax for that?

antictor: NVM if other comments are correct from design then I may have to write this a different way

antictor:
The dropdown icon was just to show you can have an icon and I had that icon already in the project. Could be any icon

To force the buttons to all be the same type may be restrictive. There may be a time when one of the buttons has to be red surely?

If not then that and the border thing will mean a re-think of the approach. Instead of a simple wrapper I'd have to pass a single size, type, variant and an array of labels/keys, and expect an onClick handler that would be triggered with the key of the button pressed.

Not saying it's wrong, just clarifying before I continue

hxltrhuxze:
> The dropdown icon was just to show you can have an icon and I had that icon already in the project. Could be any icon.
Fair point, let's not use that example in the docs though. Why is there space to the right of the close button, just out of curiosity? It should be equally spaced.

> To force the buttons to all be the same type may be restrictive. There may be a time when one of the buttons has to be red surely?

Maybe it is too restrictive. However, I do feel that from a UI perspective that having a split between multiple types of actions prevents a potential mess resulting from this component. 

Primary buttons are your main action and should be standalone - they take prominence in the interface and are _the_ thing we want the user to focus on. Button groups are there to supply complimentary actions under one theme. We're always aiming to keep the number of primary buttons per component/page low, as when we start using more of them, they lose their importance. 

I can't help but think having [Neutral][Neutral][Danger][Neutral] is a really bad UX practice, as you have a destructive action buried in between regular buttons. There's no hierarchy of responsibility there.

Regardless, I think we should force button groups to be purely Secondary buttons. The buttons in the groups are all of equal prominence.

> If not then that and the border thing will mean a re-think of the approach. Instead of a simple wrapper I'd have to pass a single size, type, variant and an array of labels/keys, and expect an onClick handler that would be triggered with the key of the button pressed.

You'd need to pass a size anyway as you can't mix multiple sized buttons inside a button group. The original design issue (#219) has the button example there. It's all using secondary buttons - there's no primary buttons in here. 

Either way, what we do to the hierarchy is irrelevant to the border issue. There's a neutral border being set on the internal buttons - why? If you wanted to make the button group accept any type of button, why is there a neutral border being set? It literally _isn't_ what the border is on any of the primary buttons, so it's going to look visually off regardless of which buttons are supplied.

antictor:
Oct-31-2019 10-43-06

m7kvqbe1 (approved):

PR #401, opened 2019-10-30
Add support for disabled buttons

antictor:
A small enhancement to the Button component that lets you mark it as 'disabled' so users cannot use it.

### DIFF
--- a/packages/css-framework/src/components/_button.scss
+++ b/packages/css-framework/src/components/_button.scss
@@ -99,3 +99,11 @@ $btn-focus-width: 0.2rem;
 fieldset:disabled a.rn-btn {
   pointer-events: none;
 }
+
+.rn-btn.rn-btn--disabled,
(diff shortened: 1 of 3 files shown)


m7kvqbe1 (approved):

packages/css-framework/src/components/_button.scss
```diff
@@ -99,3 +99,11 @@ $btn-focus-width: 0.2rem;
 fieldset:disabled a.rn-btn {
   pointer-events: none;
 }
+
+.rn-btn.rn-btn--disabled,
+.rn-btn.rn-btn--disabled:hover,
+.rn-btn.rn-btn--disabled:focus {
+  background: color(neutral, 000);
```
hxltrhuxze, line 106: We appear to have `.rn-btn.disabled` and `.rn-btn.rn-btn--disabled`. Do we need both? @antictor

Issue #223, opened 2019-08-16
Notification component


ghost:
## Description

The notification component displays important messages that are relevant to the user. These messages may be triggered while the user is in the App, or whilst they are away.

## Component classification

Is this component an atom, molecule or organism?
- [ ] Atom
- [ ] Molecule
- [x] Organism

## Details / user needs

We need some way of highlighting to users that there are important messages they should attend to, but without being highly intrusive in the interface,

Some design considerations:
- The trigger element should indicate whether there is a notification present, and the number of notifications
- The component should include a trigger element, such as an icon or button
- The layout of notifications themselves should be flexible. Each Application is likely to have very different needs
- Each notification should contain a brief description, and a call-to-action, or ability to dismiss
- Notifications need to be positioned prominently on the page so users don't miss them

_Used in 2x exemplar projects_

## Questions

- How will the user access long list of notifications (10+)?
- How will the user access an archive of notifications - do we need an archive? This mighr be application specific.

## Definition of done
- [ ] The Sketch file is updated with the component
- [ ] There are screenshots of all component variations in this ticket
- [ ] All states are included if applicable (default, hover, active, focus)
- [ ] A design documentation PR is linked from this ticket
- [ ] All questions laid out in this ticket have been resolved 
- [ ] A pull request for a component must contain a fully tested, documented and finished component in a state it can be shared with partner developers

hxltrhuxze:
# Notification Sheet

<img width="726" alt="Screenshot 2019-08-27 at 17 26 18" src="https://user-images.githubusercontent.com/48090803/63789704-cc30cd00-c8ef-11e9-8cf0-c3a8eec37dac.png">

## Answers to questions
- How will the user access long list of notifications (10+)?
The notification sheet is _not_ the notification page. That is a separate page that contains all the notifications, included expanded descriptions.
- How will the user access an archive of notifications
See above.

### Abstract Link
https://share.goabstract.com/e13d7d86-745f-4c19-bc04-d57542754099?sha=c7226d3e551d7d83470dbb287d92cbc868ae406d

09txzS1qs:
@hxltrhuxze the scope of this component is unclear from the screenshots and description - happy with the simple bell icon, but to what extent with the notification sheet be standardised?

hxltrhuxze:
@09txzS1qs In my mind, the notification sheet should be standardised into having:
- Icon (either avatar, or appropriate action icon)
- Title
- Date
- Description (optional)

[...]

PR #406, opened 2019-11-02
Update notification panel content

thyhjwb6:
## Related issue

#223 

## Overview
Updated the content of the `NotificationPanel` to be be more specific so that it matches the design. The `NotificationPanel` now accepts a collection of notifications as to be presented as a list.

## Reason
Currently the `NotificationPanel` takes arbitrary content but this needs constraining further.

## Work carried out
- [x] Add notification list component with list item
- [x] Integrate changes
- [x] Update developer documentation
- [ ] Update design documentation

## Screenshot
Screenshot 2019-11-07 at 16 27 59

## Developer notes
- There is currently a bug in the `Sidebar` which means the notifications are not displayed
- Use the `Masthead` `With navigation` Storybook story for an example
- The `rn-notifications-item__indicator` could be reused instead of [a SVG circle](https://github.com/Royal-Navy/standards-toolkit/blob/24fb581c99324d4fb2b74f02c1721dff4766fb3b/packages/react-component-library/src/fragments/NotificationPanel/index.tsx#L119). This will be done in a follow up PR


hxltrhuxze:
Looking great!

Just a couple of styling tweaks:
- Can we change the title to have `line-height: 1.3`? I realise it's set to `~1.2` in the Sketch document, however it looks more cramped in the browser (most likely due to subtle differences in font rendering between Sketch and CSS)
- When you hover on an item, can we make the Avatar and the white items (Name & Review) turn blue? At the moment, only the Avatar changes when it is hovered on which can be jarring as it implies clicking the avatar does something different to clicking the notification item. The Notification item by itself doesn't have a hover state so it's tricky trying to work out where you should click.
- When hovering on the "View all notifications" link, can we remove the underline and instead change the colour of the link & the right chevron to blue `color(primary, 300)`? And then can we also animate the chevron to the right by `~4px`?

packages/docs-site/src/library/pages/components/notifications.md
```diff
@@ -0,0 +1,125 @@
+---
+title: Notifications
+description: todo
```
m7kvqbe1: This needs populating still.

packages/react-component-library/src/fragments/NotificationPanel/NotificationPanel.tsx
```diff
@@ -0,0 +1,162 @@
...
+}) => {
+  const notificationArrowPosition =
+    notificationPlacement === 'right' ? 'left_bottom' : 'top_right'
+
+  const [showNotifications, setShowNotifications] = useState(false)
+  const [notificationPosition, setNotificationPosition] = useState()
+  const node = useRef()
+
+  function documentClick(event: Event) {
+    // workaround for undefined error in typescript
```
m7kvqbe1, line 59: Anyone know what this is about? Is this to deal with browser APIs missing in node when doing SSR?

packages/react-component-library/src/fragments/NotificationPanel/NotificationPanel.tsx
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    if (current.contains(event.target)) {
+      return
+    }
+
+    setShowNotifications(false)
+  }
+
+  function toggleNotifications(event: React.SyntheticEvent) {
```
m7kvqbe1, line 71: I know this isn't your code but maybe we could refactor some of this into a custom hook?

m7kvqbe1 (approved):
Looks good - just a couple of comments - mainly about some of the pre-existing code.

packages/react-component-library/src/fragments/NotificationPanel/NotificationPanel.tsx
thyhjwb6: I think that is a good idea which I would like to do but would prefer to keep this change focused and do a follow up PR.

PR #409, opened 2019-11-07
Converting CSS Variables to Strings

hxltrhuxze:
## Overview
This PR is a manual find-and-replace effort to convert all variables used in the SCSS framework to strings. This bug was noticed when trying to create `border-colour` helpers - the colours `white` and `black` that we were using for keys in our Colour map are protected objects in SCSS, so you can't interpolate them into a class name (Sass moans thinking you're trying to literally use a colour, so breaks. yay).

## Work carried out

- [x] Converted _all_ files to use strings for their variables.
- [x] Updated docs instances to have strings around their variables.

## Developer notes

Yes - massive amount of commits, I know.

#### Couldn't you have broken this up?
I needed to make sure the whole thing compiled first and actually worked using strings. I also had to manually find-and-replace all these instances - there's no easy upgrade path here. The safest way to do this was searching by each of the functions in turn (e.g. `spacing(` as the search item) and replacing their content with the correct strings.

#### What has actually changed in this PR?
Every instance where a SCSS function call has happened (e.g. `color(neutral, 100)`) has been replaced by strings inside double quotes (`color("neutral", "100")`). This prevents Sass from panicking when it thinks you're trying to use a protected object. This is all that is changed in this PR. Yes there are a lot of commits, but it's simply adding a _lot_ of quotes.


thyhjwb6 (approved):
The change is good. I think it could be a single commit 🤷‍♂

PR #418, opened 2019-11-11
Refactor notification panel to custom hook

thyhjwb6:
## Related issue
#223 

## Overview
Refactor of `NotificationPanel`

## Reason
Having done the recent notifications work, this was a good opportunity to refactor the `NotificationPanel`.

## Work carried out
- [x] Refactor
- [x] Update tests

## Screenshot

## Developer notes
Note that the `fixup` commit was to address issues with Code Climate. This will be squashed once approved.


packages/react-component-library/src/hooks/useDocumentClick.ts
```diff
@@ -0,0 +1,27 @@
+import React, { useEffect } from 'react'
+
+// workaround for undefined error in typescript, node.current might not be defined
```
m7kvqbe1, line 3: Did you manage to figure out what this was about?

m7kvqbe1 (approved):
Looks lovely to me!

packages/react-component-library/src/hooks/useDocumentClick.ts
thyhjwb6: There is a TypeScript error when there is not an alternative for `node.current` being set. Whether this is the only option, I don't know.

PR #419, opened 2019-11-11
Refactor table sort order constants

thyhjwb6:
## Related issue
NA

## Overview
Refactoring `SORT_ORDER` constants to use an `enum` and therefore be cleaner when used.

## Reason
This is cleaner code.

## Work carried out
- [x] Refactor `SORT_ORDER`

## Screenshot
No visual difference.


m7kvqbe1 (approved):
yay for enum

PR #420, opened 2019-11-11
Colour Rework 

hxltrhuxze:
## Overview

This PR achieves two objectives:
1. Expands the colour palette to contain more shades, giving the new colours a more consistent grouping (supplementary colours).
2. Renames "Primary" to "Action" to prevent confusion between primary buttons etc

## Reason
The current colour palette is limited when using it for categorisation/data visualisation. This PR increases the current palette to include more hues. It also groups these hues into the new Supplementary colour palette - this also future-proofs us if we ever want to expand the colours in the future.

This PR also renames the "Primary" colour to "Action". This name is more appropriate for its use, and it also prevents the clash in primary buttons where you could have a `primary` value for both `variant` and `state`.

## Work carried out

- [x] Converted all CSS variables to strings as `black` and `white` are reserved objects, meaning they can't be interpolated into generated SCSS classes. Strings however can have any value we choose, so won't break the scss compiler.
- [x] Expanded the colour palette to contain new hues from `sup a` to `sup f`.
- [x] Renamed Primary to Action to better describe its usage.
- [x] Renamed all references of "Primary" colour to "Action", including documentation.


m7kvqbe1 (approved):
Looks good to me - just an idea for the future - maybe we could make some kind of 'cheatsheet' (https://devhints.io/vim) for the CSS framework?

Issue #348, opened 2019-10-14
Drawer


ghost:
## Description

A draw sliding out from the right hand side. 

## Component classification

This component is an atom.

## Details

- Animates open and close
- Some means of opening and closing the drawer
- Independently scrolling drawer content
- No fade for the rest of the screen
- Should stick over all other content
- Not on mobile
- Documentation explicit about differences to side navigation, etc.

## Questions

Outstanding questions for the Product Owner, Designers or Developers.

## Definition of done
- [ ] The Sketch file is updated with the component
- [ ] There are screenshots of all component variations in this ticket
- [ ] All states are included if applicable (default, hover, active, focus)
- [ ] A design documentation PR is linked from this ticket
- [ ] All questions laid out in this ticket have been resolved 
- [ ] A pull request for a component must contain a fully tested, documented and finished component in a state it can be shared with partner developers

hxltrhuxze:
The Drawer component is a simple panel that can be triggered to appear on the right hand side of the screen. This current version only contains the shell and close button.

The Drawer should animate from off screen and slide in to the right of the window.

<img width="324" alt="Screenshot 2019-11-04 at 15 49 59" src="https://user-images.githubusercontent.com/48090803/68137200-7e9f9480-ff1e-11e9-9f42-bc5fa1669f45.png">

ghost:
A couple of observations/questions:

Will we recommend:
- how far it should come out?
- how quickly it comes out
- how quickly it goes back when x is clicked?
and do the devs need to know that stuff to implement it?

Will a user be able to scroll vertically within the card. Here's an example of what I mean:
- Go to the Standards Tactical Roadmap
- Click on the title of one of our cards
- Scroll up and down inside the thing on the right (I'm assuming that's a drawer?)

hxltrhuxze:
#### how far it should come out?
This is tricky - I'd like to avoid dictating sizes for this as there may be multiple use cases that require different sizes. For a rule of thumb though, I think the Drawer should never occupy more than `33.33%` of the width of the page.

#### how quickly it comes out?
We try to keep all animations to a maximum of `0.3s`.

#### how quickly it goes back when x is clicked?
Should also be `0.3s` - it's simply reversing the animation that brought it out.

#### Do the devs need to know that stuff to implement it?
We have basic animations already in the project and this should follow suit. This'll be something we'll have to work with the devs on though, as it'll take a working example to test what sort of animation timing curves we use.

[...]

PR #424, opened 2019-11-12
Drawer Component

m7kvqbe1:
## Related issue

https://github.com/Royal-Navy/standards-toolkit/issues/348

## Overview

Implement, test and document the Drawer component.

## Reason

>The Drawer component is a simple panel that can be triggered to appear on the right hand side of the screen. This current version only contains the shell and close button.
>
>The Drawer should animate from off screen and slide in to the right of the window.

## Work carried out

- [x] Implement base functionality
- [x] Style the component
- [x] Add automated tests
- [x] Add design documentation


packages/react-component-library/src/components/Drawer/Drawer.test.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { render, RenderResult, fireEvent } from '@testing-library/react'
+
+import { Drawer } from '.'
+
+describe('Drawer', () => {
+  let wrapper: RenderResult
+  let onCloseSpy: (event: React.FormEvent<HTMLButtonElement>) => void
+  let children: React.ReactElement
+
+  describe('default props', () => {
```
thyhjwb6, line 18: When there are nullable props I usually add some tests that cover those scenarios

packages/react-component-library/src/components/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,44 @@
...
+}) => {
+  const [drawerOpen, setDrawerOpen] = useState(isOpen)
+
+  const classes = classNames('rn-drawer', {
+    'is-open': drawerOpen,
+  })
+
+  function handleClick(event: React.FormEvent<HTMLButtonElement>) {
+    setDrawerOpen(false)
+    onClose(event)
```
thyhjwb6: Because `onClose` is nullable, this results in an error when not set, eg:

`Uncaught TypeError: onClose is not a function`

packages/react-component-library/src/components/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  return (
+    <div className={classes} data-testid="drawer-wrapper">
+      <div className="rn-drawer__inner">
+        <button
+          className="rn-drawer__close"
+          onClick={handleClick}
+          data-testid="drawer-close"
```
thyhjwb6, line 35: I wonder if this button can be found by using the text rather than `data-testid`..? I feel that `data-testid` should be a last resort

thyhjwb6 (commented):
Looks good, few comments 👍

packages/react-component-library/src/components/Drawer/Drawer.tsx
m7kvqbe1: It's slightly annoying to target because i'm using the HTML character code `&times;`.

I figure by using the testid we can modify the implementation without breaking the test, so they are resilient to change?

packages/react-component-library/src/components/Drawer/Drawer.tsx
```diff
@@ -20,7 +20,10 @@ export const Drawer: React.FC<DrawerProps> = ({
 
   function handleClick(event: React.FormEvent<HTMLButtonElement>) {
     setDrawerOpen(false)
-    onClose(event)
+
+    if (typeof onClose === 'function') {
```
thyhjwb6: 👍 don't know if we need to check that it's a function because that's implicit with the strong typing around `onClose`, ie if it's defined then it will definitely be a function

thyhjwb6 (approved):
👍 looks good, just a small comment which is nothing to worry about

Issue #349, opened 2019-10-14
Alert Component


ghost:
## Description

It's a way to communicate events / information to a user. 

## Component classification

This component is an atom.

## Details / user needs

- Display a title, message and severity
- Needs to be dismissed by the user
- Displays within the flow of the page (not a toast).

## Questions

Outstanding questions for the Product Owner, Designers or Developers.

## Definition of done
- [ ] The Sketch file is updated with the component
- [ ] There are screenshots of all component variations in this ticket
- [ ] All states are included if applicable (default, hover, active, focus)
- [ ] A design documentation PR is linked from this ticket
- [ ] All questions laid out in this ticket have been resolved 
- [ ] A pull request for a component must contain a fully tested, documented and finished component in a state it can be shared with partner developers

hxltrhuxze:
# Alert Component

The Alert Component has 4 variations - Info, Danger, Warning, and Success. Each can be dismissed by clicking the x in the top right hand corner. There is the option to supply a Title and Description, or just a Description to the component.

<img width="1175" alt="Screenshot 2019-11-06 at 10 56 23" src="https://user-images.githubusercontent.com/48090803/68292608-24204880-0084-11ea-8a60-89c8e09407b1.png">

ghost:
We've discussed some work about colour-blindness. With the changes that will come from that, it gets sign-off from me. Out of interest, is the positioning of the alert in the window at the discretion of the developer/designer?

hxltrhuxze:
Updated the designs to include identifying icons from our icon library:

<img width="1201" alt="Screenshot 2019-11-06 at 11 46 29" src="https://user-images.githubusercontent.com/48090803/68295815-1de19a80-008b-11ea-8d52-f79c2618b418.png">


Regarding positioning: These Alerts sit within the flow of the document, so yeah they are at the discretion of the designer/developer using them.

PR #429, opened 2019-11-14
Add default alert component

thyhjwb6:
## Related issue
#349 

## Overview
Add an alert component which defaults to the information colour blue. When the close button is clicked, it hides the alert and calls and optional callback.

## Reason
This is a component required in dependent applications.

## Work carried out
- [x] Add alert component
- [x] Handle close click

### With title
Screenshot 2019-11-14 at 16 55 02

### Without title
Screenshot 2019-11-14 at 16 55 12

## Developer notes
Subsequent PRs will include more alert variants and documentation.

### DIFF
--- a/packages/css-framework/src/components/_alert.scss
+++ b/packages/css-framework/src/components/_alert.scss
@@ -11,6 +11,8 @@
 }
 
 .rn-alert__icon {
+  display: inline-flex;
+  align-items: center;
   color: color("primary", 800);
   padding-right: 6px;
 }
@@ -36,7 +38,8 @@
 }
 
 .rn-alert__content {
-  display: flex;
+  display: inline-flex;
+  align-items: center;
 }
 
 .rn-alert__description {
--- a/packages/css-framework/src/components/_alert.scss
+++ b/packages/css-framework/src/components/_alert.scss
@@ -21,6 +21,20 @@
   font-weight: bold;
 }
 
+.rn-alert__close {
+  position: absolute;
+  top: spacing(1);
+  right: spacing(1);
+  border: none;
+  font-size: font-size(l);
+  color: color("primary", 500);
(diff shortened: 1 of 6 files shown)


packages/css-framework/src/components/_alert.scss
```diff
@@ -1,100 +1,45 @@
...
-  ),
-  success: (
-    background: color("success", 000),
-    border: color("success", 200),
-    copy: color("success", 800),
-    title: color("success", 900),
-  ),
-);
-
 .rn-alert {
```
m7kvqbe1, line 1: In the 'without title' screenshot the line height looks a little bit off - maybe this could be vertically centered with `display: inline-flex;` and `align-items: center;`?

packages/react-component-library/src/components/Alert/Alert.test.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+    beforeEach(() => {
+      onCloseSpy = jest.fn()
+
+      wrapper = render(<Alert onClose={onCloseSpy}>Description</Alert>)
+    })
+
+    describe('when the close button is clicked', () => {
+      beforeEach(() => {
+        wrapper.getByTestId('close').click()
```
m7kvqbe1, line 103: I've been doing the same in a much more verbose way - I didn't realise `click` existed.

m7kvqbe1 (commented):
Looks great to me just one comment about alignment.

packages/react-component-library/src/components/Alert/Alert.test.tsx
thyhjwb6: Yeah this seems to work nice. I think there can be scenarios where you want to be more specific with the `click` 🤷‍♂

m7kvqbe1 (approved):

Issue #227, opened 2019-08-16
Popover component


ghost:
## Description

A popover is a self-contained modal or dialogue that appears in context next to a trigger element. It contains more content that a tooltip, but is directly related to a trigger element in the same way as a tooltip.

## Component classification

This component is a Molecule.

## Details / user needs

Popovers provide larger volumes of contextual information to a specific element on an interface. This content can be informative, or include interactive elements.
- Shows on hover.
- Can display left, right, above or below the target element.
- Can take any JSX.
- Popover by default only supplies the wrapper.
- If just putting content, use content sub-component.

## Definition of done
- [ ] The Sketch file is updated with the component
- [ ] There are screenshots of all component variations in this ticket
- [ ] All states are included if applicable (default, hover, active, focus)
- [ ] A design documentation PR is linked from this ticket
- [ ] All questions laid out in this ticket have been resolved 
- [ ] A pull request for a component must contain a fully tested, documented and finished component in a state it can be shared with partner developers

ghost:
@aavvjg6j - we thought you might have some other examples of where the popover is being designed in another application. Can you add anything to the Details/User needs section above please?

antictor:
FB will need this early on, and there is a basic version already in place. Would be good to get some good examples of it's requirements documented and finish it up.

hxltrhuxze:
The Popover functions similarly to the tooltip component, but with space for rich content. This design contains just the Frame.

<img width="418" alt="Screenshot 2019-11-04 at 16 19 49" src="https://user-images.githubusercontent.com/48090803/68137526-0d141600-ff1f-11e9-83d1-2b452c197150.png">

A working example of what could _potentially_ be created with the Popover.
<img width="418" alt="Screenshot 2019-11-04 at 16 20 35" src="https://user-images.githubusercontent.com/48090803/68137630-346ae300-ff1f-11e9-9792-684e3adab3a4.png">

PR #434, opened 2019-11-18
Popover Component

m7kvqbe1:
## Related issue

https://github.com/Royal-Navy/standards-toolkit/issues/227

## Overview

Enhance existing Popover atom implementation and expose.

## Reason

>A popover is a self-contained modal or dialogue that appears in context next to a trigger element. It contains more content that a tooltip, but is directly related to a trigger element in the same way as a tooltip.

## Work carried out

- [x] Implement hover functionality
    - [x] Preserve open state on Popover mouseOver
    - [x] Resolve issues with positioning based on placement
- [x] Update styles in line with design
- [x] Decouple Popover from NotificationPanel (add FloatingBox primitive)
- [x] Implement automated tests

## Developer notes

The GIF recording makes the box-shadow look broken. It doesn't look like that in browser.


m7kvqbe1:
Adding explanation from Slack for reference:

>We have resolved the Popover issue by adding another level of abstraction with an non-themed / non-styled common ancestor called 'FloatingBox'. This component sits within the 'objects' directory within the component library and is consumed by both Popover and NotificationPanel.
>
>We have chosen the name 'objects' as it matches a convention already used within the css-framework package. Think of FloatingBox as a primitive.
>
>The idea is that within the component library we should try to avoid consuming atomic components themselves and instead use un-styled 'primitives' like FloatingBox. (edited) 
>
>This way we see the benefits of DRY implementation while also maintaining decoupled styles.

m7kvqbe1:
We've changed the name to primitives instead of objects.

m7kvqbe1:
We are moving all shared styles into the `FloatingBox` stylesheet for now.

This way we avoid heavy duplication but the styles remain open for extension.

m7kvqbe1:
Issue raised and PR with fix submitted for missing type definition: https://github.com/danreeves/react-tether/issues/218

packages/react-component-library/src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,97 @@
...
+  const timerRef = useRef(null)
+  const PLACEMENTS = POPOVER_PLACEMENTS[placement]
+
+  const classes = classNames('rn-popover', className, {
+    'is-visible': isVisible,
+  })
+
+  function handleOnMouseEnter(e: React.MouseEvent) {
+    if (timerRef.current !== null) {
+      clearTimeout(timerRef.current)
```
thyhjwb6, line 47: There's no test for this line

packages/react-component-library/src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    setIsVisible(true)
+  }
+
+  function handleOnMouseLeave(_: React.MouseEvent) {
+    timerRef.current = setTimeout(() => {
```
thyhjwb6, line 54: There's no test covering these lines

packages/css-framework/src/objects/_floating-box.scss
```diff
@@ -0,0 +1,231 @@
+$floating-box-light-background: color(neutral, white);
```
thyhjwb6: I wonder if we should prefix these with `_` as they are private and this will be neater going forward. I did this in `_alert.scss`. Maybe it's worth having a discussion to make sure we are happy with the pattern.

packages/react-component-library/src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,96 @@
...
+  FLOATING_BOX_PLACEMENT,
+  FLOATING_BOX_PLACEMENT_ARROW_POSITION_MAP,
+} from '../../primitives/FloatingBox'
+
+import { calculate } from './popoverPosition'
+
+interface PopoverProps
+  extends Omit<FloatingBoxProps, 'onMouseEnter' | 'onMouseLeave'> {
+  children: React.ReactElement
+  popoverJSX: React.ReactElement
```
thyhjwb6: I feel like a nicer name for this could be `content` or something similar.

packages/react-component-library/src/primitives/FloatingBox/FloatingBox.tsx
```diff
@@ -0,0 +1,74 @@
...
+    | FLOATING_BOX_ARROW_POSITION.TOP_LEFT
+    | FLOATING_BOX_ARROW_POSITION.TOP_RIGHT
+    | FLOATING_BOX_ARROW_POSITION.BOTTOM_LEFT
+    | FLOATING_BOX_ARROW_POSITION.BOTTOM_RIGHT
+  onMouseEnter?: (e: React.MouseEvent) => void
+  onMouseLeave?: (e: React.MouseEvent) => void
+  children?: React.ReactElement
+}
+
+export const FloatingBox = forwardRef(
```
thyhjwb6, line 28: This could also be declared as `forwardRef<FloatingBoxProps, HTMLDivElement>`. Not sure what reads nicer 🤷‍♂

packages/react-component-library/src/primitives/FloatingBox/FloatingBox.tsx
```diff
@@ -0,0 +1,74 @@
...
+    const style = {
+      width,
+      height,
+      top,
+      right,
+      bottom,
+      left,
+    }
+
+    const classes = classNames('rn-floating-box', className, {
```
thyhjwb6: Possibly this could be

```
const classes = classNames(
  "rn-floating-box",
  className,
  `rn-floating-box--${scheme}`,
  `rn-floating-box--${position}`
)

File: packages/react-component-library/src/fragments/NotificationPanel/notificationPosition.test.ts [hunk 7]
```diff
@@ -43,8 +43,8 @@ describe('notificationPosition', () => {
     })
 
     it('should calculate the position to be below the element', () => {
-      expect(result).toHaveProperty('top', 203)
-      expect(result).toHaveProperty('left', 590)
+      expect(result).toHaveProperty('top', 200)
```
thyhjwb6, line 46: If the previous commit was atomic then this would be there.

packages/docs-site/src/library/pages/components/popover.md
```diff
@@ -0,0 +1,98 @@
...
+---
+
+import { Popover, Tab, TabSet } from '@royalnavy/react-component-library'
+import DataTable from '../../../components/presenters/data-table'
+import CodeHighlighter from '../../../components/presenters/code-highlighter'
+import SketchWidget from '../../../components/presenters/sketch-widget'
+
+# Overview
+
+A popover is a self-contained modal or dialogue that appears in context next to a trigger element. It contains more content that a tooltip, but is directly related to a trigger element in the same way as a tooltip.
```
thyhjwb6: `that` 👉  `than`

packages/react-component-library/src/components/Popover/Popover.test.tsx
```diff
@@ -0,0 +1,98 @@
+import React from 'react'
```
thyhjwb6, line 1: No worries for now but something to think about is that if your previous commit was atomic then it would have included tests.

packages/react-component-library/src/primitives/FloatingBox/FloatingBox.tsx
```diff
@@ -26,7 +26,7 @@ export interface FloatingBoxProps extends PositionType, ComponentWithClass {
 }
 
 export const FloatingBox = forwardRef(
-  (props: FloatingBoxProps, ref?: React.Ref<HTMLDivElement>) => {
+  (props: FloatingBoxProps, ref?: React.Ref<any>) => {
```
thyhjwb6, line 29: Does this need to be `any`? Does `HTMLElement` work?

thyhjwb6 (commented):
Looks good. Some of these comments are more thoughts rather than compulsory.

packages/react-component-library/src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+import {
+  FloatingBox,
+  FloatingBoxProps,
+  FLOATING_BOX_SCHEME,
+} from '../../primitives/FloatingBox'
+
+interface PopoverProps
+  extends Omit<FloatingBoxProps, 'onMouseEnter' | 'onMouseLeave'> {
+  children: React.ReactElement
```
thyhjwb6, line 19: One thing I was thinking is whether it's more intuitive that the `children` is the content and then there is a prop for the `trigger`. Just a thought..

packages/react-component-library/src/components/Popover/Popover.tsx
m7kvqbe1: maybe `renderTarget`

packages/react-component-library/src/primitives/FloatingBox/FloatingBox.tsx
m7kvqbe1: I've just relied on prettier here and left as is - but this is cleaner.

packages/docs-site/src/library/pages/components/popover.md
m7kvqbe1: Ooops - copied pasted from the issue text.

packages/react-component-library/src/primitives/FloatingBox/FloatingBox.tsx
m7kvqbe1: I couldn't get this to play nice with anything other than `any` because there seems to be some intermediary type of `LegacyRef` being passed (I think it's coming from the the Tether library implementation).

packages/react-component-library/src/components/Popover/Popover.tsx
m7kvqbe1: The component tree made more sense to me with Target as child, as it more closely represented the pre-existing component tree had there not been a Popover. I thought of Popover as a wrapper component that enhanced a target (child).

```
Window 
   > Document 
      > Container 
        > Popover (Popover content as prop)
          > Target 

vs

Window 
   > Document 
      > Container 
         > Popover (Target as prop)
            > Content?
```

Popover is an enhancement of Target not the other way around? I could maybe implement it as a HOC instead. And you enhance target components.

Something like:

```js
const TargetComponentWithPopover = withPopover(TargetComponent, {
  // options
})
```

packages/react-component-library/src/primitives/FloatingBox/FloatingBox.tsx
m7kvqbe1: Not quite sure how you mean it could be different - could you elaborate further?

thyhjwb6: Apologies for not being clear:

```
export const FloatingBox = forwardRef<FloatingBoxProps, any>(({ ... }, ref) => {
  return (
    ...
  )
})

thyhjwb6 (approved):

packages/react-component-library/src/components/Popover/Popover.tsx
m7kvqbe1: I'm going to implement this enhancer as part of another PR in the near future. Marking as resolved for now.

Issue #225, opened 2019-08-16
Date and time picker component


ghost:
## Description

A way for a user to select a date and time. Note that there is a separate (but related) component for date and time RANGE selection.

## Component classification

Is this component an atom, molecule or organism?
- [ ] Atom
- [x] Molecule
- [ ] Organism

## Details / user needs

Users often need to select dates and times to complete simple tasks.

Some design considerations:
- Should allow users to key in a date / time
- Should allow users to graphically select a time and date
- Should include optional timezone selection. This is especially relevant as users are likely to be globally distributed
- Should allow users to increment times by using arrow keys or similar
- Should allow for set time periods to be selected, for example in six-hourly increments

_Used in 2x exemplar projects_

## Variations

- Version that just allows a date to be selected
- Version that just allows a time to be selected 

## Questions

- What time fidelity is useful here? Minutes, seconds?

## Definition of done
- [ ] The Sketch file is updated with the component
- [ ] There are screenshots of all component variations in this ticket
- [ ] All states are included if applicable (default, hover, active, focus)
- [ ] A design documentation PR is linked from this ticket
- [ ] All questions laid out in this ticket have been resolved 
- [ ] A pull request for a component must contain a fully tested, documented and finished component in a state it can be shared with partner developers

hxltrhuxze:
Variation of this component can be found in the closed issue #226

hxltrhuxze:
# Date & Time Picker

### Inputs
The Date and Time inputs look similar to the regular text input, however they have additional formatting to add the `/` to the dates and `:` to the time. Clicking either the Date or Time input will open a dropdown containing the pickers.
<img width="434" alt="Screenshot 2019-11-18 at 12 20 09" src="https://user-images.githubusercontent.com/48090803/69052028-dc8faa00-09fd-11ea-933d-a85c3a65ac9e.png">
### Date Picker
The Date picker has two variations - A Single Date Picker and a Single Date with Quarterly Hour picker.
<img width="714" alt="Screenshot 2019-11-18 at 12 20 17" src="https://user-images.githubusercontent.com/48090803/69052029-dc8faa00-09fd-11ea-8a05-decdd666ec47.png">
### Start and End
Often, we'll need the user to pick a start and end date. This has two calendars side-by-side, with arrows to navigate left/right between months. Click on either calendar to add a start date. Clicking again will add the end date and highlight all days in the range. This picker also has the variation to allow the user to select and Quarterly hour period.
[...]

PR #443, opened 2019-11-22
Add scrollable tab set

thyhjwb6:
## Related issue
#225 

## Overview
Adding a scrollable tab set component 

## Reason
When there are a large number of tabs then these need to be scrolled through.

## Work carried out
- [x] Add scroll buttons
- [x] Add scrollable tab set

## Screenshot
Screenshot 2019-11-27 at 09 28 42

## Developer notes
Still to do:
- Smooth scrolling
- Mobile
- Documentation

### DIFF
--- a/packages/react-component-library/src/components/TabSet/TabSet.tsx
+++ b/packages/react-component-library/src/components/TabSet/TabSet.tsx
@@ -1,44 +1,82 @@
 import React, { useState, Children } from 'react'
+import classNames from 'classnames'
 
 import { TabProps, TabContent, TabItem } from '.'
+import { ScrollButton } from './ScrollButton'
+import { useScrollableTabSet } from './useScrollableTabSet'
+import { SCROLL_DIRECTION } from './constants'
 
 interface TabSetProps {
   className?: string
-  children?: React.ReactElement<TabProps>[]
-  onChangeCallback?: (id: number, title: string) => void
+  children: React.ReactElement<TabProps>[]
+  onChangeCallback?: (id: number) => void
+  scrollable?: boolean
 }
 
 export const TabSet: React.FC<TabSetProps> = ({
   className = '',
   children,
   onChangeCallback,
+  scrollable,
 }) => {
   const [activeTab, setActiveTab] = useState(0)
+  const { updateCurrentScrollToTab, tabsRef, itemsRef } = useScrollableTabSet(
+    children
+  )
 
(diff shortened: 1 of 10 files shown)


m7kvqbe1:
Is this WIP?

packages/react-component-library/src/components/TabSet/TabSet.tsx
```diff
@@ -1,44 +1,81 @@
...
+    children
+  )
 
-  function handleClick(index: number, title: string) {
+  function handleClick(index: number) {
     setActiveTab(index)
 
     if (onChangeCallback) {
-      onChangeCallback(activeTab, title)
+      onChangeCallback(index)
```
m7kvqbe1, line 31: I know this was already here before and it's annoyingly a breaking change but maybe we could lose the word 'callback' and it could become `onChange`.

m7kvqbe1: Maybe there is a task here to carry out ahead of the v2.0.0 release - we could make our interfaces more consistent across the entire library.

packages/react-component-library/src/components/TabSet/TabSet.tsx
```diff
@@ -1,44 +1,81 @@
...
     }
   }
 
+  const headerClassNames = classNames('rn-tab-set__head', {
+    'is-scrollable': scrollable,
```
m7kvqbe1: Could we just add this stateful class in once, at the root of the component?

packages/react-component-library/src/components/TabSet/useScrollableTabSet.ts
```diff
@@ -0,0 +1,33 @@
...
+export function useScrollableTabSet(items: React.ReactElement<TabProps>[]) {
+  const [currentScrollToTab, setCurrentScrollToTab] = useState(0)
+  const itemsRef = useRef<Array<HTMLLIElement>>([])
+  const tabsRef = useRef<HTMLDivElement>()
+
+  useEffect(() => {
+    itemsRef.current = itemsRef.current.slice(0, items.length)
+  }, [items])
+
+  function updateCurrentScrollToTab(change: (currentTab: number) => number) {
```
m7kvqbe1, line 14: Lovely bit of inversion of control :)

m7kvqbe1 (approved):
Looks great to me - just a couple of tiny comments

PR #446, opened 2019-11-25
Feature/sass modules

hxltrhuxze:
## Overview

Adds SASS module support to the CSS Framework.

## Reason

SASS has introduced a new module system that works more like traditional `js` imports, rather than the traditional `@import` syntax.

By updating the CSS Framework to use modules, we remove the bug where multiple imports of the SCSS stylesheet would result in a number of repeated rules in devtools. Modules also allow us to privatise functions and variables, letting us expose only the functions, mixins, and variables that we want the consumer to interact with.

## Work carried out

- [x] Converted the SCSS Framework to SCSS Modules
- [x] Converted the Docs Site to use modules


packages/css-framework/src/contexts/_light.scss
```diff
@@ -172,125 +177,125 @@
...
-    "warning": (
-      "default": color("warning", "500"),
-      "faded": color("warning", "200"),
-    ),
-    "success": (
-      "default": color("success", "500"),
-      "faded": color("success", "200"),
-    ),
-  );
+  // defaults.$states: (
```
m7kvqbe1: Does this stuff need to stay in? Can we just remove it if it's commented out?

packages/css-framework/src/_config.scss
```diff
@@ -1,17 +1,25 @@
...
+$typography:  () !default;
+$zindex:      () !default;
 
 
 // Utility Namespace
 $utility-ns: "rn_";
 
-$navy-content-width: 1280px;
+$content-width: 1280px;
+$animation-timing: 0.3s;
```
m7kvqbe1, line 25: Regarding missing newlines at end of files - depending on what is parsing / processing these files it can cause problems. My editor adds them on save automatically just to be safe.

https://unix.stackexchange.com/a/18789

>A text file, under unix, consists of a series of lines, each of which ends with a newline character (\n). A file that is not empty and does not end with a newline is therefore not a text file.
>
>Utilities that are supposed to operate on text files may not cope well with files that don't end with a newline; historical Unix utilities might ignore the text after the last newline, for example. GNU utilities have a policy of behaving decently with non-text files, and so do most other modern utilities, but you may still encounter odd behavior with files that are missing a final newline¹.

packages/css-framework/index.scss
```diff
@@ -1,109 +1,94 @@
...
+@use "src/components/modal";
+@use "src/components/nav";
+@use "src/components/numberinput";
+@use "src/components/pagination";
+// @use "src/components/panel";
+@use "src/components/popover";
+@use "src/components/progress";
+@use "src/components/phase-banner";
+@use "src/components/radio";
+// @use "src/components/range-slider";
```
m7kvqbe1: Should these still be commented out?

packages/css-framework/src/components/_alert.scss
```diff
@@ -1,48 +1,52 @@
+@use "sass:map";
+@use "../abstracts/mixins" as m;
+@use "../abstracts/functions" as f;
```
m7kvqbe1, line 3: Do we have to expose it via this aliased object (`f`)? Can we no longer just import and use `color()` if we use modules?

m7kvqbe1 (commented):
Looks good - just a couple of comments - definitely think it needs a quick hangouts just to make sure we are all on the same page about how it works.

packages/css-framework/src/contexts/_light.scss
hxltrhuxze: Yeah good shout, I'll remove it.

packages/css-framework/src/components/_alert.scss
hxltrhuxze: We can easily import functions or mixins as `@use "../abstracts/mixins" as *;`, where the * adds the function to the global namespace. This'll then let you call `color()` without needing the namespace.

I decided to namespace them with `f` & `m` though as if we decide to import a different external project, it'll mean we won't have clashes.

m7kvqbe1 (approved):

m7kvqbe1:
I think it probably needs a rebase as there were some security fixes applied a few days ago.

PR #448, opened 2019-11-26
DatePicker Component

m7kvqbe1:
## Related issue

https://github.com/Royal-Navy/standards-toolkit/issues/225

## Overview

Implement DatePicker that allows the selection of a single date value.

## Reason

>A way for a user to select a date and time. Note that there is a separate (but related) component for date and time RANGE selection.

## Work carried out

- [x] Implement single value DatePicker
- [x] Add automated tests

## Developer notes

Additional functionality (including a range enabled variant) will come in subsequent PRs.

### DIFF
--- a/packages/react-component-library/src/components/DatePicker/DatePicker.test.tsx
+++ b/packages/react-component-library/src/components/DatePicker/DatePicker.test.tsx
@@ -1,0 +1,179 @@
+import React from 'react'
+import '@testing-library/jest-dom/extend-expect'
+
+import { render, RenderResult, fireEvent } from '@testing-library/react'
+
+import { DatePicker, StateObject } from './DatePicker'
+import { DATEPICKER_PLACEMENT } from './constants'
+
+const NOW = '2019-12-05T11:00:00.000Z'
+
+function click(element: HTMLElement) {
+  fireEvent.mouseDown(element)
+  fireEvent.click(element)
+}
+
+describe('DatePicker', () => {
+  let wrapper: RenderResult
+  let startDate: Date
+  let label: string
+  let onChange: (data: StateObject) => void
+  let onBlur: (e: React.FormEvent) => void
+  let dateSpy: jest.SpyInstance
+
+  beforeAll(() => {
+    dateSpy = jest
+      .spyOn(Date, 'now')
+      .mockImplementation(() => new Date(NOW).valueOf())
+  })
+
+  afterAll(() => {
+    dateSpy.mockRestore()
+  })
+
+  describe('default props', () => {
+    beforeEach(() => {
+      startDate = new Date('12/01/2019')
+      onChange = jest.fn()
+      onBlur = jest.fn()
+
+      wrapper = render(
+        <>
+          <DatePicker
+            startDate={startDate}
+            placement={DATEPICKER_PLACEMENT.BELOW}
+            onChange={onChange}
+            onBlur={onBlur}
+          />
+          <div data-testid="datepicker-outside" />
+        </>
+      )
+    })
+
+    it('renders the component', () => {
+      expect(wrapper.queryByTestId('datepicker-wrapper')).toBeInTheDocument()
+    })
+
+    describe('when the end user focuses on the Input', () => {
+      beforeEach(() => {
+        wrapper.getByTestId('datepicker-input').focus()
+      })
+
+      it('displays the container', () => {
+        expect(
+          wrapper.getByTestId('datepicker-input-wrapper').classList
+        ).toContain('is-open')
+
+        expect(wrapper.getByTestId('floating-box').classList).toContain(
+          'is-visible'
+        )
+      })
+
+      describe('and the end user clicks on the navigate month buttons', () => {
+        describe('and clicks next', () => {
+          beforeEach(() => {
+            wrapper.getByTestId('datepicker-input').focus()
+
+            click(wrapper.getAllByTestId('datepicker-nav-button')[1])
+          })
+
+          it('changes to the next month', () => {
+            expect(
+              wrapper.getByTestId('datepicker-month-label').innerHTML
+            ).toBe('January 2020')
+          })
+
+          it('does not hide the container', () => {
+            expect(
+              wrapper.getByTestId('datepicker-input-wrapper').classList
+            ).toContain('is-open')
+
+            expect(wrapper.getByTestId('floating-box').classList).toContain(
+              'is-visible'
+            )
+          })
+        })
+
+        describe('and clicks previous', () => {
+          beforeEach(() => {
+            wrapper.getByTestId('datepicker-input').focus()
+
+            click(wrapper.getAllByTestId('datepicker-nav-button')[0])
+          })
+
+          it('changes to the previous month', () => {
+            expect(
+              wrapper.getByTestId('datepicker-month-label').innerHTML
+            ).toBe('November 2019')
+          })
+
+          it('does not hide the container', () => {
+            expect(
+              wrapper.getByTestId('datepicker-input-wrapper').classList
+            ).toContain('is-open')
+
+            expect(wrapper.getByTestId('floating-box').classList).toContain(
+              'is-visible'
+            )
+          })
+        })
+
+        describe('and clicks on a day', () => {
+          beforeEach(() => {
+            click(wrapper.getByTestId('datepicker-day-01'))
+          })
+
+          it('set the value of the component to this date', () => {
+            expect(
+              wrapper.getByTestId('datepicker-input').getAttribute('value')
+            ).toBe('12/1/2019')
+          })
+
+          it('invokes the onChange callback', () => {
+            expect(onChange).toHaveBeenCalledTimes(1)
+          })
+
+          it('does not hide the container', () => {
+            expect(
+              wrapper.getByTestId('datepicker-input-wrapper').classList
+            ).toContain('is-open')
+
+            expect(wrapper.getByTestId('floating-box').classList).toContain(
+              'is-visible'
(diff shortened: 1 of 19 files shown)


packages/react-component-library/src/components/DatePicker/Month.tsx
```diff
@@ -0,0 +1,94 @@
...
+export interface MonthProps extends ComponentWithClass {
+  year: number
+  month: number
+  firstDayOfWeek: FirstDayOfWeek
+  goToPreviousMonths: () => void
+  goToNextMonths: () => void
+}
+
+const formatDayLabel = (dayLabel: string): string => {
+  switch (dayLabel) {
```
thyhjwb6: Switch statements go in the same category as `else if else if else if` 😄 

Another option is:

```
const dayFormat = {
  'Mo': 'Mon',
  'Tu': 'Tue',
  'We': 'Wed',
  'Th': 'Thu',
  'Fr': 'Fri',
  'Sa': 'Sat',
  'Su': 'Sun',
}

const formatDayLabel = (dayLabel: string): string => {
  return dayFormat[dayLabel] || dayLabel
}
```

thyhjwb6 (commented):
After a first glance at the code I think this looks good.

I am still confused about that `goToPreviousYear` test passing in the date picker hooks library.

packages/react-component-library/src/components/DatePicker/Month.tsx
m7kvqbe1: 😂

packages/react-component-library/src/components/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,183 @@
...
+    startDate,
+    endDate,
+    focusedInput: START_DATE, // START_DATE | END_DATE | null
+  })
+
+  function handleDateChange(data: StateObject) {
+    if (!data.focusedInput) {
+      setState({ ...data, focusedInput: START_DATE })
+    } else {
+      setState(data)
```
thyhjwb6: This is not being covered by tests

packages/react-component-library/src/components/DatePicker/Day.tsx
```diff
@@ -0,0 +1,76 @@
...
+    isDateBlocked,
+    isFirstOrLastSelectedDate,
+    onDateFocus,
+    onDateSelect,
+    onDateHover,
+    dayRef,
+  })
+
+  if (!dayLabel) {
+    return <div />
```
thyhjwb6: This is not being covered by tests

packages/react-component-library/src/components/DatePicker/useOpenClose.ts
```diff
@@ -0,0 +1,41 @@
+import React, { useState, useEffect } from 'react'
+
+function hasParentWithMatchingSelector(target: Node, selector: string): boolean {
+  return [...document.querySelectorAll(selector)].some(
```
thyhjwb6, line 9: This is not being covered by tests

packages/react-component-library/src/components/DatePicker/useOpenClose.ts
```diff
@@ -0,0 +1,41 @@
...
+    el => el !== target && el.contains(target)
+  )
+}
+
+export function useOpenClose(ref: React.RefObject<HTMLDivElement>) {
+  const [isOpen, setIsOpen] = useState<boolean>(false)
+
+  function handleClick(e: MouseEvent) {
```
thyhjwb6: This handler is not being covered by tests

packages/react-component-library/src/components/DatePicker/DatePicker.test.tsx
```diff
@@ -0,0 +1,152 @@
...
+          placement={DATEPICKER_PLACEMENT.BELOW}
+        />
+      )
+    })
+
+    it('renders the component', () => {
+      expect(wrapper.queryByTestId('datepicker-wrapper')).not.toBeNull()
+    })
+
+    describe('when the end user focusses on the Input', () => {
```
thyhjwb6: Small typo in "focusses"

packages/react-component-library/src/components/DatePicker/DatePicker.test.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+import { DatePicker, StateObject } from './DatePicker'
+import { DATEPICKER_PLACEMENT } from './constants'
+
+describe('DatePicker', () => {
+  let wrapper: RenderResult
+  let startDate: Date
+  let label: string
+  let onChange: (data: StateObject) => void
+  let onBlur: (e: React.FormEvent<Element>) => void
```
thyhjwb6: The type `<Element>` can be removed as it's implicit

packages/react-component-library/src/components/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,183 @@
...
+  endDate: Date | null
+  focusedInput: FocusedInput
+}
+
+export interface DatePickerProps extends ComponentWithClass {
+  id?: string
+  name?: string
+  value?: string
+  label?: string
+  onBlur?: (event: React.FormEvent<Element>) => void
```
thyhjwb6: `Element` type is implicit

packages/react-component-library/src/components/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,183 @@
...
+import DatepickerContext from './datepickerContext'
+
+import { Input } from './Input'
+import { Month } from './Month'
+import { useOpenClose } from './useOpenClose'
+import { DATEPICKER_PLACEMENT, DATEPICKER_PLACEMENTS } from './constants'
+import { FloatingBox, FLOATING_BOX_SCHEME } from '../../primitives/FloatingBox'
+
+export interface StateObject {
+  startDate: Date | null
```
thyhjwb6: If `startDate` or `endDate` are nullable ie `startDate?`, is the ` | null` required?

packages/react-component-library/src/components/DatePicker/Month.tsx
```diff
@@ -0,0 +1,90 @@
...
+      <div className="rn-date-picker__day-label-grid">
+        {weekdayLabels.map(dayLabel => (
+          <div className="rn-date-picker__day-label" key={dayLabel}>
+            {formatDayLabel(dayLabel)}
+          </div>
+        ))}
+      </div>
+      <div className="rn-date-picker__day-grid">
+        {days.map((day, index) => {
+          if (typeof day === 'object') {
```
thyhjwb6, line 73: If we get inside here, what else could `day` be?

thyhjwb6 (commented):
Looks good. A few comments and maybe you will say some of the tests are hard to cover. Would be keen to see more of the custom hook covered as some of the code appears to be tricky.

Also when I test this in Storybook, I would expect the floating box to be hidden when selecting a date. Is that not right?

packages/react-component-library/src/components/DatePicker/Month.tsx
m7kvqbe1: I'm actually handling the blank days somewhere else - this has now been changed. Worth noting i've had to cast the type of items in the days collection returned by the hook.

Original union type caused a type error:

```typescript
(parameter) day: number | {
    dayLabel: string;
    date: Date;
}
```

m7kvqbe1: There appears to be an edge case where the hook will return a number instead of a Date object.

packages/react-component-library/src/components/DatePicker/useOpenClose.ts
m7kvqbe1: The thing is this stuff is covered by the following test:

https://github.com/Royal-Navy/standards-toolkit/pull/448/files#diff-1b225051935f3e919127b8f0d38191beR125

Do you think I need to test the hook itself independent of the component? Isn't that now a test of implementation detail?

thyhjwb6: I am not necessarily a slave to the coverage metric but what I would have concerns about is that if I change the hook to what's below then all tests still pass.

I agree that we need to be careful of testing implementation details but I will favour getting the coverage right to prevent regressions. There must be a way to cover this without explicitly testing the hook.

```
import { useState, useEffect } from 'react'

export function useOpenClose() {
  const [isOpen, setIsOpen] = useState<boolean>(false)

  useEffect(() => {
    document.addEventListener('mousedown', () => {})

    return () => {
      document.removeEventListener('mousedown', () => {})
    }
  }, [])

  const onFocus = () => {
    setIsOpen(true)
  }

  return {
    isOpen,
    onFocus
  }
}
```

packages/react-component-library/src/components/DatePicker/DatePicker.tsx
thyhjwb6: All tests still pass when this is:

```
    if (!data.focusedInput) {
      setState({ ...data, focusedInput: START_DATE })
    } else {
      // setState(data)
    }
```

packages/react-component-library/src/components/DatePicker/useOpenClose.ts
m7kvqbe1: Looks like the test I thought was covering that behaviour isn't. I'll dig a little deeper.

thyhjwb6 (approved):
@m7kvqbe1  and I both agree this is ready for merge.

m7kvqbe1 (commented):
LGTM

Issue #218, opened 2019-08-15
Tabset component


ghost:
## Description

Tabsets allow users to view large quantities of data without changing page.

## Component classification

Is this component an atom, molecule or organism?
- [ ] Atom
- [x] Molecule
- [ ] Organism

## Details / user needs

Tabsets allow users to view large quantities of screen content from a single page. They can be used for a variety of purposes such alternative views of a data set.

Tabsets should contain a box, and ability to switch between data sets.

We have previously discussed two types of tabset:
- A design where the tabset is a dominant part of the page. This can have larger controls.
- A smaller tabset taking up a subsection of a page where the user can cycle between tabs using arrow buttons.

Other considerations:
- Tab switches should be able to contain more than just text (icons etc)

_Used in 2x exemplar projects_

## Variations

It would be useful to have tabsets that are laid out vertically and horizontally

## Definition of done
- [ ] The Sketch file is updated with the component
- [ ] There are screenshots of all component variations in this ticket
- [ ] All states are included if applicable (default, hover, active, focus)
- [ ] A design documentation PR is linked from this ticket
- [ ] All questions laid out in this ticket have been resolved 
- [ ] A pull request for a component must contain a fully tested, documented and finished component in a state it can be shared with partner developers

hxltrhuxze:
# Proposed Design

### Fixed Tabset
The Fixed Tabset is already included in the component library https://docs.royalnavy.io/components/tab-set

<img width="688" alt="Screenshot 2019-08-22 at 13 30 18" src="https://user-images.githubusercontent.com/48090803/63514672-01e54880-c4e1-11e9-98c4-9281cc8f8483.png">

### Scrollable Tabset
The scrollable Tabset allows the user to scroll though multiple tabs. The most common use case for this is multiple dates:

<img width="702" alt="Screenshot 2019-08-22 at 13 30 13" src="https://user-images.githubusercontent.com/48090803/63514699-1590af00-c4e1-11e9-8f04-d10706e3dd75.png">


## States
The Tabset states for default and hover:

![image](https://user-images.githubusercontent.com/48090803/63514602-d95d4e80-c4e0-11e9-8636-68c2af875c1b.png)


### Abstract Link
https://share.goabstract.com/69d7af2d-6041-49fe-a486-0efec77aeced?sha=14402486a99ae165999cc710b15ac680e9faec70

09txzS1qs:
Looks good to me!

PR #449, opened 2019-11-27
Add smooth scrolling to scrollable tab set

thyhjwb6:
## Related issue
#218 

## Overview
Make scrolling smooth when browsing tabs in scrollable tab set.

## Reason
Without smooth scrolling the user experience is awkward because it feels like the tabs are disappearing.

## Work carried out
- [x] Add smooth scrolling

## Screenshot
scroll

### DIFF
--- a/packages/react-component-library/src/components/TabSet/TabSet.test.tsx
+++ b/packages/react-component-library/src/components/TabSet/TabSet.test.tsx
@@ -1,8 +1,13 @@
 import React from 'react'
 import '@testing-library/jest-dom/extend-expect'
-import { render, RenderResult } from '@testing-library/react'
+import { fireEvent, render, RenderResult, wait } from '@testing-library/react'
 
-import { TabSet, Tab } from '.'
+import { Tab, TabSet } from '.'
+import { SCROLL_DIRECTION } from './constants'
+
+function flushPromises() {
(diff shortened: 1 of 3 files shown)


packages/react-component-library/src/components/TabSet/TabSet.test.tsx
```diff
@@ -118,12 +131,17 @@ describe('TabSet', () => {
...
       const tabs = wrapper.getByTestId('tabs')
       tabs.scrollTo = () => {}
-      scrollToSpy = jest.spyOn(tabs, 'scrollTo').mockImplementation(x => {
-        Object.defineProperty(tabs, 'scrollLeft', {
-          configurable: true,
-          value: x,
+
+      scrollToSpy = jest
+        .spyOn(tabs, 'scrollTo')
+        // @ts-ignore
```
m7kvqbe1, line 137: What did it get upset about here?

m7kvqbe1 (approved):
Looks good to me - it took me a while to process the useScrollableTabset hook but I don't see any issues.

packages/react-component-library/src/components/TabSet/TabSet.test.tsx
thyhjwb6: The destructuring of the `scrollTo` parameter.

PR #450, opened 2019-11-27
Allow else if exception for no-else-return

m7kvqbe1:
Let me do this...

```javascript
function foo() {
    if (x) {
        return y;
    } else if (z) {
        return w;
    }

    return t;
}
```

but not this

```javascript
function foo() {
    if (x) {
        return y;
    } else if (z) {
        return w;
    } else { // this last else is unnecessary, as it will just fall through
        return t;
    }
}
```

### DIFF
--- a/packages/react-component-library/.eslintrc.js
+++ b/packages/react-component-library/.eslintrc.js
@@ -35,6 +35,7 @@ module.exports = {
     'react/prop-types': 0,
     '@typescript-eslint/ban-ts-ignore': 0,
     '@typescript-eslint/explicit-function-return-type': 0,
+    'no-else-return': ['error', { allowElseIf: true }],
   },
   settings: {
     'import/resolver': {


thyhjwb6 (commented):
I am not really a big fan of using `else` or `else if` as I feel there are better patterns.

Using your example:

```
function foo() {
  if (x) {
    return y;
  }

  if (z) {
    return w;
  }

  return t;
}
```

or..

```
const blah = {
  x: () => y,
  z: () => w,
}

function foo(b: x | z) {
  return blah[b]() || t
}
```

m7kvqbe1:
Hmm - yeah ok there are other ways I could achieve the same thing while avoiding the else if.

For context this is what I was doing: https://github.com/Royal-Navy/standards-toolkit/blob/601533228b4fef7b72f17017da13ca3d0041adc5/packages/react-component-library/src/components/DatePicker/Day.tsx#L11

I'll close this issue.

PR #453, opened 2019-11-27
Update styles to be mobile first for scrollable tab set

thyhjwb6:
## Related issue
#218 

## Overview
Make scrollable tab set more suitable for mobile.

## Reason
Without this it is difficult to use.

## Work carried out
- [x] Update CSS

### Before
Screenshot 2019-11-28 at 17 05 05

### After
Screenshot 2019-11-28 at 17 05 54


m7kvqbe1 (commented):
This feedback should have actually been in the first PR, but I just noticed the icons in the left and right buttons aren't quite centered.

m7kvqbe1 (approved):

Issue #255, opened 2019-09-04
Develop tab has horizontal scrolling below certain browser widths


hxltrhuxze:
At small screen sizes, selecting the Develop tab on the component pages causes the page to horizontally scroll. This is due to the `<DataTable />` component not collapsing at smaller screen sizes.

**To Reproduce**
1. Navigate to the [Button Component](https://docs.royalnavy.io/components/button)
2. reduce the browser width to ~`1200px`
3. Note horizontal scroll

<img width="944" alt="Screenshot 2019-09-04 at 18 13 52" src="https://user-images.githubusercontent.com/48090803/64276483-fb72ba00-cf3f-11e9-9fa7-0be1cdb19886.png">

# Suggested Fix
To fix the issue of the horizontal scrolling, we'll need to reengineer the DataTables so that they can be used responsively. To solve this issue, we should make the tables themselves scrollable, rather than the browser:

#### Desktop
<img width="935" alt="Screenshot 2019-09-04 at 18 17 20" src="https://user-images.githubusercontent.com/48090803/64276620-53112580-cf40-11e9-825f-2b5642305476.png">

#### Tablet/Mobile
<img width="1127" alt="Screenshot 2019-09-04 at 18 26 49" src="https://user-images.githubusercontent.com/48090803/64277174-9d46d680-cf41-11e9-9863-7d6b536aa5c2.png">

The proposed DataTable also has some "niceties" added to improve the general look of the component and to satisfy user feedback.

### Changes made
- As the component decreases in width, I've ensured the table Header stays full width of the component
- Added the ability for the main section of the table to scroll on the `x-axis` when the table width exceeds the width of its parent.
- Reduced the overall number of columns visible by moving the description to a tooltip next to the Prop name.
- Added a "Default" key to the top right of the table. The Default for each prop is highlighted in the table.
- Ensured the entire table is set in a `monospace` font (excluding tooltips and the header).

PR #472, opened 2019-12-04
Horizontal scrolling docs site fix

hxltrhuxze:
## Related issue

#255 

## Overview

Adds horizontal scrolling to the tables on the Develop tab of the component pages to prevent the layout from breaking.

## Work carried out

- [x] Added horizontal scrolling to the tables

## Screenshot

Screenshot 2019-12-04 at 13 18 40

## Developer Notes
The solution isn't the cleanest as it requires a hack on flexbox to prevent the table from extending outside its parent.


m7kvqbe1 (approved):
Nice looks good

Issue #437, opened 2019-11-19
Docs site SVG rendering bug


hxltrhuxze:
**Describe the bug**
When adding SVG images for the docs site components, sometimes the SVG doesn't display correctly, causing visual bugs (see Screenshots). The SVGs have been exported from Sketch and run through [SVGOMG](https://jakearchibald.github.io/svgomg/). The optimised SVG is then added to the docs site, where the MDX compiler picks it up from the `.mdx` documentation file.

When opened in the browser, the optimised SVG displays correctly. The visual bugs appear to only happen when it goes through the `mdx` compiler.

**Screenshots**
What it should look like:
<img width="812" alt="Screenshot 2019-11-19 at 17 16 20" src="https://user-images.githubusercontent.com/48090803/69169490-56a05b80-0af0-11ea-8b5d-f464b896f899.png">
What renders instead: 
<img width="859" alt="Screenshot 2019-11-19 at 15 45 58" src="https://user-images.githubusercontent.com/48090803/69169527-6881fe80-0af0-11ea-9e04-604fc3de5a29.png">

PR #474, opened 2019-12-05
Generate components from SVGs in doc site

thyhjwb6:
## Related issue
#437 

## Overview
Replaced the loading of inline SVGs with a build process which converts SVGs to React components to be added to the documentation.

## Reason
There is a dependency on `gatsby-plugin-react-svg` which has upstream issues with outputting SVGs that contains elements with the same ID. This leads to the document being badly structure and therefore issues with presenting SVGs.

There is the added benefit that this automates the manual task which is optimising SVGs.

## Work carried out
- [x] Add SVGR to build
- [x] Use React components
- [x] Tidy up package.json scripts
- [x] Tidy up docs site files
- [x] Remove `gatsby-plugin-react-svg` dependency

## Screenshot
Screenshot 2019-12-06 at 08 17 13

## Developer notes
Special attention should be given to [the styles which have been moved to `masthead.scss`](https://github.com/Royal-Navy/standards-toolkit/pull/474/files#diff-228f3718fb36a5e22854473054d1e78fR119). This is because the `masthead` is dependent on `site-logo.svg` which had embedded styles and SVGR could not convert it to a component without errors.


m7kvqbe1 (approved):
Looks good to me

PR #480, opened 2019-12-09
Upgrade dependencies

thyhjwb6:
## Related issue
#481 

## Overview
Upgrading dependencies

Note that `eslint-plugin-import` has not been upgraded due to [a bug](https://github.com/benmosher/eslint-plugin-import/issues/1558) which has been [fixed](https://github.com/benmosher/eslint-plugin-import/pull/1563) but not released.

Change also addresses [a security vulnerability in serialize-javascript](https://github.com/advisories/GHSA-h9rv-jmmf-4pgx).

A mistake was found in the `Button` documentation as `onClick` is not required.

## Reason
Security

## Work carried out
- [x] Upgrade dependencies
- [x] Address security vulnerability

## Screenshot
No visual difference.

## Developer notes
After making changes to `TextInput` and `TextArea`, these components needed refactoring in order to reduce code complexity.


packages/react-component-library/src/components/TabSet/TabSet.test.tsx
```diff
@@ -18,7 +18,9 @@ describe('TabSet', () => {
       beforeEach(() => {
         const props = {
           className: 'rn-tab-set--modifier',
-          onChangeCallback: () => {},
+          onChangeCallback: () => {
+            return true
```
m7kvqbe1, line 22: I assume the empty functions here are just being used to stop it from failing loudly where a prop is required? Is this change because of a new lint rule or were these warnings?

m7kvqbe1 (approved):
Few different things going on in this PR but looks good. Not quite sure where the problem came from with the empty anonymous functions?

Issue #467, opened 2019-12-04
Ensure that component APIs are consistent across components


m7kvqbe1:
Ensure that component APIs are consistent across components ahead of the 2.0.0 release.

PR #489, opened 2019-12-12
Refactor ButtonGroup component

thyhjwb6:
## Related issue
#467 

## Overview
Make `ButtonGroup` consistent with other components.

## Reason
Consistency

## Work carried out
- [x] Make consistent

## Screenshot
No visual difference.

### DIFF
--- a/packages/react-component-library/src/components/ButtonGroup/ButtonGroup.tsx
+++ b/packages/react-component-library/src/components/ButtonGroup/ButtonGroup.tsx
@@ -23,6 +23,12 @@ export const ButtonGroup: React.FC<ButtonGroupProps> = ({
       {React.Children.map(
         children,
         (child: React.ReactElement<ButtonGroupItemProps>) => {
+          if (child.props.size) {
+            console.warn(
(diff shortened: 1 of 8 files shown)


packages/react-component-library/src/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -27,19 +20,21 @@ export const ButtonGroup: React.FC<ButtonGroupProps> = ({
...
-        >
-          {label}
-        </Button>
-      ))}
+      {React.Children.map(
+        children,
+        (child: React.ReactElement<ButtonGroupItemProps>) => {
+          if (child.props.size) {
+            console.warn(
+              'Prop `size` on `ButtonGroupItem` will be replaced by `size` from `ButtonGroup`'
```
m7kvqbe1, line 28: Yep really like this - I think this is nice

m7kvqbe1 (approved):
This looks great to me. 

I'm trying to think of edge cases or maybe props in addition to size that might need the same treatment. Looks like it's only size though?

m7kvqbe1 (approved):

PR #490, opened 2019-12-12
[WIP] Refactor Nav component

thyhjwb6:
After discussion we are parking this refactor for now. This component is primarily being used for the docs site. Currently half of the nested item functionality lives in the component library and the other half is in the docs site.

Rather than implement this now, we are marking the docs for `Nav` as `draft` as we're not officially supporting the component.

If/when this is picked up again, there are some items outstanding:
- position of SVG arrows in items with nested nav items
- Storybook story component composition is not quite right as:
```
react.development.js?17ac:167 Warning: Each child in a list should have a unique "key" prop.

Check the render method of `storyFn`. See https://fb.me/react-warning-keys for more information.
    in NavItem
    in storyFn
    in ErrorBoundary
```
- more scenarios need covering in tests
- use of `uuid` for `key`
- replace strings with constants
- fix integration with docs site
- reuse shared `Nav` and `NavItem` types


packages/react-component-library/src/components/Nav/NavItem.tsx
```diff
@@ -1,37 +1,56 @@
...
-const NavItem: React.FC<NavItemProps> = ({ children, hasChildren = false }) => {
-  // @ts-ignore
-  const foundActive = !!Children.deepFind(children, ({ props }) => props.active)
-  const [isOpen, setIsOpen] = useState(foundActive && hasChildren)
+function hasActive(
+  children:
+    | React.ReactElement<NavItemProps>
+    | React.ReactElement<NavItemProps>[]
+) {
+  const flat = flatten(children as React.ReactElement<NavItemProps>[])
```
m7kvqbe1: Could we use this here now instead of lodash?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat

packages/react-component-library/src/components/Nav/NavItem.tsx
```diff
@@ -1,37 +1,56 @@
...
+  return some(flat, ({ props }) => props.isActive)
```
m7kvqbe1: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

thyhjwb6:
Closing this as anticipating it will be redundant once we replace the docs site.

PR #492, opened 2019-12-13
Refactor breadcrumbs

thyhjwb6:
## Related issue
#467 

## Overview
Refactors `Breadcrumbs` to fully utilise a JSX interface rather than injecting JSON.

## Reason
Consistency across components.

## Work carried out
- [x] Refactor
- [x] Update docs


packages/react-component-library/src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react'
+import classNames from 'classnames'
+
+import { BreadcrumbProps } from '.'
```
m7kvqbe1: I'm a bit confused by this import? Isn't this the interface defined below?

m7kvqbe1: Ignore me I re-read it and makes [s]en[s]e now.

m7kvqbe1 (approved):
Looks good to me

PR #494, opened 2019-12-16
Refactor Masthead

thyhjwb6:
## Related issue
#467 

## Overview
Refactor the `Masthead` component so that the interface is consistent.

This change includes:
- adding shared types for nav as they all use a similar pattern
- making the component more composable so that the consumer can effectively supply their own nav, 
nav items or user
- consolidating repeated code ie notifications in the stories

## Reason
Part of making interfaces consistent before launching v2.

## Work carried out
- [x] Add `Nav` and `NavItem` shared types
- [x] Refactor `Masthead`

## Screenshot

### DIFF
--- a/packages/css-framework/src/components/_scrollable-nav.scss
+++ b/packages/css-framework/src/components/_scrollable-nav.scss
@@ -13,7 +13,7 @@ $btn-focus-width: 0.2rem;
   margin: 0;
 }
 
-.rn-scrollable-nav__link {
+.rn-scrollable-nav__item > a {
   display: flex;
   align-items: center;
   flex-wrap: nowrap;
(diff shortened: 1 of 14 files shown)


packages/css-framework/src/components/_scrollable-nav.scss
```diff
@@ -13,32 +13,34 @@ $btn-focus-width: 0.2rem;
...
-    text-decoration: none;
+    &:hover {
+      border-bottom: 4px solid color(primary, 600);
+      text-decoration: none;
+    }
+    &:focus {
+      box-shadow: 0 0 0 $btn-focus-width rgba(color(primary, 700), 0.5);
+    }
+
+    &:hover {
```
m7kvqbe1: Is this a duplicate selector?

m7kvqbe1 (commented):
Looks good.

Are the intermediate components there to more strictly enforce what Masthead should be composed of? Does this in turn create a tighter coupling between Masthead and some of the atoms like Avatar?

m7kvqbe1 (approved):

PR #506, opened 2020-01-06
Refactor Sidebar

thyhjwb6:
## Related issue
#467 

## Overview
Refactor the `Sidebar` component so that the interface is consistent.

This change includes:
- reusing shared types for nav as they all use a similar pattern
- making the component more composable so that the consumer can effectively supply their own nav, 
nav items or user
- tidying up
- removing the unused types as a result of adding shared types for nav, nav item
- adding a helper for `console.warn` when overwriting props

## Reason
Part of making interfaces consistent before launching v2.

## Work carried out
- [x] Refactor `Sidebar`
- [x] Tidy up
- [x] Reused shared types for `TabNav`
- [x] Add `console.warn` helper

## Screenshot

## Developer notes
There were not unit tests for this component so these have been backfilled with the refactor.


packages/react-component-library/src/fragments/Sidebar/Sidebar.test.tsx
```diff
@@ -0,0 +1,249 @@
...
+
+        it('should not have the `is-open` class', () => {
+          expect(wrapper.getByTestId('sidebar').classList).not.toContain(
+            'is-open'
+          )
+        })
+      })
+    })
+  })
+
```
m7kvqbe1, line 218: Extra line here

packages/react-component-library/src/fragments/Sidebar/Sidebar.test.tsx
```diff
@@ -0,0 +1,249 @@
...
+      consoleWarnSpy = jest.spyOn(global.console, 'warn')
+
+      wrapper = render(
+        <Sidebar
+          nav={(
+            <SidebarNav>
+              <SidebarNavItem
+                link={<Link href="/">Home</Link>}
+                onClick={() => {
+                  /* oops */
```
m7kvqbe1: Is the comment needed to get past a lint rule?

packages/react-component-library/src/fragments/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+export const Sidebar: React.FC<SidebarProps> = ({
+  nav,
+  notifications,
+  hasUnreadNotification,
+  user,
+}) => {
+  const [open, setOpen] = useState()
+  const classes = classNames('rn-sidebar', {
+    'is-open': open, // docs site
```
m7kvqbe1: Is this there to set an initial open state in docs-site example?

packages/react-component-library/src/helpers.ts
```diff
@@ -0,0 +1,9 @@
+function warnIfOverwriting<P>(props: P, propertyName: string, componentName: string) {
```
m7kvqbe1, line 1: Not for now - but maybe as the library evolves we could wrap console.log so we can enable / disable logging based on message types / change verbosity. Something to think about in the future - might be completely overkill.

m7kvqbe1 (approved):
Looks great to me  - just a few small comments

PR #524, opened 2020-01-14
Fix yarn types not generating dist types correctly

thyhjwb6:
## Related issue
NA

## Overview
Addresses silent failing of `yarn types`.


m7kvqbe1 (approved):
🕵

Issue #505, opened 2019-12-19
Modal - Can display modal only once


jk68mnpo:
**Description**
Can display modal only once (if isOpen=true on initialisation) and successfully close. After it is closed, I can no longer reopen it as expected.

**Example Code Snippet**

```javascript
const [modalOpen, setModalOpen] = useState(false)
 
 return (
  <Button onClick={(_ => setModalOpen(true)}>Click me</Button>
  <Modal
    title="Example"
    isOpen={modalOpen}
    onClose={_ => setModalOpen(false)}
  />
 )
 ```

m7kvqbe1:
This is a known issue that needs resolving. The internal state in the modal needs removing and the show/hide behaviour should be delegated to the consumer.

A workaround for now is to use routes:

https://github.com/Royal-Navy/coffee/blob/a4234b192553bc12dbcc02496e093909bf4fa260/src/pages/Subscriptions/index.jsx#L57

PR #528, opened 2020-01-16
Fix modal hide show from side effect

thyhjwb6:
## Related issue
#505 

## Overview
If the modal visibility is being controlled by a side effect then the rendered output will not change.

## Reason
Components need to use `useEffect` hook when side effects update state.

## Work carried out
- [x] Fix

## Screenshot
Visual appearance should not change.

## Developer notes
This is the same fix for the recent table bug.

### DIFF
--- a/packages/react-component-library/src/components/Dialog/Dialog.test.tsx
+++ b/packages/react-component-library/src/components/Dialog/Dialog.test.tsx
@@ -47,7 +47,7 @@ describe('Modal', () => {
       })
 
       it('should apply the `rn-dialog` class', () => {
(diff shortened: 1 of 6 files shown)


packages/react-component-library/src/components/Modal/Modal.stories.tsx
```diff
@@ -1,6 +1,6 @@
 import { action } from '@storybook/addon-actions'
 import { storiesOf } from '@storybook/react'
-import React from 'react'
+import React  from 'react'
```
m7kvqbe1: Extra whitespace here

```suggestion
import React from 'react'
```

m7kvqbe1 (approved):

PR #536, opened 2020-01-21
Use ...rest to drill props through Select wrapper

m7kvqbe1:
The Select component simply wraps a library. Additional props were not being respect by the wrapper.

This fix exposes additional functionality to the end user (in particular `isDisabled`).

### DIFF
--- a/packages/react-component-library/src/components/Select/Select.stories.tsx
+++ b/packages/react-component-library/src/components/Select/Select.stories.tsx
@@ -16,11 +16,18 @@ const options = [
 ]
 
 stories.add('Default', () => (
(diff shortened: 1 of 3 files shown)


thyhjwb6:
Nice I think we have the issue on other components.

Is it possible to write a test for this?

m7kvqbe1:
Test added for the `...rest` - also extended the wrapped components interface.

I've had to use `querySelector` on the container because there's no markup to attach a `data-testid` and I can't modify the upstream react-select component.

thyhjwb6 (approved):

PR #539, opened 2020-01-21
Improve datepicker disabled styling

m7kvqbe1:
Applies correct styling, adds story and tests and renames prop to `isDisabled`.

### DIFF
--- a/packages/react-component-library/src/components/DatePicker/DatePicker.tsx
+++ b/packages/react-component-library/src/components/DatePicker/DatePicker.tsx
@@ -38,7 +38,6 @@ export interface DatePickerProps extends ComponentWithClass {
   label?: string
   onBlur?: (event: React.FormEvent) => void
   onChange?: (data: StateObject) => void
-  disabled?: boolean
   placement:
     | typeof DATEPICKER_PLACEMENT.ABOVE
     | typeof DATEPICKER_PLACEMENT.BELOW
@@ -47,6 +46,7 @@ export interface DatePickerProps extends ComponentWithClass {
   startDate?: Date
   endDate?: Date
   isRange?: boolean
+  isDisabled?: boolean
 }
 
 function transformDates(startDate: Date, endDate: Date) {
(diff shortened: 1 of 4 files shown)


packages/react-component-library/src/components/DatePicker/DatePicker.tsx
```diff
@@ -47,6 +46,7 @@ export interface DatePickerProps extends ComponentWithClass {
   startDate?: Date
   endDate?: Date
   isRange?: boolean
+  isDisabled?: boolean
```
thyhjwb6: Not for now but we should consider reordering these props to be alphabetical.

packages/react-component-library/src/components/DatePicker/DatePicker.tsx
```diff
@@ -69,11 +69,11 @@ export const DatePicker: React.FC<DatePickerProps> = ({
   label = 'Select Date',
   onBlur,
   onChange,
-  disabled,
   placement = DATEPICKER_PLACEMENT.BELOW,
   startDate,
   endDate,
   isRange,
+  isDisabled = false,
```
thyhjwb6: ~`isDisabled` will default to `false` if it is not set, ie it is not nullable~

Sorry I just reread that.. regardless, `isDisabled` does not need to be set as it will evaluate the same.

packages/react-component-library/src/components/DatePicker/DatePicker.tsx
```diff
@@ -200,7 +200,7 @@ export const DatePicker: React.FC<DatePickerProps> = ({
                 </NavButton>
 
                 <div className={gridClasses}>
-                  {activeMonths.map((month, index) => (
+                  {activeMonths.map(month => (
```
thyhjwb6: I suppose an option could have been:

```
{activeMonths.map({ year, month } => (
  <Month
    key={`${year}-${month}`}
    year={year}
    month={month}
    firstDayOfWeek={firstDayOfWeek}
  />
))}
```

thyhjwb6 (approved):
Looks good, comments are just observations and don't need to be actioned.

packages/react-component-library/src/components/DatePicker/DatePicker.tsx
m7kvqbe1: Yeah I see it'll still be falsey i'll remove the default

packages/react-component-library/src/components/DatePicker/DatePicker.tsx
m7kvqbe1: Yeah cool - i'll destructure I was deleting the rogue param the IDE flagged up

PR #543, opened 2020-01-21
DatePicker overlay appears behind modal

m7kvqbe1:
Fix styling issues and overlay issue preventing elements below being targeted.


m7kvqbe1:
I tried to write a test to target the tether component to check for the `is-visible` - but couldn't figure out how to target it (even with the container querySelector trick) as the library took it out of the regular DOM flow.

thyhjwb6 (approved):

Issue #412, opened 2019-11-08
Integrate Conventional Commits


thyhjwb6:
- [ ] Add commit linting
- [ ] Add semantic release

PR #619, opened 2020-02-04
Add linting for commit messages

thyhjwb6:
## Related issue
#412 

## Overview
Adds linting for commit messages using conventional commit format. Linting runs both locally and within CI process.

## Reason
This is the first step towards automating release notes as it will ensure the commit messages are in a readable format.

## Work carried out
- [x] Add local commit lint
- [x] Add script for CI commit lint

## Developer notes
A new build script was introduced to handle commit linting during CI.

### DIFF
--- a/scripts/commitlint/index.js
+++ b/scripts/commitlint/index.js
@@ -1,0 +1,50 @@
+const { Octokit } = require('@octokit/rest')
+const lint = require('@commitlint/lint')
+
+const { rules } = require('../../commitlint.config')
+
+function getPrOptions() {
+  const prUrl = process.argv[2]
+  const prUrlParts = prUrl.split('/')
+
+  return {
+    owner: prUrlParts[prUrlParts.length - 4],
+    repo: prUrlParts[prUrlParts.length - 3],
+    pull_number: prUrlParts[prUrlParts.length - 1],
+  }
+}
+
+function getCommitsByPr(options) {
+  const github = new Octokit()
+  return github.pulls.listCommits(options)
+}
+
+function lintCommits(commits) {
+  const promises = commits.data.map(commit => {
+    return lint(commit.commit.message, rules)
+  })
+
+  return Promise.all(promises)
+}
+
+(async function runCommitLint() {
+  try {
+    const prOptions = getPrOptions()
+    const prCommits = await getCommitsByPr(prOptions)
+    const lintResults = await lintCommits(prCommits)
+
+    const resultsWithErrors = lintResults.filter(result => !result.valid)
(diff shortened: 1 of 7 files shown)


scripts/commitlint/index.js
```diff
@@ -0,0 +1,51 @@
...
+}
+
+const runCommitLint = async () => {
+  try {
+    const prNumber = getPrNumber()
+    const prCommits = await getCommitsByPr(prNumber)
+    const lintResults = await lintCommits(prCommits)
+
+    const resultsWithErrors = lintResults.filter(result => !result.valid)
+    resultsWithErrors.forEach(resultsWithError => {
```
m7kvqbe1: The naming of the lambda param here (`resultsWithErrors`) confused me a bit when I was scanning.

Maybe we could destructure the error bag out of it?
```suggestion
    resultsWithErrors.forEach(({ errors }) => {
```

scripts/commitlint/index.js
```diff
@@ -0,0 +1,51 @@
...
+  const prUrl = process.argv[2]
+  const prUrlParts = prUrl.split('/')
+
+  return prUrlParts[prUrlParts.length - 1]
+}
+
+function getCommitsByPr(number) {
+  const github = new Octokit()
+  return github.pulls.listCommits({
+    owner: 'Royal-Navy',
```
m7kvqbe1: Are we able to get `owner` and `repo` from somewhere too so it doesn't need to be hardcoded?

scripts/commitlint/index.js
```diff
@@ -0,0 +1,51 @@
...
+    repo: 'standards-toolkit',
+    pull_number: number,
+  })
+}
+
+function lintCommits(commits) {
+  const promises = commits.data.map(commit => {
```
m7kvqbe1, line 23:
```suggestion
  const promises = commits.data.map(({ commit: { commit: message }}) => {
```

Can take it or leave it with this kind of thing - arguably not better, just different.

m7kvqbe1 (commented):
Looks great to me - just a couple of comments.

Loving Octokit! 

:octocat:

scripts/commitlint/index.js
```diff
@@ -0,0 +1,51 @@
...
+      })
+    })
+
+    process.exit(resultsWithErrors.length ? 1 : 0)
+  } catch (e) {
+    console.error('Error running commitlint', e)
+  }
+}
+
+runCommitLint()
```
m7kvqbe1: Maybe we could use an IIFE instead of this line?

m7kvqbe1 (approved):

Issue #608, opened 2020-01-30
withFormik puts the errors outside the component


antictor:
the if a field has an error the withFormik function puts the error message below the field, but does not do this within the component.

```
<FormComponent
        {...field}
        {...props}
        className={formComponentClassNames}
      />
      {hasError && (
        <div className="rn-form__invalid-feedback" data-testid="error">
          {errors[field.name]}
        </div>
      )}
```

In most cases this is ok, but in others this can cause the error to appear out of line with the field
<img width="662" alt="Screenshot 2020-01-30 at 21 24 59" src="https://user-images.githubusercontent.com/48056118/73491803-484fa500-43a7-11ea-8f52-b66bc334abd0.png">

Instead the error should be passed as a property to the 'FormComponent' so that the component can render the error within it's wrapper and control the layout.

PR #643, opened 2020-02-18
Wrap Formik error with component

m7kvqbe1:
## Related issue

https://github.com/Royal-Navy/standards-toolkit/issues/608

## Overview

Wrap the error in with the enhanced Formik component instead of returning a fragment.

## Reason

>If a field has an error the withFormik function puts the error message below the field, but does not do this within the component.

## Work carried out

- [x] Wrap error in with enhanced component
- [x] Test in dummy application
- [x] Display Storybook validation error on mount (for visual regression test)


thyhjwb6 (commented):
This looks good. As discussed some test coverage at least through Chromatic could work well.

thyhjwb6 (approved):

Issue #509, opened 2020-01-06
Select Component - On viewing the inputs options, the label of the input below shows through


jk68mnpo:
**Describe the bug**
On viewing the inputs options, the label of the input below shows through

**To Reproduce**
1. Create a form with the select component as the first input and a text/number input directly below it.
2. When there is 2 or more options in the option dropdown, the label of the input below can still be seen.

**Environment**
This is the case on:
- Firefox v69.0.1
- Chrome v69.0.81

Issue #390, opened 2019-10-23
Component CSS tweaks to match Sketch file


hxltrhuxze:
The `v2.0.0` update to the SCSS framework provides a great opportunity to fine tweak the SCSS of each React component to ensure they match as closely as possible to the Sketch designs.

`v2.0.0` modifies the base Typography, Spacing, and Breakpoint scales that SCSS framework is built on. This gives us the ability to fine tune the React components, producing components that are much closer to their static Sketch twins.

Components that need updating:
- [x] Alert
- [x] Avatar
- [x] Badge
- [x] Breadcrumbs
- [x] Button
- [x] Button Group
- [x] Card Frame
- [x] Date Picker
- [x] Dialog
- [x] Drawer
- [x] Checkbox
- [x] Radio Buttons
- [x] Select
- [x] Masthead
- [x] Modal
- [x] Pagination
- [x] Panel
- [x] Phase Banner
- [x] Popover
- [ ] Sidebar
- [x] Tab Set
- [x] Table
- [x] Toast
- [x] Tooltip

---

- Dropdowns
- Number Input
- Text Input
- Textarea
- Switch

PR #716, opened 2020-03-09
Select component improvements

### LINKED ISSUE #308: Select size changes slightly upon open
When a select component is expanded the content below it moves up a pixel.

hxltrhuxze:
#390 
#308 
#509 

# Select Improvements
- [x] Add `xsmall` badges to be used in the Select component
- [x] Stop select changing heights on focus
- [x] Update styles to match Sketch library
- [x] Stop non-active Select labels rendering above other components


thyhjwb6 (approved):
These changes and commit messages look good.

An option for one of your commit messages would be:

```
fix(Select): Stack context to stop label render bug
```

Rather than the final merge commit, an option could have been:

```
git checkout master
git pull
git checkout bugfix/SelectInput
git rebase origin/master
git push origin bugfix/SelectInput
```

I see that there were Chromatic snapshot updates. Are we happy that the issue with stacking `select` elements is covered in these snapshots?

hxltrhuxze:
@thyhjwb6 Thanks for the feedback.

Regarding snapshot tests: I'm unsure how we would use Chromatic to test for this. The bug originated when the [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context) for the Select wasn't scoped to the component. This meant the Label would be visible above _other_ active selects. We can make a storybook story for two selects, however I don't know if Chromatic supports the ability to trigger different component states? (e.g. triggering the focus state for the top Select to see if the label from the bottom select bleeds through)

PR #727, opened 2020-03-13
feat(button group): use z-index stacking for hover states

hxltrhuxze:
## Related issue

#390 

## Overview

Updates the Button Group component to ensure parity with the Sketch file.

## Work carried out

- [x] Swaps out the removal of borders in favour of z-index stacking for hovered/active items. This ensures we can have a border on every side of the hovered button.
- [x] Renames `_button.scss` to `_btn.scss` to keep consistency with the button group and the underlying class names.


thyhjwb6 (approved):
👍 nice commit message

PR #766, opened 2020-03-25
Create upnext.yml

### LINKED ISSUE #767: Project board automation - Up next
Add GitHub action to automate movement of candidate tickets into the 'Ready' column on the project board

elglup:
## Overview

Adding a GitHub action to automate movement of issues labelled 'Up next' into the ready column on the project board

### DIFF
--- a/.github/workflows/upnext.yml
+++ b/.github/workflows/upnext.yml
@@ -1,0 +1,14 @@
+on:
+  issues:
+    types: [labeled]
+jobs:
+  Move_Labeled_Issue_On_Project_Board:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: konradpabjan/move-labeled-or-milestoned-issue@v2.0
+      with:
+        action-token: "${{ secrets.GHA_UPNEXT_TOKEN }}"
+        project-url: "https://github.com/orgs/Royal-Navy/projects/6"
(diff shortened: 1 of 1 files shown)


.github/workflows/upnext.yml
```diff
@@ -0,0 +1,14 @@
+on:
+  issues:
+    types: [labeled]
+jobs:
+  Move_Labeled_Issue_On_Project_Board:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: konradpabjan/move-labeled-or-milestoned-issue@v2.0
+      with:
+        action-token: "${{ secrets.GHA_UPNEXT_TOKEN }}"
```
thyhjwb6, line 10: Is the token name quite specific and maybe we need something more generic for other actions?

thyhjwb6 (commented):
Small comment and noticed the commit messages needs the `chore` prefix. Not sure why this hasn't been picked up by the linter.

.github/workflows/upnext.yml
elglup: The token is deliberately specific as it has fine grained read only permissions just for this task. Other actions will use a different token according to the type of action and required privileges.

thyhjwb6 (approved):
👍

PR #785, opened 2020-03-30
feat(component library): Icon placement left of button text

### LINKED ISSUE #333: You can’t place icons to the left of the button text
You can’t place icons to the left of the button text without overriding styles.

This appears to be a pattern based on the Coffee app mocks.

m7kvqbe1:
## Related issue

Closes #333 

## Overview

Add the ability to place an icon to the left of the text within the Button component.

## Reason

>You can’t place icons to the left of the button text without overriding styles.
>
>This appears to be a pattern based on the Coffee app mocks.

## Work carried out

- [x] Refactor the button styles to avoid nesting
- [x] Add ability to place icon to the left of Button
- [x] Update docs-site documentation

## Screenshot

Screenshot 2020-03-30 at 10 56 10

### DIFF
--- a/packages/react-component-library/src/components/Button/constants.ts
+++ b/packages/react-component-library/src/components/Button/constants.ts
@@ -13,6 +13,11 @@ const BUTTON_VARIANT = {
 
 const BUTTON_COLOR = {
   DANGER: 'danger',
-}
+} as const
 
-export { BUTTON_SIZE, BUTTON_VARIANT, BUTTON_COLOR }
+const BUTTON_ICON_POSITION = {
+  LEFT: 'left',
+  RIGHT: 'right'
+} as const
+
+export { BUTTON_SIZE, BUTTON_VARIANT, BUTTON_COLOR, BUTTON_ICON_POSITION }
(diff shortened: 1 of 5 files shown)


packages/react-component-library/src/components/Button/constants.ts
```diff
@@ -15,4 +15,9 @@ const BUTTON_COLOR = {
   DANGER: 'danger',
 }
 
-export { BUTTON_SIZE, BUTTON_VARIANT, BUTTON_COLOR }
+const BUTTON_ICON_POSITION = {
```
thyhjwb6, line 18: This needs to be:

```javascript
const BUTTON_ICON_POSITION = {
  LEFT: 'left',
  RIGHT: 'right'
} as const
```

Otherwise you don't get an error with arbitrary text.

Screenshot 2020-04-02 at 09 55 52

packages/react-component-library/src/components/Button/constants.ts
```diff
@@ -15,4 +15,9 @@ const BUTTON_COLOR = {
   DANGER: 'danger',
 }
```
thyhjwb6: I notice that `as const` is missing here

thyhjwb6 (commented):
Looks good, just a minor comment with the `constants`

thyhjwb6 (approved):
👍

PR #840, opened 2020-04-14
refactor(component library): Remove UUID React key usage

### LINKED ISSUE #771: Remove use of uuid for component keys
**Description (Value):**
 It's an anti-pattern which will be causing performance issues

**Acceptance Criteria:**

- [x] No longer referencing UUID

m7kvqbe1:
## Related issue

Closes #771 

## Overview

Stop using UUID for generating React keys.

## Reason

>It's an anti-pattern which will be causing performance issues.

## Work carried out

- [x] Add getKey function to helpers
- [x] Refactor where UUID is used

### DIFF
--- a/packages/react-component-library/src/helpers.ts
+++ b/packages/react-component-library/src/helpers.ts
@@ -1,3 +1,7 @@
+function getKey(prefix: string, index: string | number): string {
+  return `${prefix}-${index}`.replace(/\s/g,'')
+}
+
 function warnIfOverwriting<P>(props: P, propertyName: string, componentName: string) {
   if (props[propertyName]) {
     console.warn(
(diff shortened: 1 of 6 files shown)


packages/react-component-library/src/helpers.ts
```diff
@@ -6,4 +6,8 @@ function warnIfOverwriting<P>(props: P, propertyName: string, componentName: str
   }
 }
 
-export { warnIfOverwriting }
+function getKey(prefix: string, index: string | number): string {
+  return `${prefix}-${index}`.replace(/\s/g,'')
+}
+
+export { warnIfOverwriting, getKey }
```
thyhjwb6: It could be a nice touch if these were alphabetical

thyhjwb6 (commented):
I am seeing other areas where `uuidv4` is being referenced ie `Radio`

Can we also remove the reference from `package.json`?

thyhjwb6 (approved):
After review it is clear that the other areas `uuid` is used are not an issue.

PR #894, opened 2020-04-28
test(select): Backfill tests

### LINKED ISSUE #862: Increase Select test coverage
- [ ] Increase test coverage > 90%

thyhjwb6:
## Related issue
Closes #862 

## Overview
Backfill tests for `Select` component.

## Reason
Get test coverage up.

## Work carried out
- [x] Add tests

### DIFF
--- a/packages/react-component-library/src/components/Select/Select.test.tsx
+++ b/packages/react-component-library/src/components/Select/Select.test.tsx
@@ -1,0 +1,89 @@
+import React from 'react'
+import '@testing-library/jest-dom/extend-expect'
+import { render, RenderResult, fireEvent } from '@testing-library/react'
+
+import { Select } from '.'
+
+describe('Select', () => {
+  let wrapper: RenderResult
+  let onChangeSpy: (event: any) => void
+
+  describe('when provided with all props', () => {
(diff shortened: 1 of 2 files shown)


packages/react-component-library/src/components/Select/Select.test.tsx
```diff
@@ -0,0 +1,89 @@
+import React from 'react'
+import '@testing-library/jest-dom/extend-expect'
+import { render, RenderResult, fireEvent } from '@testing-library/react'
+
+import { Select } from '.'
+
+describe('Select', () => {
+  let wrapper: RenderResult
+  let onChangeSpy: (event: any) => void
```
m7kvqbe1, line 9: Can we be more explicit about that arg type?

m7kvqbe1 (approved):
LGTM - just left a quick comment about type.

packages/react-component-library/src/components/Select/Select.test.tsx
thyhjwb6: The prop on the interface is defined as below. We should not change this until we address all raised events in `v3`.

```javascript
onChange?: (event: any) => void
```

PR #1014, opened 2020-05-29
fix(Select): Add ability to drill data-testid

### LINKED ISSUE #989: Arbitrary props need to be drilled in Select component
>how can i get hold of the select component and fire change event?
>
>the data-testid i am giving to select component is getting ignored..
>
>screen.getByTestId is not working
>
>but if i do screen.getByLabelText I am getting input html component. But when I fire change event doesnt get to my change handler...

thyhjwb6:
## Related issue
Fixes #989 

## Overview
Consumer of the library requires the ability to pass in a custom `data-testid` which is particularly evident when there are multiple `Select` components on a page.

## Reason
Convenience for identifying `Select` components when writing tests.

## Work carried out
- [x] Drill `data-testid`

### DIFF
--- a/packages/react-component-library/src/components/Select/Input.tsx
+++ b/packages/react-component-library/src/components/Select/Input.tsx
@@ -1,12 +1,12 @@
 import React from 'react'
-import { components } from 'react-select'
-import { InputProps } from 'react-select/src/components/Input'
+import get from 'lodash/get'
+import { components, InputProps } from 'react-select'
 
 export interface SelectInputProps extends InputProps {
   id?: string
 }
 
-export const Input: React.FC<SelectInputProps> = props => (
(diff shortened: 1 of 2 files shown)


packages/react-component-library/src/components/Select/Input.tsx
```diff
@@ -17,6 +16,11 @@ export const Input: React.FC<SelectInputProps> = props => (
         {props['aria-label']}
       </label>
     )}
-    <components.Input data-testid="react-select-vendor-input" {...props} />
+    <components.Input
+      data-testid={
+        props.selectProps['data-testid'] || 'react-select-vendor-input'
```
m7kvqbe1: Nice - clever trick

m7kvqbe1 (approved):
LGTM

PR #1024, opened 2020-06-01
Fix extra ellipsis when moving between page numbers in pagination

### LINKED ISSUE #947: Pagination scrolling adds extra ellipsis
The pagination will increase in width when navigating back and forth.

To reproduce:

1. Load the [Pagination component](https://storybook.royalnavy.io/?path=/story/pagination--default)
2. Repeatedly click `1` and then `5`. This will start to add multiple ellipsis to the pagination.

thyhjwb6:
## Related issue
Fixes #947 

## Overview
Add fix so that extra ellipsis does not appear when scrolling between a set of page numbers that need the ellipsis vs those that don't. The issue is being cause by duplicate `key`.

## Reason
Confusing for the user.

## Work carried out
- [x] Fix
- [x] Refactor

### DIFF
--- a/packages/react-component-library/src/components/Pagination/Pagination.test.tsx
+++ b/packages/react-component-library/src/components/Pagination/Pagination.test.tsx
@@ -1,176 +1,216 @@
 import React from 'react'
 import '@testing-library/jest-dom/extend-expect'
-import { render, fireEvent, RenderResult } from '@testing-library/react'
+import { render, RenderResult } from '@testing-library/react'
 
 import { Pagination } from '.'
 
-afterEach(() => jest.clearAllMocks())
-
 describe('Pagination', () => {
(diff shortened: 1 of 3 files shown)


packages/react-component-library/src/components/Pagination/index.tsx
```diff
@@ -36,43 +36,43 @@ const Pagination: React.FC<PaginationProps> = ({
             disabled={currentPage === 1}
             className="rn-pagination__button"
             onClick={() => {
-              changePage('previous')
+              changePage(currentPage - 1)
```
m7kvqbe1: 👍

m7kvqbe1 (approved):
LGTM - debatable about reduction in complexity but nicer seeing it in a more functional style.

PR #1027, opened 2020-06-02
style(ESLintConfigReact): Update configuration

m7kvqbe1:
Review and update published ESLint config in line with that used by both downstream application teams and the react-component-library package.


thyhjwb6 (approved):
👍

PR #1033, opened 2020-06-03
Add condensed variation of NumberInput

### LINKED ISSUE #981: Condensed number stepper without label
### Description
In some scenarios, it is preferable to have a number stepper (and potentially other form elements) without a label.

In our exemplar project, number steppers appear in a table, and repeating the label each time is unnecessary.

It is also preferable to condense the size of the number stepper so it is more compact.

number-stepper-without-label

thyhjwb6:
## Related issue
Closes #981 

## Overview
Add `isCondensed` prop to condense the `NumberInput` component.

## Reason
Required for when presenting `NumberInput` in a `table`.

## Work carried out
- [x] Add prop


m7kvqbe1 (approved):
LGTM

Issue #990, opened 2020-05-20
Make field errors accessible and fixable


antictor:
When an error occurs on a field the error is displayed below the field but there is no markup to allow a user with a screen reader to know which field the error is associated with. This also makes it very hard to programatically test these errors too as you have no way to know that the error is associated with the input due to it's position in the DOM

```
<div class="formik-form-component formik-form-component--TextInput">
  <div class="rn-textinput is-invalid">
    <div class="rn-textinput__outer-wrapper">
    <div class="rn-textinput__input-wrapper">
      <label class="rn-textinput__label" for="...">Start date:</label>
      <input 
           class="rn-textinput__input" 
           id="..." 
           name="startDate" 
           placeholder="" 
           type="datetime-local" value=""
        >
      </div>
    </div>
  </div>
  <div class="rn-form__invalid-feedback">The plan must have a start date</div>
</div>
```

PR #1036, opened 2020-06-04
feat(Accessibility): Associate error with field

thyhjwb6:
## Related issue
#990 

## Overview
When a form field is in an error state, the input needs aria-invalid and aria-describedby which is set to the ID of the div containing the error.

## Reason
Screen readers will not work when in an error state.

## Work carried out
- [x] Associate error with field
- [x] Refactor


m7kvqbe1 (approved):
LGTM

packages/react-component-library/src/enhancers/withFormik.tsx
```diff
@@ -16,7 +17,7 @@ export const withFormik = (FormComponent: React.FC<any>) => ({
   className,
   ...props
 }: FormikProps) => {
-  const hasError = touched[field.name] && errors[field.name]
+  const hasError = !!(touched[field.name] && errors[field.name])
```
antictor: Maybe not one for here but one comment I got from our designer was he didn't want to see errors until the first submit, but this is outside the scope of this

packages/react-component-library/src/enhancers/withFormik.tsx
```diff
@@ -27,15 +28,25 @@ export const withFormik = (FormComponent: React.FC<any>) => ({
...
-        {...field}
-        {...props}
-        className={formComponentClassNames}
-      />
+      <FormComponent {...formComponentProps} />
       {hasError && (
-        <div className="rn-form__invalid-feedback" data-testid="error">
+        <div
+          className="rn-form__invalid-feedback"
+          data-testid="error"
```
antictor: Can we include the field name in the test id? When I'm writing integration tests for a form I want to be able to test that an error for a particular field is raised and check it's contents.

packages/react-component-library/src/enhancers/withFormik.tsx
```diff
@@ -27,15 +28,25 @@ export const withFormik = (FormComponent: React.FC<any>) => ({
     'is-invalid': hasError,
   })
 
+  const errorId = `${field.name}-error`
```
antictor: Isn't there be a chance that two fields could have the same name?

packages/react-component-library/src/enhancers/withFormik.tsx
thyhjwb6: We are stripping out `data-testid` with babel. If this is Cypress could you do something like `<FormikField data-testid="id" />` and then `cy.get('[data-testid="id"]').siblings()`?

m7kvqbe1:
re: Slack discussion

It's probably a good idea for us to spike the creation of some different forms to identify potential design concerns with form handling (across the library as a whole).

This looks good but we now have some nuance and complexity is growing. Consumption requires knowledge of both Formik and how we support its use. 

Dedicated documentation for 'forms' with working examples is needed.

PR #1041, opened 2020-06-05
fix(Modal): Fix font weight and button alignment

hxltrhuxze:


packages/react-component-library/src/components/Modal/Footer.tsx
```diff
@@ -13,7 +13,14 @@ export const Footer: React.FC<FooterProps> = ({
   tertiaryButton,
 }) => {
   return (
-    <footer className="rn-modal__footer" data-testid="modal-footer">
+    <footer
+      className={
+        tertiaryButton
```
m7kvqbe1: We try to use the classNames utility instead of ternaries. This could become:

```javascript

// ...

const classes = classNames('rn-modal__footer', {
  'rn-modal__footer--has-tertiary': tertiaryButton
})

<footer className={classes} ...
```

m7kvqbe1 (changes requested):
Small comment about applying dynamic css classes.

packages/react-component-library/src/components/Modal/Footer.tsx
hxltrhuxze: Thought I may have strayed from convention there...

m7kvqbe1 (approved):
LGTM - I assume it's covered by a Storybook story?

PR #1042, opened 2020-06-05
feat(CRATemplateRoyalNavy): Create CRA template

### LINKED ISSUE #808: Custom CRA template

m7kvqbe1:
## Related issue

Closes #808 

## Overview

Create and publish custom CRA template for generating application boilerplate.

## Reason

> Application developers need a quick and easy way of getting started.

## Work carried out

- [x] Create `@royalnavy/cra-template-royalnavy` package
- [x] Add automation for publishing
- [x] Add new label to GitHub
- [x] Test by generating new applications from template

## Developer notes

You can test the template locally like so:

```
npx create-react-app my-app --template file:../path/to/your/template/cra-template-[template-name]
```


m7kvqbe1:
I need to publish a pre-release of this package in order to test it in the closed environment.

antictor:
Can you allow the dev to pass the --use-npm flag like the regular one? Some devs may want to use npm, I know I do for projects that don't require workspaces

packages/cra-template-royalnavy/template/.yarnrc
```diff
@@ -0,0 +1 @@
+registry "http://mirror.nelson.rn:8081/repository/npm-proxy"
```
antictor: I would leave this out. Light side apps will report this as an error, dark side apps have this set in the users home folder and its already set in all the docker images.

antictor (changes requested):
npm start fails due to missing Typescript.
yarnrc should be removed
Like CRA it would be better to give the option to use npm instead of yarn
After installing Typescript it then fails due to no sass.

Needs testing before a PR is raised.

packages/cra-template-royalnavy/template/.yarnrc
m7kvqbe1: Both TypeScript and SASS are included as dependencies - i've just removed my global deps and I was still able to successfully generate and build a working boilerplate application.

Could you post a screenshot of the error you are getting re: missing deps.

m7kvqbe1: `.yarnrc` has been removed.

antictor (approved):

PR #1108, opened 2020-06-12
Add unit to NumberInput

### LINKED ISSUE #978: Number stepper with unit
### Description
A need has been identified to have a number stepper component with the unit displayed inline with the value, as opposed to in a separate box at the end. This will allows the user to clearly see the unit associated with the value that they are entering.

number-stepper

thyhjwb6:
## Related issue
Closes #978 

## Overview
Adds the ability to specify a unit before or after the actual number valiue.

## Reason
Required for an exemplar and it's useful for other consumers.

## Work carried out
- [x] Add unit after
- [x] Add unit before
- [x] Refactor code

### Unit after
after

### Unit before
before

## Developer notes
A request was made to set the unit as neutral 400 but this is not possible without significant rework.

### DIFF
--- a/packages/react-component-library/src/components/NumberInput/NumberInput.tsx
+++ b/packages/react-component-library/src/components/NumberInput/NumberInput.tsx
@@ -1,7 +1,6 @@
 import React, { FormEvent } from 'react'
 import classNames from 'classnames'
-import isFinite from 'lodash/isFinite'
-import isNil from 'lodash/isNil'
+import { isFinite, isNil, without } from 'lodash'
 import { v4 as uuidv4 } from 'uuid'
 
 import { EndAdornment } from './EndAdornment'
@@ -12,7 +11,7 @@ import { useFocus } from './useFocus'
 import { useValue } from './useValue'
 import { UNIT_POSITION } from './constants'
 
-type UnitPosition = typeof UNIT_POSITION.AFTER
+type UnitPosition = typeof UNIT_POSITION.AFTER | typeof UNIT_POSITION.BEFORE
 
 export interface NumberInputProps {
   autoFocus?: boolean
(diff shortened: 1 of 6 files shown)


packages/react-component-library/src/components/NumberInput/NumberInput.tsx
```diff
@@ -2,50 +2,84 @@ import React, { FormEvent } from 'react'
 import classNames from 'classnames'
 import isFinite from 'lodash/isFinite'
 import isNil from 'lodash/isNil'
+import without from 'lodash/without'
```
m7kvqbe1: I think you can safely do `{ isFinite, isNil, without } from 'lodash'` nowadays because they generate ES modules that can be tree shaken.

packages/react-component-library/src/components/NumberInput/NumberInput.tsx
```diff
@@ -2,50 +2,84 @@ import React, { FormEvent } from 'react'
...
+
+  return displayValue
+}
+
+function getNewValue(event: React.FormEvent<HTMLInputElement>, unit: string) {
+  const { value } = event.currentTarget
+  const valueParts = value.split(' ')
+  const valueWithoutUnit = without(valueParts, unit).join()
+
+  return parseInt(valueWithoutUnit, 10)
```
m7kvqbe1, line 60: Is there ever a possibility this value is sometimes a stepped float? 

Maybe the component doesn't support that currently?

packages/react-component-library/src/components/NumberInput/constants.ts
```diff
@@ -3,4 +3,9 @@ const END_ADORNMENT_TYPE = {
   INCREASE: 'increase',
 } as const
 
-export { END_ADORNMENT_TYPE }
+const UNIT_POSITION = {
+  AFTER: 'after',
+  BEFORE: 'before',
+} as const
+
+export { END_ADORNMENT_TYPE, UNIT_POSITION }
```
m7kvqbe1, line 11: I don't think these are currently exported by the adjacent `index.ts` file. 

Is that an issue for consumers?

m7kvqbe1 (commented):
LGTM just a few comments / queries

packages/react-component-library/src/components/NumberInput/NumberInput.tsx
thyhjwb6: The component does not support that currently. I will add something to the backlog as it's a more involved change.

m7kvqbe1 (approved):
LGTM

PR #1123, opened 2020-06-16
feat(RangeSlider): Add threshold functionality

### LINKED ISSUE #979: RangeSlider with user defined thresholds
### Description
An exemplar project has identified the need for coloured range selector.

To be clear, the need isn't to have the colour change depending on the value of the slider, it's just to change the colour of the whole range selector.

It was suggested that red, amber and green might be suitable colours, but I would suggest allowing the range slider to adopt any colour from the Design System colour palette would be best.

range-selectors

m7kvqbe1:
## Related issue

Closes #979

## Overview

Add threshold functionality to RangeSlider component.

## Reason

>We're implementing a "threshold" version of this component.
>
>The devs will be able to supply thresholds for the range selector, changing the colour appropriately. When the user then moves the range slider handle, the range slider will change colour based on its position of the handle. The range slider can have up to three colours - action, warning, and danger.
>
>The order of the colours are danger, then warning, and then the default action. danger and warning are both optional, and their limits are custom set via prop values.

## Work carried out

- [x] Implement Threshold feature
- [x] Add example stories to Storybook
- [x] Add automated tests
- [x] Update documentation

### DIFF
--- a/packages/react-component-library/src/components/RangeSlider/ThresholdTrack.tsx
+++ b/packages/react-component-library/src/components/RangeSlider/ThresholdTrack.tsx
@@ -1,0 +1,81 @@
+import React from 'react'
+import classNames from 'classnames'
+import { TrackItem, GetTrackProps } from 'react-compound-slider'
+
+interface ThresholdTrackProps extends TrackItem {
+  getTrackProps: GetTrackProps
+  thresholds?: number[]
+}
+
+interface ChunkProps extends ComponentWithClass {
+  left: number
+  width: number
+  maxWidth: number
+  testId: string
+}
+
+export const ThresholdTrack: React.FC<ThresholdTrackProps> = ({
+  target,
+  getTrackProps,
+  thresholds,
+}) => {
+  const singleThreshold = thresholds?.length === 1
+  const doubleThreshold = thresholds?.length === 2
+
+  const Chunk: React.FC<ChunkProps> = ({
+    className,
+    left,
+    width,
+    maxWidth,
+    testId,
+  }) => {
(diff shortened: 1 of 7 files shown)


packages/react-component-library/src/components/RangeSlider/ThresholdTrack.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+export const ThresholdTrack: React.FC<ThresholdTrackProps> = ({
+  target,
+  getTrackProps,
+  thresholds,
+}) => {
+  const singleThreshold = thresholds?.length === 1
+  const doubleThreshold = thresholds?.length === 2
+
+  const Chunk: React.FC<ChunkProps> = ({
```
thyhjwb6, line 25: It feels like this deviates away from current patterns which comes at the cost of making the `ThresholdTrack` `function` less readable.

thyhjwb6 (approved):
Looks good, small observation

packages/react-component-library/src/components/RangeSlider/ThresholdTrack.tsx
m7kvqbe1: Hmm yeah you think move it out into a module of its own?

m7kvqbe1: I remember now there was actually a reason for me including it in here - if I move it out into a sub-component defined outside of this block there is a re-render bug as you move the slider left and right. I'll make a gif and show you. It only happens when you move the slider really quickly.

PR #1142, opened 2020-06-18
ci: Streamline commit checks

### LINKED ISSUE #1127: Consolidate commit linting and checking fixups in CI pipeline
There is the opportunity to save build time when checking the commit messages as part of the build process.

There are jobs `lint_commits` and `check_fixup_commits`. We could have a single job `check_commits` which runs both scripts.

- [ ] Add job that runs both scripts
- [ ] Boast about performance gain

elglup:
## Related issue

Closes #1127 

## Overview
Consolidate commit lint jobs into one

## Reason

Streamline workflows

## Work carried out

Edited Circle config

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,18 +33,13 @@ jobs:
     steps:
       - checkout
       - run: yarn audit
-  lint_commits:
+  check_commits:
     docker: *docker
     steps:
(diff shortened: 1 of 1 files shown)


m7kvqbe1 (approved):
LGTM

PR #1152, opened 2020-06-24
Replace use of grid with flexbox in `DatePicker`

### LINKED ISSUE #1001: Progressively enhance DatePicker to CSS Grid from Flexbox
Chrome 38 does not support CSS grid - use flex box and then enhance to grid.

thyhjwb6:
## Related issue
Fixes #1001 

## Overview
Replaces use of CSS grid with flex in order to support legacy browsers.

## Reason
Legacy browsers need to be supported.

## Work carried out
- [x] Rework to `flex`

### DIFF
--- a/packages/css-framework/src/components/_date-picker.scss
+++ b/packages/css-framework/src/components/_date-picker.scss
@@ -154,15 +154,18 @@
 }
 
 .rn-date-picker__grid {
-  display: grid;
-  grid-template-columns: repeat(1, 300px);
-  grid-gap: 0 f.spacing("8");
-  width: auto;
+  width: 280px;
+  display: flex;
   padding: f.spacing("8") f.spacing("5") f.spacing("5") f.spacing("5");
 }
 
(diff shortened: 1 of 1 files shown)


packages/css-framework/src/components/_date-picker.scss
```diff
@@ -231,15 +232,17 @@
...
 
 .rn-date-picker__day-grid {
-  display: grid;
-  grid-template-columns: repeat(7, 1fr);
-  justify-content: center;
+  display: flex;
+  flex-wrap: wrap;
+
+  * {
+    width: calc(100% / 7) %;
```
m7kvqbe1: I get a SASS error from this - it seems to evaluate it as an expression I think it interprets the `%` as modulus operator.

Maybe we could be more explicit about selector instead of using the wildcard (*)?

m7kvqbe1 (changes requested):

m7kvqbe1 (approved):
LGTM

PR #1165, opened 2020-06-29
feat(Masthead): Add ability to have no service logo

### LINKED ISSUE #1106: Add ability for Masthead to have no logo
### Description

At the moment [the Masthead component](https://docs.royalnavy.io/components/masthead) has an optional Logo prop.

However, if no logo is specified, a default logo is used.

It would be good if there was a way to have no logo for services that don't have one. (Also, when there is no logo, the left margin on the service name may need to be reduced or removed.)

(If I've missed something, please let me know.)

jpveooys:
## Related issue

Closes #1106 

## Overview

This adds a new `hasDefaultLogo` prop to `Masthead` to control whether a default logo is used when no logo is specified. By setting this to `false` and not passing a value for `Logo`, there's now a way to show no logo in the `Masthead`.

## Reason

Not all services have a logo. Before, you could chose between a default logo or a custom logo, but there was no way to display no logo which some services will want to do.

## Work carried out

- [x] Add new `hasDefaultLogo` prop to Masthead
- [x] Update unit tests
- [x] Update docs

### No logo

image

### With logo (and some other bits)

image

(This one should be as it was before.)

## Developer notes

- A new prop was used to avoid introducing a breaking change.
- The CSS changes are because the padding between the logo and service name needs to be removed when a logo isn't shown.
- There was a problem with the props table in the docs where some of the descriptions were in the wrong column. I fixed this by following what was done in the docs for other components and setting Default for all rows.

### DIFF
--- a/packages/css-framework/src/fragments/_masthead.scss
+++ b/packages/css-framework/src/fragments/_masthead.scss
@@ -34,7 +34,7 @@ $masthead-main-height: 58px;
     box-shadow: 0 0 0 0.2rem rgba(f.color("action", "700") , 0.5);
   }
 
-  .rn-masthead__title {
+  &.rn-masthead__service--has-logo .rn-masthead__title {
     margin-left: f.spacing("4");
   }
 }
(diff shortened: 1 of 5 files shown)


m7kvqbe1 (approved):
LGTM

thyhjwb6 (approved):
👍

PR #1184, opened 2020-07-03
Fix Timeline keys

### LINKED ISSUE #1153: Timeline duplicates elements with custom layout in production build
### Description

We've started to use the Timeline component following the https://github.com/m7kvqbe1/fp-grey-box example of a custom layout and run into some curious behaviour when running a production build. (The example was found from the docs: https://docs.royalnavy.io/timeline#advanced-custom-layouts)

This behaviour can be seen with https://github.com/m7kvqbe1/fp-grey-box so I've used that here.

### Steps to reproduce

1. Clone `https://github.com/m7kvqbe1/fp-grey-box.git` and install dependencies
2. (Optional) upgrade the Design System
3. Run `yarn start` to see how the dev build renders
4. Run `yarn build && npx serve -s build` to see how the production build renders

#### Dev build

dev

#### Production build

timeline-prod

thyhjwb6:
## Related issue
Fixes #1153 

## Overview
#1153 was raised because there was duplicate rows when building for production. Investigation found that there was keys missing.

## Work carried out
- [x] Tweaks to `Timeline`
- [x] Fix `key` issue

## Screenshot
View #1153 for a screenshot of the original issue.
### Example of error in Storybook
Screenshot 2020-07-03 at 09 24 10

## Developer notes
`TimelineSide` is still using `index` to create a `key`. More work will be done on `TimelineSide` with the anticipation that we can fully remove the use of `index` for `key` generation.

### DIFF
--- a/packages/react-component-library/src/components/Timeline/Timeline.tsx
+++ b/packages/react-component-library/src/components/Timeline/Timeline.tsx
@@ -27,6 +27,7 @@ import {
 
 import { TimelineOptions } from './context/types'
 import { DEFAULTS } from './constants'
+import { getKey } from '../../helpers'
 
 type timelineRootChildrenType = React.ReactElement<TimelineSideProps>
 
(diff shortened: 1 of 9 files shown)


m7kvqbe1 (approved):
LGTM

m7kvqbe1:
Does this also happen to resolve this ticket (https://github.com/Royal-Navy/design-system/issues/1145)? 

We're planning to tackle this sidebar business separately?

PR #1185, opened 2020-07-03
feat(Timeline): Add ability to bound by two dates

### LINKED ISSUE #1160: Ability to bound a timeline by two dates
### Description

In our exemplar project, there's a requirement to bound the timeline by two dates, and be able to scroll horizontally within that date range.

At the moment the Timeline component supports `startDate` and `range` arguments. These (as I understand) let you control the start and end month, but don't let you specify the bounds at a day level.

m7kvqbe1:
## Related issue

Closes #1160

## Overview

Add the ability to bound a Timeline by two explicit dates.

## Reason

>In our exemplar project, there's a requirement to bound the timeline by two dates, and be able to scroll horizontally within that date range.
>
>At the moment the Timeline component supports startDate and range arguments. These (as I understand) let you control the start and end month, but don't let you specify the bounds at a day level.

## Work carried out

- [x] Add ability to bound a Timeline by two dates

## Screenshot

Screenshot 2020-07-06 at 10 41 28

## Developer Notes

I could add a test that checks for the exact values of the dynamic styles but felt like the storybook story does a better job of covering this?

### DIFF
--- a/packages/react-component-library/src/components/Timeline/Timeline.tsx
+++ b/packages/react-component-library/src/components/Timeline/Timeline.tsx
@@ -1,6 +1,7 @@
 import React from 'react'
+import { differenceInDays } from 'date-fns'
 
-import { TimelineProvider } from './context'
+import { TimelineProvider, TimelineContext } from './context'
 
 import {
   TimelineRootComponent,
@@ -28,6 +29,7 @@ import {
 import { TimelineOptions } from './context/types'
 import { DEFAULTS } from './constants'
 import { getKey } from '../../helpers'
+import { formatPx } from './helpers'
 
 type timelineRootChildrenType = React.ReactElement<TimelineSideProps>
 
@@ -48,6 +50,7 @@ export interface TimelineProps extends ComponentWithClass {
   children: timelineChildrenType | timelineChildrenType[]
   dayWidth?: number
   startDate?: Date
+  endDate?: Date
   today?: Date
   range?: number
 }
@@ -96,6 +99,7 @@ export const Timeline: React.FC<TimelineProps> = ({
   children,
   dayWidth = DEFAULTS.DAY_WIDTH,
   startDate,
+  endDate,
   today,
   range,
 }) => {
@@ -130,14 +134,38 @@ export const Timeline: React.FC<TimelineProps> = ({
   })
 
   return (
-    <TimelineProvider startDate={startDate} today={today} options={options}>
-      <article className="timeline">
-        {rootChildren}
-        <div className="timeline__inner">
-          <header className="timeline__header">{headChildren}</header>
-          {bodyChildren}
-        </div>
-      </article>
+    <TimelineProvider
+      startDate={startDate}
+      endDate={endDate}
+      today={today}
+      options={options}
+    >
+      <TimelineContext.Consumer>
+        {({ state: { months } }) => {
+          const offset = endDate
+            ? formatPx(
+                dayWidth,
+                differenceInDays(months[0].startDate, startDate)
+              )
+            : null
+
+          return (
+            <article className="timeline">
+              {rootChildren}
+              <div
+                className="timeline__inner"
+                style={{
+                  marginLeft: offset,
+                }}
+                data-testid="timeline-inner"
+              >
+                <header className="timeline__header">{headChildren}</header>
+                {bodyChildren}
+              </div>
+            </article>
+          )
+        }}
+      </TimelineContext.Consumer>
     </TimelineProvider>
   )
 }
--- a/packages/react-component-library/src/components/Timeline/TimelineRows.tsx
+++ b/packages/react-component-library/src/components/Timeline/TimelineRows.tsx
@@ -81,39 +81,65 @@ export const TimelineRows: React.FC<TimelineRowsProps> = ({
   return (
     <main className={mainClasses}>
       {hasChildren && (
-        <div className={rowClasses} data-testid="timeline-columns">
-          <TimelineContext.Consumer>
-            {({
-              state: {
-                months,
-                weeks,
-                options: { dayWidth },
-              },
-            }) => {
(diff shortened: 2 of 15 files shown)


packages/react-component-library/src/components/Timeline/TimelineRows.tsx
```diff
@@ -81,39 +81,65 @@ export const TimelineRows: React.FC<TimelineRowsProps> = ({
...
+        <TimelineContext.Consumer>
+          {({
+            state: {
+              months,
+              weeks,
+              endDate: timelineEndDate,
+              options: { dayWidth },
+            },
+          }) => {
+            const wrapperStyles = timelineEndDate
```
thyhjwb6, line 93: Rather than this ternary operator here I would find it easier to read if there was a function:

```javascript
  function getOffset(firstMonthStartDate: Date) {
    if (endDate) {
      return formatPx(
        dayWidth,
        differenceInDays(firstMonthStartDate, startDate)
      )
    }

    return null
  }
```

And then `marginLeft: getOffset(months[0].startDate)`

The result is a lot less indentation.

packages/react-component-library/src/components/Timeline/TimelineWeeks.tsx
```diff
@@ -68,44 +68,66 @@ function renderDefault(
...
+    <TimelineContext.Consumer>
+      {({
+        state: {
+          months,
+          weeks,
+          endDate: timelineEndDate,
+          options: { dayWidth },
+        },
+      }) => {
+        const wrapperStyles = timelineEndDate
```
thyhjwb6, line 80: Based on previous feedback about ternary operators, a similar pattern could be applied here to make it easier to read.

packages/react-component-library/src/components/Timeline/context/reducer.ts
```diff
@@ -100,26 +96,40 @@ export function getDays(
...
       }
     case TIMELINE_ACTIONS.GET_PREV:
       return {
         ...state,
-        ...getMonths(addMonths(startDate, -rangeInMonths), rangeInMonths),
-        weeks: getWeeks(addMonths(startDate, -rangeInMonths), rangeInMonths),
-        days: getDays(addMonths(startDate, -rangeInMonths), rangeInMonths)
+        startDate: subMonths(startDate, range),
+        endDate: null,
+        months: getMonths(subMonths(startDate, range), range),
```
thyhjwb6, line 130: Ha, 👍

packages/react-component-library/src/components/Timeline/Timeline.tsx
```diff
@@ -130,14 +134,37 @@ export const Timeline: React.FC<TimelineProps> = ({
...
-      </article>
+    <TimelineProvider
+      startDate={startDate}
+      endDate={endDate}
+      today={today}
+      options={options}
+    >
+      <TimelineContext.Consumer>
+        {({ state: { months } }) => {
+          const offset = endDate
```
thyhjwb6, line 145: If I change this line to `const offset = null` then all tests still pass. I see your note about adding a test but it doesn't feel right when coverage is not compromised because of a ternary operator.

packages/react-component-library/src/components/Timeline/TimelineRows.tsx
```diff
@@ -12,7 +12,7 @@ export interface TimelineRowsProps extends ComponentWithClass {
   children:
     | React.ReactElement<TimelineRowProps>
     | React.ReactElement<TimelineRowProps>[]
-  renderColumns?: (
+  render?: (
```
thyhjwb6: 👮 

This is a breaking change..?

thyhjwb6 (changes requested):
Looks good although some of the code is getting even more complex. It feels like it can be simplified with some moving around.

packages/react-component-library/src/components/Timeline/TimelineRows.tsx
m7kvqbe1: Oops yes i'll revert that.

packages/react-component-library/src/components/Timeline/Timeline.tsx
m7kvqbe1: I'll add a test that checks for the dynamic styles.

packages/react-component-library/src/components/Timeline/TimelineRows.tsx
m7kvqbe1: The function would actually have more nesting than the ternary as there is another style that needs including:

```typescript
function getOffsetStyles(firstMonthStartDate: Date, timelineEndDate: Date) {
  if (timelineEndDate) {
    return {
      width: formatPx(
        dayWidth,
        differenceInDays(timelineEndDate, firstMonthStartDate) + 1
      ),
      overflow: 'hidden',                  
    }
  }

  return null
}
```

vs

```typescript
const wrapperStyles = timelineEndDate
  ? {
      width: formatPx(
        dayWidth,
        differenceInDays(timelineEndDate, months[0].startDate) + 1
      ),
      overflow: 'hidden',
    }
  : null
```

thyhjwb6: Apologies, I was supposed to add this comment to `Timeline.tsx`

m7kvqbe1: We have agreed to leave the ternaries for the dynamic styles for now and instead push ahead with the upcoming styled-components works for Timeline.

packages/react-component-library/src/components/Timeline/TimelineRows.tsx
m7kvqbe1: I've created a ticket for this breaking change and labelled it for v3: https://github.com/Royal-Navy/design-system/issues/1199

thyhjwb6 (approved):
👍

PR #1189, opened 2020-07-06
fix(Badge): Add ability to drill arbitrary props

### LINKED ISSUE #1121: Badge Component - data-testid attribute ignored
### Steps to reproduce
`<Badge data-testid="badge-testid">1</Badge>
`

m7kvqbe1:
## Related issue

Closes #1121

## Overview

Make Badge apply additional arbitrary props to component wrapper element.

## Reason

> Downstream consumers want to be able to apply their own test IDs to the components.

## Work carried out

- [x] Add ability to drill additional arbitrary props
- [x] Add automated test

### DIFF
--- a/packages/react-component-library/src/components/Badge/Badge.tsx
+++ b/packages/react-component-library/src/components/Badge/Badge.tsx
@@ -38,6 +38,7 @@ export const Badge: React.FC<BadgeProps> = ({
   colorVariant = BADGE_COLOR_VARIANT.SOLID,
   size = BADGE_SIZE.REGULAR,
   variant,
+  ...rest
 }) => {
   const classes = classNames(
     'rn-badge',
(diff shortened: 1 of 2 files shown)


packages/react-component-library/src/components/Badge/Badge.tsx
```diff
@@ -49,7 +50,7 @@ export const Badge: React.FC<BadgeProps> = ({
   )
 
   return (
-    <span className={classes} data-testid="badge">
+    <span className={classes} {...rest} data-testid="badge">
```
thyhjwb6: Should `{...rest}` be last so consumers can potentially override `data-testid` too?

thyhjwb6 (commented):
A small comment about overriding `data-testid` because doing it in this order means it definitely cannot be.

packages/react-component-library/src/components/Badge/Badge.tsx
m7kvqbe1: I'll move it around but I don't think it would make a difference because Babel strips it.

thyhjwb6: Yes! I just remembered that too..

thyhjwb6 (approved):
👍

PR #1190, opened 2020-07-06
fix(Timeline): Custom components not composable at top level

### LINKED ISSUE #1161: Custom Timeline components don't render outside rows
### Steps to reproduce

1. Try to render `ExampleTimeline1` from below:

```tsx
import React from 'react'

import {
  Timeline,
  TimelineDays,
  TimelineEvent,
  TimelineEvents,
  TimelineMonths,
  TimelineRow,
  TimelineRows,
  TimelineTodayMarker,
  useTimelinePosition,
} from '@royalnavy/react-component-library'


const CustomTimelineComponent1 = () => {
  return (
    <div id="asdf1">
      <button>Custom component 1</button>
    </div>
  )
}

const CustomTimelineComponent2 = ({
  startDate,
  endDate,
}: {
  startDate: Date
  endDate: Date
}) => {
  const { width, offset } = useTimelinePosition(startDate, endDate)

  return (
    <div
      style={{
        position: 'absolute',
        display: 'inline-block',
        width,
        left: offset,
        zIndex: 1,
      }}
      id="asdf2"
    >
      Custom component 2
    </div>
  )
}

const ExampleTimeline1 = () => {
  const names = ['row 1', 'row 2']

  return (
    <Timeline
      startDate={new Date(2020, 4, 10)}
      today={new Date(2020, 4, 15)}
      range={3}
      dayWidth={30}
    >
      <TimelineTodayMarker />
      <TimelineMonths />
      <TimelineDays />
      <CustomTimelineComponent1 />
      <CustomTimelineComponent2
        startDate={new Date(2020, 4, 20)}
        endDate={new Date(2020, 4, 21)}
      />
      <TimelineRows>
        {names.map((name) => (
          <TimelineRow name={name} key={name}>
            <TimelineEvents>
              <TimelineEvent
                startDate={new Date(2020, 4, 14)}
                endDate={new Date(2020, 4, 18)}
              >
                Event 1
              </TimelineEvent>
            </TimelineEvents>
          </TimelineRow>
        ))}
      </TimelineRows>
    </Timeline>
  )
}
```

2. Run a dev build of the app

3. Notice that neither custom component appears in the DOM (e.g. searching for `asdf` in the browser inspector yields no results).

If the custom components are moved within a `TimelineRow`, they do render. Also, similar to #1153, there seem to be some differences in behaviour between dev and prod builds.  (In a prod build, the first custom component does render for me, but still not the second one.)

m7kvqbe1:
## Related issue

Closes #1161

## Overview

Give the end user the ability to add custom Timeline components at the root of their composition.

## Reason

>Custom Timeline components don't render outside rows.

## Work carried out

- [x] Fix extraction of root Timeline children
- [x] Refactor `extractRowData` method
- [x] Add automated test

### DIFF
--- a/packages/react-component-library/src/components/Timeline/Timeline.tsx
+++ b/packages/react-component-library/src/components/Timeline/Timeline.tsx
@@ -2,13 +2,7 @@ import React from 'react'
 import { differenceInDays } from 'date-fns'
 
 import { TimelineProvider, TimelineContext } from './context'
-
-import {
-  TimelineRootComponent,
-  TimelineHeadComponent,
-  TimelineBodyComponent,
-  TimelineComponent,
-} from './types'
+import { TimelineComponent } from './types'
 
 import {
   TimelineSide,
@@ -64,11 +58,14 @@ function isComponentOf<T extends TimelineComponent>(
   return names.includes(typeName)
 }
 
-function extractChildrenOf<T>(
+function extractChildren(
   children: timelineChildrenType | timelineChildrenType[],
-  names: string[]
+  names: string[],
+  inverse?: boolean
 ) {
-  return (children as []).filter((child) => isComponentOf(child, names))
+  return (children as []).filter((child) => {
+    return inverse ? !isComponentOf(child, names) : isComponentOf(child, names)
+  })
 }
 
 function extractRowData(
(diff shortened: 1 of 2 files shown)


packages/react-component-library/src/components/Timeline/Timeline.tsx
```diff
@@ -73,16 +73,13 @@ function extractRowData(
...
-        rows = items.map(
+      const rows =
+        React.Children.map(
+          children,
           ({ props: { name } }: React.ReactElement<TimelineRowProps>) => ({
             name,
           })
-        )
-      }
+        ) || []
```
thyhjwb6, line 82: If I remove `|| []` then I cannot see anything breaking.

packages/react-component-library/src/components/Timeline/Timeline.tsx
```diff
@@ -63,9 +63,12 @@ function isComponentOf<T extends TimelineComponent>(
 
 function extractChildrenOf<T>(
   children: timelineChildrenType | timelineChildrenType[],
-  names: string[]
+  names: string[],
+  inverse?: boolean
```
thyhjwb6, line 64: If this parameter is staying then let's change the function to `extractChildren`

thyhjwb6 (changes requested):
Looks good, a couple of comments.

packages/react-component-library/src/components/Timeline/Timeline.tsx
m7kvqbe1: Not short circuiting to the empty array breaks some of the storybook stories that include TimelineSide but no rows. It's expecting a nested array that it can map over.

Screenshot 2020-07-07 at 13 58 09

m7kvqbe1: https://github.com/Royal-Navy/design-system/blob/e572fbece322bed8c7f4483a0956a98054078d01/packages/react-component-library/src/components/Timeline/TimelineSide.tsx#L35

thyhjwb6: Okay, the overarching point is that unit tests do not fail either. I think if we were testing `TimelineSide` from the outside `Timeline` then there would not be this gap.

m7kvqbe1: I've added some additional tests to cover TimelineSide.

thyhjwb6 (approved):
👍

PR #1200, opened 2020-07-07
Add `List` component

### LINKED ISSUE #976: Quick-access tiles (list variation)
### Description
For an exemplar project, we need a new variation of a list component.

In this particular scenario the list is used to link to items (think recenty-opened-files).

This component has a heading, description and button. The entire list item should be clickable.

Please ensure the colour contrast matches accessibility guidelines - the example below has not been checked.

quick-access-tiles

thyhjwb6:
## Related issue
Closes #976 

## Overview
Adds a `List` component.

## Reason
Required for exemplar project.

## Work carried out
- [x] Add component
- [x] Add docs
- [x] Warn if overwriting props that are used for rendering

## Screenshot
Screenshot 2020-07-07 at 15 28 59


packages/css-framework/src/components/_list.scss
```diff
@@ -0,0 +1,63 @@
+@use "../abstracts/functions" as f;
+
+.rn-list {
+  border: f.spacing("px") solid f.color("neutral", "100");
```
m7kvqbe1, line 4: I'm not familiar with this f.spacing `px` usage - what does this do?

packages/docs-site/src/library/pages/components/list.md
```diff
@@ -0,0 +1,113 @@
+---
+title: List
+description: todo
+header: true
+---
```
m7kvqbe1, line 6: Worth adding `draft: true` to the frontmatter if the design docs are due to come at a later date.

packages/react-component-library/src/components/List/ListItem.tsx
```diff
@@ -0,0 +1,54 @@
...
+  return (
+    <li className={classes} data-testid="list-item">
+      <button
+        className="rn-list__item-content"
+        onClick={onClick}
+        onMouseEnter={onMouseEnter}
+        onMouseLeave={onMouseLeave}
+      >
+        <div>
+          <h2 className="rn-list__item-title">{title}</h2>
```
m7kvqbe1, line 41: Is `h2` the right tag? Maybe not a big deal - open to debate I think.

m7kvqbe1 (changes requested):
LGTM - just a few tiny comments

packages/react-component-library/src/components/List/ListItem.tsx
thyhjwb6: Having spoken with @hxltrhuxze we feel `h2` is fine at this stage.

packages/css-framework/src/components/_list.scss
thyhjwb6: It is for `1px`:

> as everything else is set in rems having a single pixel value is often quite useful

m7kvqbe1 (approved):

PR #1220, opened 2020-07-13
fix(Types): Rename and export ComponentWithClass

jpveooys:
## Related issue

n/a

## Overview

This renames `ComponentWithClass` to `PropsWithClassName` and starts exporting it.

## Reason

Previously the `ComponentWithClass` interface was not exported. This meant that it wasn't included in the distributed types (in `packages/react-component-library/dist/types`) and users of the component library could not set the `className` prop (for supported components e.g. `Modal`).

Additionally, the name of `ComponentWithClass` was misleading as it's used for props rather than components.

For example, the following no longer returns a type error when using the distributed version of the component library:

```
      <List className="test" >
        <ListItem title="Title" onClick={() => {}}>
          Description
        </ListItem>
      </List>
```

An example of the error that you would've previously received:
```
TypeScript error in ...:
Type '{ children: Element; className: string; }' is not assignable to type 'IntrinsicAttributes & ListProps & { children?: ReactNode; }'.
  Property 'className' does not exist on type 'IntrinsicAttributes & ListProps & { children?: ReactNode; }'.  TS2322
```

## Work carried out

- [x] Rename `ComponentWithClass` to `PropsWithClassName`
- [x] Remove incorrect inheritance from `React.ComponentType`
- [x] Export `PropsWithClassName`
- [x] Import `PropsWithClassName` where needed

## Screenshot

n/a

## Developer notes

Note there seems to be a similar problem for `LinkTypes` and possibly others. Those aren't fixed here as we also need to be careful to not introduce a breaking change by adding non-optional props. (A small workaround for `LinkTypes` was added to avoid needing to export it.)

### DIFF
--- a/packages/react-component-library/src/components/Badge/Badge.tsx
+++ b/packages/react-component-library/src/components/Badge/Badge.tsx
@@ -1,6 +1,7 @@
 import React from 'react'
 import classNames from 'classnames'
 
+import { PropsWithClassName } from '../../types/PropsWithClassName'
 import {
   BADGE_COLOR,
   BADGE_COLOR_VARIANT,
@@ -22,7 +23,7 @@ export type BadgeSizeType =
   | typeof BADGE_SIZE.LARGE
   | typeof BADGE_SIZE.XLARGE
 
-interface BadgeProps extends ComponentWithClass {
+interface BadgeProps extends PropsWithClassName {
   color?: BadgeColorType
   colorVariant?:
     | typeof BADGE_COLOR_VARIANT.FADED
--- a/packages/react-component-library/src/components/Button/Button.tsx
+++ b/packages/react-component-library/src/components/Button/Button.tsx
@@ -7,6 +7,7 @@ import {
   BUTTON_VARIANT,
   BUTTON_ICON_POSITION,
 } from './constants'
+import { PropsWithClassName } from '../../types/PropsWithClassName'
 
 export type ButtonSizeType =
   | typeof BUTTON_SIZE.SMALL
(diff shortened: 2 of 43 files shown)


packages/react-component-library/src/components/Badge/Badge.tsx
```diff
@@ -1,6 +1,7 @@
 import React from 'react'
 import classNames from 'classnames'
 
+import { PropsWithClassName } from '../../types/PropsWithClassName'
```
m7kvqbe1, line 4: I think if you name this file `PropsWithClassName.d.ts` you can get away without the explicit import at the top of each file. The `d` in the filename makes it a typescript 'ambient' module.

m7kvqbe1 (changes requested):
Makes sense to me - just one small comment from me about global types.

packages/react-component-library/src/components/Badge/Badge.tsx
jpveooys: It appears to work but only because I added `PropsWithClassName` to `Link.d.ts` :face_with_head_bandage: As soon as the imports are removed, `PropsWithClassName` is omitted from the bundled types again.

I was under the impression that declaration files are mainly for defining types for e.g. a third-party JS library without native types.

Definitely ought to remove/rename the `PropsWithClassName` this adds to `Link.d.ts` so that can't be done by accident though...

jpveooys:
Just rebased to resolve conflicts.

On reflection, I don't think `PropsWithClassName` is a good name either as it's not used in the same way as `React.PropsWithChildren` (it should probably be something like `ClassNameProp`).

Another option is to get rid of `PropsWithClassName` completely and just add `className` to the relevant prop types – probably will be the most concise thing...

jpveooys:
This is now stale so I'm closing this in favour of an issue.

PR #1226, opened 2020-07-13
fix(List): Correct type of ListItem children

jpveooys:
## Related issue

n/a

## Overview

This changes the type of the `children` prop of `ListItem` from `string` to `string | string[]` to allow embedded expressions to be used within a `ListItem`.

## Reason

When using an embedded expression within a `ListItem`, the type of `children` ends up being `string[]` instead of `string`.

For example:

```jsx
const text = 'Ha'

...

<List>
  <ListItem title="Title" onClick={() => {}}>
    Description {text}
  </ListItem>
</List>
```

was resulting in the following error:

```
TypeScript error in ...:
Type '{ children: string[]; title: string; onClick: () => void; }' is not assignable to type 'ListItemProps'.
  Types of property 'children' are incompatible.
    Type 'string[]' is not assignable to type 'string'.  TS2322
```

## Work carried out

- [x] Change type of `ListItemProps.children`
- [x] Add unit test for this case
- [x] Update docs

### DIFF
--- a/packages/react-component-library/src/components/List/List.test.tsx
+++ b/packages/react-component-library/src/components/List/List.test.tsx
@@ -207,4 +207,23 @@ describe('List', () => {
       )
     })
   })
+
+  describe('when a list item has multiple children ', () => {
+    beforeEach(() => {
+      const extraItemText = 'extra text'
+      wrapper = render(
+        <List>
+          <ListItem onClick={jest.fn()} title="List item">
+            Main text: {extraItemText}
+          </ListItem>
+        </List>
+      )
+    })
+
+    it("renders the item's text", () => {
+      expect(wrapper.getAllByTestId('list-item')[0]).toHaveTextContent(
+        'Main text: extra text'
+      )
+    })
+  })
 })
--- a/packages/docs-site/src/library/pages/components/list.md
+++ b/packages/docs-site/src/library/pages/components/list.md
@@ -82,7 +82,7 @@ todo
 <DataTable data={[
   {
     Name: 'children',
-    Type: 'string',
+    Type: 'string | string[]',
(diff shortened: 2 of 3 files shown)


packages/react-component-library/src/components/List/List.test.tsx
```diff
@@ -207,4 +207,23 @@ describe('List', () => {
...
+      wrapper = render(
+        <List>
+          <ListItem onClick={jest.fn()} title="List item">
+            Main text: {extraItemText}
+          </ListItem>
+        </List>
+      )
+    })
+
+    it("renders the item's text", () => {
```
m7kvqbe1, line 223: These should be single quotes.

```suggestion
    it('renders the item's text', () => {
```

m7kvqbe1 (approved):
LGTM

packages/react-component-library/src/components/List/List.test.tsx
m7kvqbe1, line 223: These should be single quotes.

```suggestion
    it('renders the item's text', () => {
```

m7kvqbe1 (changes requested):

packages/react-component-library/src/components/List/List.test.tsx
jpveooys: They're double due to the embedded single quote – unless you're suggesting escaping instead?

m7kvqbe1: Ah sorry I didn't spot that - i'd probably escape it but doesn't really matter either way.

I've approved it 👍

m7kvqbe1 (approved):

PR #1227, opened 2020-07-13
feat(NumberInput): Add aria attributes

### LINKED ISSUE #1067: Add aria attributes to NumberInput component
### Description

Component | Sub element | Attribute | Value | Notes
-- | -- | -- | -- | --
NumberInput |   | input | aria-valuemax |  
  |   | input | aria-valuemin |  
  |   | input | aria-valuenow |

m7kvqbe1:
## Related issue

Closes #1067

## Overview

Add aria attributes to NumberInput component.

## Reason

> Make all existing components accessible.

## Work carried out

- [x] Add aria attributes
- [x] Add automated tests

### DIFF
--- a/packages/react-component-library/src/components/NumberInput/NumberInput.test.tsx
+++ b/packages/react-component-library/src/components/NumberInput/NumberInput.test.tsx
@@ -10,11 +10,20 @@ describe('NumberInput', () => {
   let onChangeSpy: (event: any) => void
 
   function assertInputValue(expected: string) {
-    it('should set the new value', () => {
+    it('should set the new value attribute', () => {
       const input = wrapper.getByTestId(
         'number-input-input'
       ) as HTMLInputElement
       expect(input.value).toEqual(expected)
+    })
+  }
+
+  function assertAriaValue(expected: number) {
+    it('should set the new `aria-valuenow` attribute', () => {
+      const container = wrapper.getByTestId(
+        'number-input-container'
+      ) as HTMLDivElement
+      expect(container).toHaveAttribute('aria-valuenow', `${expected}`)
     })
   }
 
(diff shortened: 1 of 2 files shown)


packages/react-component-library/src/components/NumberInput/NumberInput.test.tsx
```diff
@@ -10,14 +10,23 @@ describe('NumberInput', () => {
...
-    it('should set the new value', () => {
+    it('should set the new value attribute', () => {
       const input = wrapper.getByTestId(
         'number-input-input'
       ) as HTMLInputElement
       expect(input.value).toEqual(expected)
     })
   }
 
+  function assertAriaValue(expected: number) {
```
thyhjwb6, line 21: I like what you have done here. I am wondering if the assertion in here should just be part of `assertInputValue`. Then it is done every time we are asserting the value. Do you have thoughts? Either that or we `assertAriaValue` for every `assertInputValue` as it could become confusing when maintaining these tests. When to use `assertInputValue` and when to use `assertAriaValue`.

packages/react-component-library/src/components/NumberInput/NumberInput.tsx
```diff
@@ -108,7 +108,15 @@ export const NumberInput: React.FC<NumberInputProps> = ({
   }
 
   return (
-    <div className={classes} data-testid="number-input-container">
+    <div
+      className={classes}
+      data-testid="number-input-container"
+      role="spinbutton"
```
thyhjwb6, line 114: 👍

thyhjwb6 (commented):
Looks good.. just a thought about test structure.

packages/react-component-library/src/components/NumberInput/NumberInput.test.tsx
m7kvqbe1: I kept them separate because they are actually asserting different values - the aria value expects a number and the component value expects a string (that sometimes also includes a unit as a prefix/suffix).

thyhjwb6 (approved):
👍

PR #1233, opened 2020-07-15
feat(TabSet): Add accessibility related attributes

### LINKED ISSUE #1076: Add aria attributes to TabSet
### Description

Component | Sub element | Attribute | Value | Notes
-- | -- | -- | -- | --
TabSet | article | role | tabs |  
  | ScrollButton | aria-hidden |   |  
  | ol | role | tablist |  
  | TabItem | role | tab |  
  | TabItem - selected | aria-selected | "true" |  
  | TabItem - not selected | aria-selected | "false" |  
  | TabItem - not selected | tabindex | -1 | Take tab out of tab sequence
  | TabItem | aria-controls | id of tabpanel |  
  | body div | role | tabpanel |  
  | body div | aria-labelledby | id of tab |  
  | body div | tabindex | 0 | Add tab content to tab sequenc

m7kvqbe1:
## Related issue

Closes #1076

## Overview

Add accessibility related attributes to TabSet component.

## Reason

## Work carried out

- [x] Add automated tests

### DIFF
--- a/packages/react-component-library/src/components/TabSet/ScrollButton.tsx
+++ b/packages/react-component-library/src/components/TabSet/ScrollButton.tsx
@@ -30,6 +30,7 @@ export const ScrollButton: React.FC<ScrollButtonProps> = ({
     <button
       className={classes}
       onClick={onClick}
+      aria-hidden
       data-testid={`scroll-${direction}`}
     >
       {SCROLL_DIRECTION_ICON_MAP[direction]}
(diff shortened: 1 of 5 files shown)


packages/react-component-library/src/components/TabSet/ScrollButton.tsx
```diff
@@ -30,6 +30,7 @@ export const ScrollButton: React.FC<ScrollButtonProps> = ({
     <button
       className={classes}
       onClick={onClick}
+      aria-hidden
```
thyhjwb6, line 33: I am wondering if we actually want this hidden or to give it `aria-label`? Are you thinking that the user still has full functionality with a screenreader, without using the scroll buttons?

thyhjwb6 (approved):
👍  just a question about `aria-hidden`

packages/react-component-library/src/components/TabSet/ScrollButton.tsx
m7kvqbe1: Yeah my assumption was that they wouldn't interact with this element anyway because of the screen reader.

Issue #1231, opened 2020-07-14
Timeline bound by start and end date not compatible with custom layout


jpveooys:
### Description

#1160 added the ability to bound a timeline by two dates. Unfortunately, it doesn't seem to be compatible with a custom layout in the style of https://github.com/m7kvqbe1/fp-grey-box. This seems to be due to the way it uses a negative `margin-left` when the start date isn't the first of the month, while the custom layout pushes the timeline to the right.

(Additionally, the name of the first month can be pushed off screen.)

### Steps to reproduce

1. Clone https://github.com/m7kvqbe1/fp-grey-box
2. Update Design System dependencies
3. Apply the following diff

```diff
diff --git a/src/App.tsx b/src/App.tsx
index 5967a82..daa2648 100644
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -85,11 +85,11 @@ const CustomEvent = ({
 function App() {
   return (
     <div>
-      <Timeline startDate={new Date(2020, 4, 0)} today={new Date(2020, 3, 15)}>
+      <Timeline startDate={new Date(2020, 4, 20)} endDate={new Date(2020, 5, 12)} today={new Date(2020, 3, 15)}>
         {/* <TimelineTodayMarker /> */}
         <TimelineMonths render={CustomTimelineMonth} />
         <TimelineDays render={CustomTimelineDay} />
-        <TimelineRows renderColumns={() => null}>
+        <TimelineRows renderColumns={() => <></>}>
           <div className="row-group__wrapper">
             <div className="row-group__key">
               <span className="row-group__name">Tidespring</span>
@@ -154,7 +154,7 @@ function App() {
             </div>
           </div>
         </TimelineRows>
-        <TimelineRows renderColumns={() => null}>
+        <TimelineRows renderColumns={() => <></>}>
           <div className="row-group__wrapper">
             <div className="row-group__key">
               <span className="row-group__name">Queen Elizabeth</span>
@@ -219,7 +219,7 @@ function App() {
             </div>
           </div>
         </TimelineRows>
-        <TimelineRows renderColumns={() => null}>
+        <TimelineRows renderColumns={() => <></>}>
           <div className="row-group__wrapper">
             <div className="row-group__key">
               <span className="row-group__name">Daring</span>
```

The timeline should start on 20 May 2020, but instead earlier dates are visible, and other content is pushed off screen.

### Screenshot

[...]

PR #1237, opened 2020-07-15
fix(Timeline): Improve behaviour when bounding by start and end date

jpveooys:
## Related issue

#1231

## Overview

This changes how the feature to bound a timeline by a start and end date works so that it behaves better and is compatible with custom layouts.

## Reason

- The previous implementation wasn't compatible with custom layouts in the style of [fp-grey-box](https://github.com/m7kvqbe1/fp-grey-box)
- The title of the first month was hidden if the start date wasn't the first of the month, which wasn't desirable
- There were some other artefacts like the 'No data available' message not being centred

The changes also fixed this artefact when not bound by two dates:

## Work carried out

- [x] Refactor how the Timeline state is constructed so it's built from a start and end date rather than a date and a number of months
- [x] Change `days` in the Timeline state when the Timeline is bound by two dates so that it only contains the dates between the two dates
- [x] Change how the Timeline is rendered when it is bound by two dates, so instead of hiding dates outside the date range, it doesn't display them and sizes things appropriately
- [x] Made sure the first month label is always displayed
- [x] Updated tests for when the Timeline is bound by two dates

## Screenshot

This is the example from #1231 after these changes:

image

## Developer notes

- Chromatic picked up one change for the timeline that is an improvement (the 'No data available' message now being positioned correctly when the timeline is bound by two dates)
- The existing values of `dayIndex`, `weekIndex` etc. in the state should be preserved (though I'm not sure how they're intended to be used)
- The reducer is missing some test coverage, this is unchanged in this PR at the moment
- I've had a look through and everything not related to the bounded by two dates case should be the same (apart from that empty week problem mentioned above), but please have a play in Storybook etc. to double-check

### DIFF
--- a/packages/react-component-library/src/components/Timeline/Timeline.tsx
+++ b/packages/react-component-library/src/components/Timeline/Timeline.tsx
@@ -1,7 +1,6 @@
 import React from 'react'
-import { differenceInDays } from 'date-fns'
 
-import { TimelineProvider, TimelineContext } from './context'
+import { TimelineProvider } from './context'
 import { TimelineComponent } from './types'
 
 import {
@@ -23,7 +22,6 @@ import {
 import { TimelineOptions } from './context/types'
 import { DEFAULTS } from './constants'
 import { getKey } from '../../helpers'
-import { formatPx } from './helpers'
 
 type timelineRootChildrenType = React.ReactElement<TimelineSideProps>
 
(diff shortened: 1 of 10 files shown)


packages/react-component-library/src/components/Timeline/Timeline.tsx
```diff
@@ -140,32 +140,13 @@ export const Timeline: React.FC<TimelineProps> = ({
       today={today}
```
m7kvqbe1, line 138: There are some unused imports declared at the top of this file.

packages/react-component-library/src/components/Timeline/context/reducer.ts
```diff
@@ -23,113 +16,87 @@ import {
...
-    .map((_, index) => {
-      const monthIndex = getMonth(date) + index
-
-      const newDate = setMonth(date, monthIndex)
-      const startDate = startOfMonth(newDate)
+import { WEEK_START } from '../constants'
 
+export function getMonths(start: Date, end: Date): TimelineMonth[] {
+  const months = eachMonthOfInterval({ start, end }).map(
+    (date, index) => {
```
m7kvqbe1: I think the `index` param here is unused now.

packages/react-component-library/src/components/Timeline/context/reducer.ts
```diff
@@ -23,113 +16,87 @@ import {
...
-): TimelineDay[] {
-  const months = getMonths(date, rangeInMonths)
-  const year = getYear(date)
-
-  const days = months.flatMap(({ monthIndex }) => {
-    const total = getDaysInMonth(new Date(year, monthIndex))
-    const arr = [...Array(total).keys()].map(i => i + 1) // [1, ..., ~31]
-
-    return arr.map((day, dayIndex) => {
+export function getDays(start: Date, end: Date): TimelineDay[] {
```
m7kvqbe1, line 48: Really nice to see the reduction in complexity in this reducer 🎉

m7kvqbe1 (changes requested):
Looks great! Just a couple of tiny comments.

m7kvqbe1 (approved):
LGTM

PR #1240, opened 2020-07-17
feat(DatePicker): Add accessibility attributes

### LINKED ISSUE #1064: Add aria attributes to DatePicker component
### Description
As a starting point, use the example: 
https://www.w3.org/TR/wai-aria-practices-1.1/examples/dialog-modal/datepicker-dialog.html

m7kvqbe1:
## Related issue

Closes #1064

## Overview

Add accessibility related attributes to DatePicker component.

## Reason

## Work carried out

- [x] Add automated tests
- [x] Minor refactor

## Developer notes

I've created a separate ticket to tackle the keyboard control - there is some additional nuance to this work.


thyhjwb6 (changes requested):
Looks good, some feedback:

- It seems like the first two commits are not atomic as they are not associated with a test. Should the first three commits be squashed?
- The test coverage on this component has gone from 94% to 87% reducing overall coverage of the library.
- It would be nice to address some of the lines that have not been previouslyn covered by tests in this component at some point.

m7kvqbe1:
> * It seems like the first two commits are not atomic as they are not associated with a test. Should the first three commits be squashed?

I've added fixups to include tests for the first two commits which are for changes to the FloatingBox component.

> * The test coverage on this component has gone from 94% to 87% reducing overall coverage of the library.

I removed some tests that were testing a sub-component - i've now added additional tests to check the correct days are rendered.

> * It would be nice to address some of the lines that have not been previouslyn covered by tests in this component at some point.

I can see that there are some default values not covered in `datepickerContext.ts` any idea how get these covered? I'm not entirely convinced these 4 lines warrant coverage. Might end up with coverage of this once Keyboard control is looked it.

thyhjwb6 (approved):
👍

PR #1242, opened 2020-07-17
feat(Popover): Add accessibility attributes

### LINKED ISSUE #1069: Add aria attributes to Popover component
### Description

Component | Sub element | Attribute | Value
-- | -- | -- | --
Popover |   | role | dialog
  |   | aria-describedby | id of content

m7kvqbe1:
## Related issue

Closes #1069 

## Overview

Add accessibility related attributes to Popover and FloatingBox components.

## Reason

## Work carried out

- [x] Add automated tests

### DIFF
--- a/packages/react-component-library/src/components/Popover/Popover.tsx
+++ b/packages/react-component-library/src/components/Popover/Popover.tsx
@@ -14,6 +14,8 @@ import {
   FLOATING_BOX_SCHEME,
 } from '../../primitives/FloatingBox'
 
+import { getId } from '../../helpers'
+
 interface PopoverProps
   extends Omit<FloatingBoxProps, 'onMouseEnter' | 'onMouseLeave'> {
   children: React.ReactElement
@@ -67,6 +69,8 @@ export const Popover: React.FC<PopoverProps> = ({
     )[0]
   }
 
+  const contentId = getId('popover-content')
+
   /**
    * Type error in upstream Tether package. Fix submitted.
    * Using createElement allows us to avoid the type error.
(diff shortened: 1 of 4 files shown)


thyhjwb6 (changes requested):
Looks good although it feels like we need to give the `aria-describedby` more thought.

packages/react-component-library/src/components/Popover/Popover.tsx
```diff
@@ -91,6 +95,8 @@ export const Popover: React.FC<PopoverProps> = ({
             onMouseLeave={handleOnMouseLeave}
             position={PLACEMENTS.ARROW_POSITION}
             scheme={scheme}
+            aria-describedby={contentId}
+            contentId={contentId}
```
thyhjwb6, line 99: My comment seems to have been lost but it was something along the lines of:

> Is this right that the `aria-describedby` is referencing itself? I would have thought it would clone `content` with an ID attribute which is then referenced or if `content` is always a `string` then it is not referenced at all.

m7kvqbe1: It passes contentId into the `FloatingBox` component where it is then applied to a child element, so it's actually referencing a direct descendant.

FloatingBox:

```
FloatingBox
----div aria-describedby="contentId"
--------div id="contentId"
```

thyhjwb6: I misinterpreted this change. Cool 👍

thyhjwb6 (approved):
👍

PR #1249, opened 2020-07-20
fix(Switch): minor error setting Switch option key

antictor:
The Switch component used the wrong variable to generate a unique key so all keys had the name switch-option-undefined


m7kvqbe1 (approved):
👍

PR #1254, opened 2020-07-21
ci(Gha): Add cra to npm smoke test

### LINKED ISSUE #1212: Add create react app to NPM smoke test
There is a new package to add to the publish packages list in NPM smoke test workflow, the CRA template package:

https://www.npmjs.com/package/cra-template-royalnavy

Need to add an additional job that does this to test the consumption of it:

npx create-react-app my-app --template royalnavy

elglup:
## Related issue

Closes #1212 

## Overview

Add create-react-app to NPM smoke test GHA

## Work carried out

Edited npmsmoketest.yml

### DIFF
--- a/.github/workflows/npmsmoketest.yml
+++ b/.github/workflows/npmsmoketest.yml
@@ -1,13 +1,13 @@
 name: NPM Smoke Test
 env:
-  LOCAL_REGISTRY: "http://0.0.0.0:4873"
+  LOCAL_REGISTRY: "http://localhost:4873"
 
 on:
   pull_request:
 
 
 jobs:
-     npm_smoketest:
+     NPM_Smoketest:
       runs-on: ubuntu-latest
 
       steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/npmsmoketest.yml
```diff
@@ -58,22 +62,16 @@ jobs:
...
-            path: '**/node_modules'
-            key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock') }}
-
-
         - name: Publish packages
-          run  : |
+          run: |
             cd main
             yarn install
             npm publish --registry ${{env.LOCAL_REGISTRY}} packages/css-framework
```
m7kvqbe1, line 69: Is it worth setting registry globally so you don't have to explicitly provide the `--registry` argument every time?

```
npm config set registry ${{env.LOCAL_REGISTRY}}
```

m7kvqbe1 (changes requested):
LGTM - I've just left one comment - maybe there is a reason to not do what i've suggested?

.github/workflows/npmsmoketest.yml
elglup: We could although in the next job we need to run yarn install again (not using local registry), so would need to set & unset

m7kvqbe1: Ok fair enough - feel free to just leave it as is then

m7kvqbe1 (approved):

PR #1279, opened 2020-07-28
fix(NumberInput): Call `onChange` callback

### LINKED ISSUE #1256: Number Input only calls onChange when component it blurred
### Steps to reproduce
Create a form with a number input
Add an onChange handler function that gets the value change and uses it to update it's forms state
Display the form state on the screen

When the user changes a number by editing the value, not the buttons, the onChange is not called until the field looses focus.

In previous versions of the toolkit the onChange was called for all changes and appears to have changed here:

https://github.com/Royal-Navy/design-system/commit/99143995d78f52595cc58e06ff989636e7f36f70

thyhjwb6:
## Related issue
Fixes #1256 

## Overview
A previous refactor of `NumberInput` removed the call of the `onChange` callback which was noticeable when a user would type in the field.

## Reason
Missed in a refactor. This is a regression.

## Work carried out
- [x] Add `onChange` call

### DIFF
--- a/packages/react-component-library/package.json
+++ b/packages/react-component-library/package.json
@@ -67,7 +67,7 @@
     "@svgr/webpack": "^5.1.0",
     "@testing-library/jest-dom": "^5.0.2",
     "@testing-library/react": "^10.0.2",
-    "@testing-library/user-event": "^12.0.11",
+    "@testing-library/user-event": "^12.0.15",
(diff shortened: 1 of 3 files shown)


antictor (approved):

m7kvqbe1 (commented):
LGTM

PR #1281, opened 2020-07-29
Refactor `TimelineSide` to support `rowheader` `role` attributes

### LINKED ISSUE #938: Revist TimelineSide - refactor and possibly remove component
Each vertical slice of TimelineSide could instead be part of the associated Month / Weeks / Days / Rows component `defaultRender` - the display of which could be toggled via a root Timeline prop (maybe `withKey`).

### LINKED ISSUE #1268: Add aria attributes to TimelineSide
### Description
Add the ARIA attributes to the `TimelineSide` subcomponent of `Timeline`.

Includes adding `rowheader` attributes.

thyhjwb6:
## Related issue
Closes #1268 
Closes #938

## Overview
This is a refactor of `TimelineSide` to take advantage of using a `rowheader` `role` attribute.

## Reason
`TimelineSide` was one of our costlier pieces of technical debt so this change should clear up several issues.

## Work carried out
- [x] Refactor
- [x] Add attributes

## Developer notes
An issue with horizontal scrolling as been fixed by swapping `sticky` for `fixed`.


packages/css-framework/src/components/_timeline.scss
```diff
@@ -49,28 +51,15 @@ $timeline-alt-bg-color: f.color("neutral", "white");
...
+  min-height: inherit;
 
-  li {
-    display: inline-flex;
-    align-items: center;
-    justify-content: flex-end;
+  span {
+    background-color: f.color("neutral", "white");
+    padding: f.spacing("2");
+    z-index: 1;
```
m7kvqbe1: Maybe this should be using the scoped framework function?

packages/css-framework/src/components/_timeline.scss
```diff
@@ -235,16 +162,51 @@ $timeline-alt-bg-color: f.color("neutral", "white");
...
+
+.timeline__row-header {
+  min-width: $timeline-row-header-width;
+  position: absolute;
+  left: 0;
+  height: inherit;
+  background-color: f.color("neutral", "white");
+  border-right: f.spacing("px") solid $timeline-border-color;
+  box-shadow: inset 0px 0px 0px 0px $timeline-border-color, 5px 0px 5px 0px rgba(0, 0, 0, 0.04);
+  z-index: 3;
```
m7kvqbe1: As above `z-index` function again.

packages/react-component-library/src/components/Timeline/Timeline.tsx
```diff
@@ -100,35 +107,32 @@ export const Timeline: React.FC<TimelineProps> = ({
...
+    TimelineDays.name,
+    TimelineWeeks.name,
+    TimelineMonths.name,
+    TimelineTodayMarker.name,
+  ])
+
+  const bodyChildren = extractChildren(children, [TimelineRows.name])
+
+  const innerClasses = classNames('timeline__inner', {
+    'timeline__inner--has-side': hasSide || hasTimelineSide,
```
m7kvqbe1, line 135: Could we combine the `hasSide || hasTimelineSide` into a single function or const so we don't have to repeat it?

m7kvqbe1 (changes requested):
LGTM just added a few comments

packages/react-component-library/src/components/Timeline/Timeline.tsx
thyhjwb6: I actually thought about this at the time (and just looked at it again) but it appears that `hasSide || hasTimelineSide` is the neatest it can be, for two reasons:

1. `hasTimelineSide` is actually deprecated because use of the `TimelineSide` is deprecated which means this issue goes away once we properly get rid of `TimelineSide`
2. You could repurpose the `hasTimelineSideComponent` `function` but then once you do that, what do you call it? `hasSide` and `hasTimelineSide` are currently being used so then you end up with `getHasSide` which for the sake of replacing `hasSide || hasTimelineSide` it feels like the code is becoming less readable and then the issue goes away with point 1 above

m7kvqbe1 (approved):
👍

Issue #1295, opened 2020-07-30
External changes to NumberInput value do not show up in the field


antictor:
### Steps to reproduce
1. Create a component or form that uses a number input
2. Change the value being sent to the NumberInput
3. The number input value in the visible field should change but does not, even though the value prop has changed

```
const NumberInputContainer: React.FC = () => {
  const [value, setValue] = useState(0)

  const onChange = (event: React.ChangeEvent<HTMLInputElement>) => {
    setValue(parseInt(event.target.value, 10))
  }

  const onClick = () => {
    const newValue = value + 1
    setValue(newValue)
  }

  return (
    <div>
      <NumberInput name="number-input" onChange={onChange} value={value} />
      <button data-testid="button" type="button" onClick={onClick}>
        Click me
      </button>
    </div>
  )
}```

PR #1296, opened 2020-07-30
fix(NumberInput): Fix value update bug #1295

antictor:
When the value property for a NumberInput changes it should update the input element.

This fix ensures that if the value prop changes it changes the value in the component state and as a result the field updates

The works by using the useEffect hook to detect a change in the value and set the value accordingly.

## Related issue
#1295

### DIFF
--- a/packages/react-component-library/src/components/NumberInput/useValue.ts
+++ b/packages/react-component-library/src/components/NumberInput/useValue.ts
@@ -1,4 +1,4 @@
-import React, { useMemo, useState } from 'react'
+import React, { useEffect, useMemo, useState } from 'react'
 import isNil from 'lodash/isNil'
 
 function isWithinRange(max: number, min: number, newValue: number) {
@@ -24,6 +24,10 @@ export function useValue(value: number) {
   const displayValue = useMemo(() => {
     return nextValue || committedValue
   }, [committedValue, nextValue])
+
+  useEffect(() => {
+    setCommittedValue(value)
+  }, [value])
 
   function setCommittedValueIfWithinRange(
     max: number,
--- a/packages/react-component-library/src/components/NumberInput/NumberInput.test.tsx
+++ b/packages/react-component-library/src/components/NumberInput/NumberInput.test.tsx
@@ -1,10 +1,37 @@
-import React from 'react'
+import React, { useState } from 'react'
 import '@testing-library/jest-dom/extend-expect'
-import { fireEvent, render, RenderResult } from '@testing-library/react'
+import {
+  fireEvent,
+  render,
+  RenderResult,
+  waitFor,
+} from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
 
 import { NumberInput } from './NumberInput'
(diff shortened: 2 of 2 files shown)


packages/react-component-library/src/components/NumberInput/NumberInput.tsx
```diff
@@ -87,6 +87,12 @@ export const NumberInput: React.FC<NumberInputProps> = ({
     setNextValue,
   } = useValue(value)
 
+  // if something changes the value, such as a call to setField in Formik
+  // then update the value of the field.
+  useEffect(() => {
```
thyhjwb6: This could work well if:

```
  useEffect(() => {
    setCommittedValue(value)
  }, [value])
```

was placed inside the `useValue` hook.

packages/react-component-library/src/components/NumberInput/NumberInput.tsx
```diff
@@ -87,6 +87,12 @@ export const NumberInput: React.FC<NumberInputProps> = ({
     setNextValue,
   } = useValue(value)
 
+  // if something changes the value, such as a call to setField in Formik
```
thyhjwb6: It feels like these comments are not required as this is a pattern we have across the codebase.

thyhjwb6 (changes requested):
Thanks for looking at this. A couple of comments.. 👍

antictor:
The unit tests fail and I'm not sure why. I've tried adding the waitFor to test the value, and it works in a realm browser... just not sure why it's failing the unit test

packages/react-component-library/src/components/NumberInput/NumberInput.tsx
antictor: I'll try

antictor:
Tests are correct, it works when it's in something that accepts the onChange to update the value but not when it manages its own state, though surely thats correct

packages/react-component-library/src/components/NumberInput/useValue.ts
```diff
@@ -25,6 +25,10 @@ export function useValue(value: number) {
     return nextValue || committedValue
   }, [committedValue, nextValue])
 
+  useEffect(() => {
+    setNextValue(value)
```
thyhjwb6: Does it work if this is `setCommittedValue`? That's what worked for me locally

antictor: That seems to have done the trick

packages/react-component-library/src/components/NumberInput/NumberInput.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { FormEvent } from 'react'
+import React, { FormEvent, useEffect } from 'react'
```
thyhjwb6: This `useEffect` is no longer required

thyhjwb6 (approved):
Looks good 👍 

It could work well if these commits are all squashed so it's a single commit for the change

antictor:
Merged commits, now just blocked by the security vulnerability

Issue #1336, opened 2020-08-12
Modal should respect primaryButton.icon if provided


jpveooys:
### Description

Currently, `Modal` accepts an `icon` prop on `primaryButton` but always overrides it with the forward arrow icon.

For a cancellation or deletion modal dialogue you may want to use a different or no icon, so it would be better if the forward arrow icon was only a default.

jpveooys:
Done in #1337

PR #1337, opened 2020-08-12
feat(Modal): Respect primaryButton.icon if set

jpveooys:
## Related issue

#1336

## Overview

This changes `Modal` to respect `primaryButton.icon` if it's set, instead of always overriding it with a forward arrow.

## Reason

From #1336:

> Currently, `Modal` accepts an `icon` prop on `primaryButton` but always overrides it with the forward arrow icon.
>
> For a cancellation or deletion modal dialogue you may want to use a different or no icon, so it would be better if the forward arrow icon was only a default.

## Work carried out

- [x] Allow Modal primaryButton.icon to be overriden
- [x] Added unit test
- [x] Added story

## Screenshot

image


m7kvqbe1 (approved):
LGTM 👍

Issue #1363, opened 2020-08-25
Type of DatePicker onChange prop no longer matches actual arguments


jpveooys:
### Steps to reproduce

The DatePicker's `onChange` prop is called with an object that has `startDate` and `endDate` properties. This hasn't changed, but the type of the prop was changed to a function that takes an object with `from` and `to` properties. 

This no longer matches how it's called.
 
This looks like a regression in:

https://github.com/Royal-Navy/design-system/commit/91365e8d105bdf332a32f7ee098abf1620c034d2#diff-420e48d44da60da55aa62adfed90a3b7R81-R84

jpveooys:
Fixed by #1365

PR #1365, opened 2020-08-25
fix(DatePicker): Correct type of onChange prop

jpveooys:
## Related issue

#1363 

## Overview

This corrects the type of the onChange prop to match how it's called and also so that it's the same as it was prior to 91365e8.

## Reason

> The DatePicker's `onChange` prop is called with an object that has `startDate` and `endDate` properties. This hasn't changed, but the type of the prop was changed to a function that takes an object with `from` and `to` properties. 
> 
> This no longer matches how it's called.

## Work carried out

- [x] Revert change to type of DatePicker onChange prop


m7kvqbe1 (approved):
👍

PR #1379, opened 2020-08-28
Add styling for Radio and Checkbox disabled states

### LINKED ISSUE #1345: Improve styling for disabled Checkbox
### Description

The `Checkbox` component has an `isDisabled` prop which prevents it from being toggled. This works, however the styling (including hover effect) is the same as a non-disabled checkbox.

It would be good if the component had specific styling for the disabled state (i.e. no hover effect and possibly greyed out) to avoid giving the impression it can be toggled.

Below is the test I used and how it currently looks:

```jsx
      <Checkbox label="This is disabled" name="field1" isDisabled />
      <Checkbox label="This is not disabled" name="field2" />
```

disabled-checkbox

m7kvqbe1:
## Related issue

Closes #1345

## Overview

Add styling for disabled states.

## Reason

>The Checkbox component has an isDisabled prop which prevents it from being toggled. This works, however the styling (including hover effect) is the same as a non-disabled checkbox.
>
>It would be good if the component had specific styling for the disabled state (i.e. no hover effect and possibly greyed out) to avoid giving the impression it can be toggled.

## Work carried out

- [x] Add styling for Radio and Checkbox
- [x] Add storybook examples for visual regression
- [x] Use classNames utility

## Screenshot

Screenshot 2020-08-28 at 11 29 43

## Developer notes

Also applies `cursor: not-allowed`.

### DIFF
--- a/packages/css-framework/src/components/_radio.scss
+++ b/packages/css-framework/src/components/_radio.scss
@@ -88,3 +88,22 @@ $radio-focus-width: 0.1rem;
   margin: f.spacing("4") f.spacing("6");
   font-size: f.font-size("base");
 }
+
+.rn-radio--is-disabled {
+  .rn-radio__label {
+    cursor: not-allowed;
+    color: f.color("neutral", "200");
+  }
+
+  input ~ .rn-radio__checkmark {
+    background-color: f.color("neutral", "000");
+    border: 1px solid f.color("neutral", "100");
+  }
+
+  input:hover ~ .rn-radio__checkmark,
+  input:active ~ .rn-radio__checkmark {
+    border: 1px solid f.color("neutral", "100");
+    box-shadow: none;
+    cursor: not-allowed;
+  }
+}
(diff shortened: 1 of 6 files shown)


packages/css-framework/src/components/_radio.scss
```diff
@@ -88,3 +88,21 @@ $radio-focus-width: 0.1rem;
...
 }
+
+.rn-radio--is-disabled {
+  .rn-radio__label {
+    cursor: not-allowed;
+    color: f.color("neutral", "200");
+  }
+
+  input ~ .rn-radio__checkmark {
+    border: 1px solid f.color("neutral", "100");
```
hxltrhuxze, line 100: Could we add `background-color: f.color("neutral", "000")` here? Will help differentiate the selectable and non-selectable checkboxes rather than just relying on the label for indication.

m7kvqbe1: Added fixups for this.

Screenshot 2020-08-28 at 11 48 48

thyhjwb6 (commented):
👍  looks good from my perspective

packages/css-framework/src/components/_radio.scss
hxltrhuxze: Sorry to be pedantic, but can we set the background to `000` and not `100`?

m7kvqbe1: No problem - this change has now been made.

Screenshot 2020-08-28 at 13 48 16

PR #1401, opened 2020-09-07
feat(DatePicker): Add ability to disable days

### LINKED ISSUE #1346: Ability to restrict DatePicker date selection to within two dates
### Description

For our exemplar project, we'd like the ability to restrict selection of a date in the DatePicker component to within two dates (including when in range mode).

In other words, the ability to specify an upper and lower bound such that it's not possible to select a date outside of those bounds. (It probably also shouldn't be possible to navigate to months outside of those bounds.)

(As far as I can see, this functionality doesn't exist at the moment, but let me know if I've missed something.)

m7kvqbe1:
## Related issue

Closes #1346

## Overview

Add ability to disable specific days.

## Link to preview

More detail about the underlying implementation is available here:

https://react-day-picker.js.org/examples/disabled

## Reason

>For our exemplar project, we'd like the ability to restrict selection of a date in the DatePicker component to within two dates (including when in range mode).

## Work carried out

- [x] Extend base interface
- [x] Add ability to disable dates (individual and chunks)
- [x] Allow `initialMonth` to be set without `startDate`
- [x] Add automated tests

## Screenshot

Screenshot 2020-09-07 at 14 22 53

### DIFF
--- a/packages/react-component-library/src/components/DatePicker/DatePicker.tsx
+++ b/packages/react-component-library/src/components/DatePicker/DatePicker.tsx
@@ -3,7 +3,12 @@ import differenceInMinutes from 'date-fns/differenceInMinutes'
 import { v4 as uuidv4 } from 'uuid'
 import classNames from 'classnames'
 import TetherComponent from 'react-tether'
-import DayPicker, { DateUtils, RangeModifier } from 'react-day-picker'
+import DayPicker, {
+  DateUtils,
+  RangeModifier,
+  DayPickerProps,
+  DayModifiers,
+} from 'react-day-picker'
 
 import { ComponentWithClass } from '../../common/ComponentWithClass'
 import { DatePickerInput } from './DatePickerInput'
@@ -34,6 +39,7 @@ export interface DatePickerProps extends ComponentWithClass {
   startDate?: Date
   value?: string
   isOpen?: boolean
+  disabledDays?: DayPickerProps['disabledDays']
 }
 
 function transformDates(startDate: Date, endDate: Date) {
@@ -62,6 +68,7 @@ export const DatePicker: React.FC<DatePickerProps> = ({
   startDate,
   value,
   isOpen,
+  disabledDays,
 }) => {
   const [state, setState] = useState<StateObject>({
     from: startDate,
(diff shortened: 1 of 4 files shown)


packages/react-component-library/src/components/DatePicker/DatePicker.tsx
```diff
@@ -17,15 +22,15 @@ export interface StateObject {
   to?: Date
 }
 
-export interface DatePickerProps extends ComponentWithClass {
+export interface DatePickerProps extends ComponentWithClass, DayPickerProps {
```
jpveooys: Looks like extending from `DayPickerProps` is bringing in a lot of stuff which seems to be mostly discarded – perhaps it could use a `Pick<>` or alternatively `disabledDays?: DayPickerProps['disabledDays']`?

jpveooys (commented):
One thing is if `startDate` isn't specified (i.e. you don't want to pre-select the start date), the initial month seems to be based on the current date so it might only contain disabled dates.

When I wrote the issue, I imagined it would do something like start on the first month that has some enabled dates. However, looking at the implementation, perhaps it would be easier to allow `initialMonth` to be set?

Let me know if you think this should be a separate issue.

m7kvqbe1:
> One thing is if `startDate` isn't specified (i.e. you don't want to pre-select the start date), the initial month seems to be based on the current date so it might only contain disabled dates.
> 
> When I wrote the issue, I imagined it would do something like start on the first month that has some enabled dates. However, looking at the implementation, perhaps it would be easier to allow `initialMonth` to be set?
> 
> Let me know if you think this should be a separate issue.

At the moment the `initialMonth` prop for the wrapped component is set by the exposed `startDate` prop - this was done to preserve the old API when we moved away from the previous underlying implementation. 

Would this satisfy your use case:

```typescript
initialMonth={startDate || initialMonth}
```

https://github.com/Royal-Navy/design-system/blob/a349d55a874de549aa6bac952edf8e4d11d3da65/packages/react-component-library/src/components/DatePicker/DatePicker.tsx#L162

jpveooys:
> initialMonth={startDate || initialMonth}

Yep, that should do the trick.

jpveooys (approved):
Looks good! :+1: 

Edit: Do the docs need updating for the new props?

PR #1402, opened 2020-09-07
fix(DatePicker): Prevent warning with empty initial value

### LINKED ISSUE #1380: Warning when DatePicker used in a form and without an initial value
### Steps to reproduce

1. Render:

    ```jsx
    <form>
      <DatePicker/>
    </form>
    ```

2. Run in dev mode
3. Check the browser console

The following warning is printed:

```
Warning: `value` prop on `input` should not be null. Consider using an empty string to clear the component or `undefined` for uncontrolled components.
```

This may be due to `transformDates()` returning `null` instead of an empty string when initial values aren't specified:

https://github.com/Royal-Navy/design-system/blob/8ea8fa9035783f91665b0e9987f7523b3f75ef8a/packages/react-component-library/src/components/DatePicker/DatePicker.tsx#L48

m7kvqbe1:
## Related issue

Closes #1380

## Overview

transformDates needs to return an empty string instead of `null` to prevent a React warning being thrown for the input value prop value.

### DIFF
--- a/packages/react-component-library/src/components/DatePicker/DatePicker.tsx
+++ b/packages/react-component-library/src/components/DatePicker/DatePicker.tsx
@@ -25,7 +25,7 @@ export interface DatePickerProps extends ComponentWithClass {
   label?: string
   name?: string
   onBlur?: (event: React.FormEvent) => void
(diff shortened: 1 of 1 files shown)


jpveooys (approved):
Looks good :+1:

PR #1411, opened 2020-09-08
ci(DesignTokens): Persist explicit paths for packages

m7kvqbe1:
## Overview

Persist only the package at hand instead of the entire monorepo.


m7kvqbe1:
This should fix the following error from Circle:

```
Error applying workspace layer for job e897e4d3-79e6-4507-aaa6-e6854f97a21c: Concurrent upstream jobs persisted the same file(s)
```

thyhjwb6 (approved):

PR #1415, opened 2020-09-08
fix(Timeline): Fix duplicated elements in some environments

### LINKED ISSUE #1153: Timeline duplicates elements with custom layout in production build

jpveooys:
## Related issue

Fixes #1153

## Overview

This fixes problems with duplicated elements in the Timeline component in production builds and/or in older browsers.

## Reason

Duplicated elements make the Timeline unusable.

## Work carried out

- [x] Changed Timeline logic to use reorganise child components based on function references instead of function names
- [x] Manually tested in various environments and browsers

#### Before

timeline-prod-dupe

#### After

timeline-prod-after

## Developer notes

These occurred because the Timeline logic was using the name property of functions (the functions being the Timeline components) to reorganise child components, however this failed in a few situations:

- after minification, the component function names weren't always unique
- after minification, the component functions were changed to anonymous functions, of which names are not supported in older browsers
- very old browsers don't support Function.name at all

(See also https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name.)

This change switches to using a strict equality on the types instead. This compares the functions by reference instead of name, which overcomes those three classes of problems from testing.

### DIFF
--- a/packages/react-component-library/src/components/Timeline/Timeline.tsx
+++ b/packages/react-component-library/src/components/Timeline/Timeline.tsx
@@ -3,7 +3,6 @@ import classNames from 'classnames'
 
 import { ComponentWithClass } from '../../common/ComponentWithClass'
 import { TimelineProvider } from './context'
-import { TimelineComponent } from './types'
 
 import {
   TimelineDays,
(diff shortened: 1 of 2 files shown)


m7kvqbe1 (approved):
LGTM - nice fix! 

Approved but worth waiting to also get eyes on from @thyhjwb6

thyhjwb6 (approved):
Nice change 💥 

I wonder if `extractChildren` becomes:

```
function extractChildren(
  children: timelineChildrenType | timelineChildrenType[],
  types: React.ReactElement['type'][],
  inverse?: boolean
) {
  return (children as []).filter((child: timelineChildrenType) => {
    return inverse ? !types.includes(child.type) : types.includes(child.type)
  })
}
```

to get rid of what seems like an unnecessary abstraction now.

jpveooys:
> I wonder if extractChildren becomes (...)

:+1:

m7kvqbe1 (approved):

PR #1428, opened 2020-09-14
feat(RangeSlider): Threshold and multiple handle enhancements

### LINKED ISSUE #1318: RangeSlider with thresholds
### Description
In our exemplar project, we need a range slider with multiple threshold points. Initially there are two thresholds, and later there will be three.

Points to consider:
- The slider should have graduations at specific intervals, for example a range of 0-100%, with separators every 10%.
- The thresholds should be draggable with a mouse cursor. They should also be keyboard accessible.
- In our use case, there should be no snapping to a specific value.
- We need rules to allow the top thresholds to always be the same as, or bigger than the next threshold down.
- The lower threshold should not be possible to drag to higher then the top threshold.
- We need colour coding of the base line, graduations and badges
- We need a badge to display a percentage value, and an actual numeric value.
- The badges should not appear over the top of each other. I'm not sure of the best way to do this when values are very close together.
- Optionally we may need a scale / values to be added to the slider and graduations.

Range slider with thresholds

m7kvqbe1:
## Related issue

Closes #1318

## Overview

Enhance the RangeSlider for threshold and multiple handle use cases.

## Reason

>In our exemplar project, we need a range slider with multiple threshold points. Initially there are two thresholds, and later there will be three.

## Work carried out

- [x] Render ticks active if values above
- [x] Add ability to apply custom value unit
- [x] Colour code ticks and handles with thresholds
- [x] Allow for tick display without step
- [x] Fix broken tick display with multiple handles
- [x] Add percentage badge variant
- [x] Update multiple handle example story

## Developer notes

I've created another ticket for keyboard control: https://github.com/Royal-Navy/design-system/issues/1429

### DIFF
--- a/packages/react-component-library/src/components/RangeSlider/RangeSlider.stories.tsx
+++ b/packages/react-component-library/src/components/RangeSlider/RangeSlider.stories.tsx
@@ -19,6 +19,23 @@ stories.add('Default', () => (
   />
 ))
 
+examples.add('Multiple handles', () => (
+  <div style={{ margin: '4rem' }}>
+    <RangeSlider
+      domain={[0, 40]}
+      mode={2}
+      values={[10, 20, 30]}
+      onChange={action('onChange')}
+      onUpdate={action('onUpdate')}
+      tracksLeft
+      tickCount={10}
+      thresholds={[40, 60]}
+      hasPercentage
+      displayUnit="pt"
+    />
+  </div>
+))
+
 examples.add('Stepped', () => (
   <RangeSlider
     domain={[0, 40]}
@@ -53,18 +70,6 @@ examples.add('Double threshold', () => (
     onUpdate={action('onUpdate')}
     tracksLeft
     thresholds={[40, 60]}
-  />
-))
-
-examples.add('Multiple handles', () => (
-  <RangeSlider
-    domain={[0, 40]}
-    step={10}
-    mode={1}
-    values={[10, 30]}
-    onChange={action('onChange')}
-    onUpdate={action('onUpdate')}
-    tickCount={4}
   />
 ))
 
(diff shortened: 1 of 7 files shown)


packages/react-component-library/src/components/RangeSlider/Slider.tsx
```diff
@@ -128,7 +128,7 @@ export const RangeSlider: React.FC<RangeSliderProps> = ({
             </div>
           )}
         </Tracks>
-        {step && (
+        {tickCount && (
```
thyhjwb6, line 135: Wondering why there isn't more accompanying this change? Like a test?

packages/react-component-library/src/components/RangeSlider/RangeSlider.stories.tsx
```diff
@@ -19,6 +19,23 @@ stories.add('Default', () => (
   />
 ))
 
+examples.add('Multiple handles', () => (
```
thyhjwb6, line 22: It feels strange this change is here because `thresholds` would not have any impact until it is implemented in a subsequent commit.

thyhjwb6 (approved):
Looks good 👍 

Some of the commits felt difficult to follow and wondering if some of them belonged as a single commit.

packages/react-component-library/src/components/RangeSlider/Slider.tsx
m7kvqbe1: > Looks good 👍
> 
> Some of the commits felt difficult to follow and wondering if some of them belonged as a single commit.

That's probably because of the way I was stepping through the bullet points in this issue: https://github.com/Royal-Navy/design-system/issues/1318

packages/react-component-library/src/components/RangeSlider/RangeSlider.stories.tsx
m7kvqbe1: Ah yeah you are right - this started off as a different commit with a temporary 'MSP' story.

I'll move this commit.

PR #1434, opened 2020-09-15
feat(RangeSlider): Improve keyboard control

### LINKED ISSUE #1429: RangeSlider Keyboard Control
Add keyboard control to RangeSlider to improve accessibility.

m7kvqbe1:
## Related issue

Closes #1429

## Overview

~~Use `button` instead of `div` for handle. Improve keyboard control.~~

Apply `tabIndex` to Handles and maintain `div` element due to accessibility AXE role error.

## Reason

>There were some weird edge cases where the handle wouldn't be controllable via the keyboard. Despite automated tests that were passing.

## Work carried out

- [x] Change handle element

### DIFF
--- a/packages/react-component-library/src/components/RangeSlider/Handle.tsx
+++ b/packages/react-component-library/src/components/RangeSlider/Handle.tsx
@@ -11,6 +11,7 @@ interface HandleProps {
   getHandleProps: GetHandleProps
   displayUnit?: string
   thresholds?: number[]
+  tabIndex: number
 }
 
 export const Handle: React.FC<HandleProps> = ({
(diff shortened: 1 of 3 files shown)


packages/react-component-library/src/components/RangeSlider/Handle.tsx
```diff
@@ -29,7 +29,7 @@ export const Handle: React.FC<HandleProps> = ({
   })
 
   return (
-    <div
+    <button
```
thyhjwb6: This was changed recently. You will notice the a11y error:

```
Rule: "aria-allowed-role" (ARIA role must be appropriate for the element)
```

m7kvqbe1: I've reverted back to the `div` and resolved the issue by applying `tabIndex` to handles.

thyhjwb6 (approved):

PR #1435, opened 2020-09-15
Add `TimelineHours` to display hour blocks in `Timeline`

### LINKED ISSUE #1196: Break Timeline day into block of hours
### Description
The _Timeline_  component only goes to detail level of days.  Sometime there is requirement  to break a day into block of hours like half(two 12 hours blocks each) or quarters(four 6 hour blocks each). 

Therefore similar to  _TimelineMonths, TimelineWeeks, TimelineDays_  would be nice to introduce _TimelineHours_

_TimelineHours_ should have a parameter about number of hours to group together like _hoursToGroup_ by default to have 1. with 12 as parameter one day will be brokendown into two six hour blocks and similarly with 6 as parameter one day will be brokendown into four six hour blocks. 

The Timeline component should start supporting Date with time not just Date

Please see below wireframe of one day broken down into four blocks of six hours each.

DayQuarterBreakDown

thyhjwb6:
## Related issue
Closes #1196

## Overview
Adds a `TimelineHours` component so that a `Timeline` can be composed with hour blocks.

## Reason
More granularity is required by downstream applications.

## Work carried out
- [x] Add `TimelineHours`
- [x] Specify block size
- [x] Offset `Event` based on time
- [x] Allow custom render of hour
- [x] Add constraints
- [x] Update docs 
- [x] Deprecate `dayWidth` in favour of `unitWidth`

## Screenshot
Screenshot 2020-09-15 at 13 48 38


packages/css-framework/src/components/_timeline.scss
```diff
@@ -122,30 +122,42 @@ $timeline-row-header-width: 16rem;
...
 .timeline__week-title--renderDefault,
-.timeline__day-title--renderDefault {
+.timeline__day-title--renderDefault,
+.timeline__hour-title--renderDefault {
   font-weight: 600;
   font-size: f.font-size("m");
   color: f.color("neutral", "400");
-  margin-left: f.spacing("4");
   background-color: $timeline-bg-color;
   @include m.z-index("overlay", 2);
```
m7kvqbe1: Does this work with an integer instead of a string for the second arg?

m7kvqbe1 (commented):
I notice the midnight hours block spanning across days though on my storybook (see screenshot). Looks like there is something wrong with that hour width calc?

Screenshot 2020-09-21 at 09 07 47

packages/css-framework/src/components/_timeline.scss
thyhjwb6: Yeah this is in the `README` of the `css-framework`:

```
.foo {
  @include z-index("modal", 132);
}
```

m7kvqbe1 (approved):
LGTM

PR #1455, opened 2020-09-21
fix(RangeSlider): Set correct handle `tabIndex`

m7kvqbe1:
# Overview

`tabIndex` for all handles should actually be 0. Although it works in isolation values greater than 0 interferes with the tab flow of a page and flags an AXE a11y violation.


m7kvqbe1:
https://dequeuniversity.com/rules/axe/3.5/tabindex?application=axe-puppeteer

thyhjwb6 (approved):
👍

PR #1465, opened 2020-09-22
fix(DatePicker): Ensure `en-GB` is used for format

### LINKED ISSUE #1427: DatePicker - Date input is an unexpected date format (mm/dd/yyyy) in older browsers
In Chrome 69 the date input is as expected i.e dd/mm/yyyy, however in older browsers it becomes mm/dd/yyyy

Noticeable on v2.13.0 and v.2.14.0-alpha.22

### Steps to reproduce
<Field
  component={DatePicker}
  ...
</>

thyhjwb6:
## Related issue
Fixes #1427 

## Overview
Old versions of Chrome do not have English UK in the language list. Rather than expecting the browsers to update language settings this change will provide consistency for all implementations of the `DatePicker`.

## Reason
Consistency across different versions of Chrome.

## Work carried out
- [x] Specify format

### DIFF
--- a/packages/react-component-library/src/components/DatePicker/DatePicker.stories.tsx
+++ b/packages/react-component-library/src/components/DatePicker/DatePicker.stories.tsx
@@ -15,7 +15,7 @@ const examples = storiesOf('DatePicker/Examples', module)
 stories.add('Default', () => {
   return (
     <DatePicker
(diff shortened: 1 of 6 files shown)


m7kvqbe1 (commented):
Could we add a test for this?

m7kvqbe1 (approved):

PR #1475, opened 2020-09-23
Fix TextArea label styling

### LINKED ISSUE #1409: TextArea label styling is not correct
The TextArea label should reduce in size when content is added.

At the moment it sits flush with the content within the input.

Screenshot 2020-09-08 at 10 38 50

thyhjwb6:
## Related issue
Closes #1409 

## Overview
The `TextArea` `label` styling is not correct when the element has text content.

## Reason
Classes are missing and the state is not being correctly managed. It becomes difficult to read.

## Work carried out
- [x] Fix CSS
- [x] Fix state management
- [x] Apply same change to `TextInput`

## Screenshot
textarea


m7kvqbe1 (approved):
👍

PR #1477, opened 2020-09-24
fix(Timeline): Allow arbitrary event `children`

### LINKED ISSUE #1444: Compound timeline component children prop type throws error
### Steps to reproduce
1. add a child to a timeline event 
2. type error: Types of property 'children are incompatible

thyhjwb6:
## Related issue
Closes #1444 

## Overview
`TimelineEvent` only allows `string` `children` and not arbitrary `children`. This changes the `children` type to `ReactNode`.

## Reason
It should be possible to compose `TimelineEvent` with elements other than `string`.

## Work carried out
- [x] Update type

## Developer notes
A previously developed `render` prop allows for custom content too.


m7kvqbe1 (approved):
LGTM

PR #1511, opened 2020-10-05
Fix NumberInput with unit and user keyboard input

### LINKED ISSUE #1458: Number input with trailing Unit usability
When the number input has a trailing unit, some strange behaviours emerge:

To replicate:
- Fill the number input with a value
- move your cursor to the left of the `m3`
- If your cursor is after the space directly before `m3`, it bumps the cursor to the end and doesn't edit the number

Editing the number sometimes also produces strange behaviour (e.g. the number not being deleted/can't update)

thyhjwb6:
## Related issue
Fixes #1458 

## Overview
Separates out the `unit` from the `input` `value` as this was causing a range of issues.

## Reason
User is not able to edit value as expected.

## Work carried out
- [x] Fix issues
- [x] Refactor existing code to simplify

## Screenshot
number-input

## Notes
- A separate ticket #1526 has been raised to convert the component to use Styled Components

### DIFF
--- a/packages/react-component-library/src/components/NumberInput/NumberInput.tsx
+++ b/packages/react-component-library/src/components/NumberInput/NumberInput.tsx
@@ -1,4 +1,4 @@
-import React, { FormEvent } from 'react'
+import React from 'react'
 import classNames from 'classnames'
 import { isFinite, isNil, without } from 'lodash'
 import { v4 as uuidv4 } from 'uuid'
@@ -12,7 +12,9 @@ import { useFocus } from './useFocus'
 import { useValue } from './useValue'
 import { UNIT_POSITION } from './constants'
 
-export type UnitPosition = typeof UNIT_POSITION.AFTER | typeof UNIT_POSITION.BEFORE
+export type UnitPosition =
+  | typeof UNIT_POSITION.AFTER
+  | typeof UNIT_POSITION.BEFORE
 
 export interface NumberInputProps {
   autoFocus?: boolean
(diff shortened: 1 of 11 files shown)


packages/react-component-library/src/components/NumberInput/NumberInput.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { FormEvent } from 'react'
+import React from 'react'
 import classNames from 'classnames'
 import { isFinite, isNil, without } from 'lodash'
```
m7kvqbe1: Looks like `without` is an unused import.

packages/react-component-library/src/components/NumberInput/useInputText.ts
```diff
@@ -0,0 +1,65 @@
+import { useEffect, useRef, useState } from 'react'
+import { isFinite } from 'lodash'
+
+import { UnitPosition } from './NumberInput'
+import { UNIT_POSITION } from './constants'
+
+const EXTRA_SPACING = 3
+
+function getTextWidth({
```
m7kvqbe1, line 9: Clever technique! I assume it plays nice cross browser?

m7kvqbe1 (commented):
Quite a bit of complexity but I think the solution is nice.

I tried hard to break it and couldn't - user experience is really nice now.

packages/react-component-library/src/components/NumberInput/useInputText.ts
thyhjwb6: This is IE 11..

number-input-ie11

m7kvqbe1 (approved):

PR #1518, opened 2020-10-06
fix(Timeline): Resolve daylight savings offset issue

### LINKED ISSUE #1489: Timeline positioning slightly off after clocks go back
### Steps to reproduce

1. Check local machine is set to the UK (Europe/London) time zone
2. Render:

```jsx
  <Timeline
    startDate={new Date(2020, 8, 1)}
    endDate={new Date(2020, 11, 1)}
    today={new Date(2020, 10, 1)}
  >
    <TimelineTodayMarker />
    <TimelineMonths />
    <TimelineDays />
    <TimelineRows>
      <TimelineRow name="Row 1">
        <TimelineEvents>
          <TimelineEvent
            startDate={new Date(2020, 8, 2)}
            endDate={new Date(2020, 8, 4)}
          >
            Event 1
          </TimelineEvent>
        </TimelineEvents>
      </TimelineRow>
    </TimelineRows>
  </Timeline>
```

The today marker is slightly offset to the right (presumably as an extra hour is introduced when the clocks go back at the end of October):

image

Older versions:

image

This seems to have been introduced when support for time was added to the timeline. It affects custom timeline elements using `useTimelinePosition` too. For our exemplar app it causes some alignment problems with vertical elements in the timeline.

A possible simple mitigation (not tested...) might be to calculate the difference between the two dates as the number of days first, and then only calculate the number of seconds (or hours) for the remainder. The size of these two components could then be calculated independently and added together.

So e.g. in the example, 2020-09-01 00:00 to 2020-11-01 00:00 is 61 days and 0 seconds/hours. Similarly 2020-09-01 00:00 to 2020-11-01 06:00 would be 61 days and 6 hours.

m7kvqbe1:
## Related issue

Closes #1489

## Overview

Resolve daylight savings offset issue in `useTimelinePosition` hook.

## Reason

>Daylight savings meant that objects utilising the useTimelinePosition hook were incorrectly offset.

## Work carried out

- [x] Fix some styling issues
- [x] Change the way `getOffset` and `getWidth` are calculated

## Screenshot

`new Date(2020, 10, 1, 12)`:

Screenshot 2020-10-06 at 11 11 29

`new Date(2020, 10, 1)`:

Screenshot 2020-10-06 at 11 09 43

## Developer notes

I've limited the granularity to hours.

### DIFF
--- a/packages/react-component-library/src/components/Timeline/hooks/useTimelinePosition.ts
+++ b/packages/react-component-library/src/components/Timeline/hooks/useTimelinePosition.ts
@@ -1,7 +1,7 @@
 import { useContext } from 'react'
 
 import {
-  differenceInSeconds,
+  differenceInCalendarDays,
   isBefore,
   isAfter,
 } from 'date-fns'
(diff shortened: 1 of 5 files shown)


packages/react-component-library/src/components/Timeline/hooks/useTimelinePosition.ts
```diff
@@ -1,28 +1,34 @@
...
-    SECONDS_IN_DAY
-  )
+function getWidth (
+  startDate: Date,
+  endDate: Date
+): number {
+  return differenceInCalendarDays(
+    new Date(endDate),
+    new Date(startDate)
+  ) + timeOffset(endDate)
```
jpveooys: Do we need to consider the case where the first day is partial one here as well?

m7kvqbe1: Please see latest fixup.

jpveooys (approved):
Looks good – seems to be behaving well from some quick tests :+1:

PR #1529, opened 2020-10-09
fix(Timeline): Forward className prop

jpveooys:
## Overview

The `Timeline` had a `className` prop that wasn't used. This forwards it to the root child component so library users can override styles more easily.

The prop was also added to the docs.

## Reason

To make it easier for library users to override styles (e.g. do `styled(Timeline)`).

## Work carried out

- [x] Forward className prop
- [x] Add test
- [x] Update timeline docs

## Developer notes

The prop already existed as `TimelineProps` inherits from `ComponentWithClass`.


m7kvqbe1 (approved):
👍

thyhjwb6 (approved):
👍

PR #1531, opened 2020-10-09
fix(Formik): Display nested field errors correctly

### LINKED ISSUE #1498: withFormik doesn't show errors for nested fields
If the field is in a nested object in FormikValues (i.e. `x.y`) then `withFormik` doesn't display error text related to that field.

It looks like a small change from `errors[fieldName]` to `fieldError` on this line would sort this:

https://github.com/Royal-Navy/design-system/blob/5f509e7ddd8858c66a1dd66c2dab9b94ac092da2/packages/react-component-library/src/enhancers/withFormik.tsx#L33

jpveooys:
## Related issue

Fixes #1498

## Overview

This fixes a small problem with the `withFormik` HOC so that errors for nested fields display correctly.

## Reason

The error text for nested fields was not displayed (although error styling etc. functioned correctly).

## Work carried out

- [x] Corrected logic to handle nested fields
- [x] Added tests

## Developer notes

Nested field support in Formik is described here: https://formik.org/docs/guides/arrays#nested-objects

### DIFF
--- a/packages/react-component-library/src/enhancers/__tests__/withFormik.test.tsx
+++ b/packages/react-component-library/src/enhancers/__tests__/withFormik.test.tsx
@@ -1,0 +1,67 @@
+import '@testing-library/jest-dom/extend-expect'
+import { render, RenderResult } from '@testing-library/react'
+import React from 'react'
+
+import { FieldProps } from '../../common/FieldProps'
+import { FormProps } from '../../common/FormProps'
+import { withFormik } from '../withFormik'
+
+const DummyComponent: React.FC = (props) => (
+  <div data-testid="dummy-component" {...props} />
+)
+const DummyFormikComponent = withFormik(DummyComponent)
+
+describe('withFormik', () => {
+  let field: FieldProps
+  let form: FormProps
+  let wrapper: RenderResult
+  const errorText = 'Test error'
+
+  describe('when a touched nested field has an error', () => {
+    beforeEach(() => {
+      field = {
+        name: 'nested.field',
+        value: 'false',
+        onChange: jest.fn(),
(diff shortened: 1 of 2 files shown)


packages/react-component-library/src/enhancers/__tests__/withFormik.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+          nested: {
+            field: true,
+          },
+        },
+      }
+
+      wrapper = render(<DummyFormikComponent field={field} form={form} />)
+    })
+
+    it('adds the aria aria-invalid attribute', () => {
```
jpveooys, line 45: I based these tests on similar tests for components elsewhere in the code base.

packages/react-component-library/src/enhancers/__tests__/withFormik.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+      expect(wrapper.getByTestId('dummy-component')).toHaveAttribute(
+        'aria-invalid',
+        'true'
+      )
+    })
+
+    it('adds the aria aria-describedby attribute', () => {
+      expect(wrapper.getByTestId('dummy-component')).toHaveAttribute(
+        'aria-describedby'
```
thyhjwb6: It could work well to assert the value of the attribute.

thyhjwb6 (approved):
👍  Looks good

packages/react-component-library/src/enhancers/__tests__/withFormik.test.tsx
jpveooys: Done!

PR #1571, opened 2020-10-19
fix(RangeSlider): Only invoke `onUpdate` if provided

### LINKED ISSUE #1528: `RangeSlider` `onUpdate` should not be called if it has not been specified
`onUpdate` is optional and should not be called when it has not been specified.

m7kvqbe1:
## Related issue

Closes #1528

## Overview

Only invoke the optional `onUpdate` callback if it is provided.


thyhjwb6 (approved):
👍

PR #1585, opened 2020-10-20
feat(RangeSlider): Add support for custom value formatter

### LINKED ISSUE #1561: RangeSlider: Add support for custom value formatter
### Background

In the range slider in our exemplar project the primary unit of the slider is percentages, with steps of 1%. This implies a domain of [0, 100] and step of 1.

However, we also need to show a secondary label showing the value the percentage is equivalent to. Currently, to get this to display, it seems we need to use a domain representing the secondary value (e.g. [0, 954]) and convert from and to percentages in the client app. This has a few undesirable side effects, for example the ticks snap to round values in the domain (e.g. 0, 100, 200), instead of the desired 0%, 10%, 20% etc.

### Proposed solution

My proposal is to add a new optional prop to the component, named e.g. `valueFormatter`, that takes a function with a signature of roughly `(value: number, percentage: number, displayUnit: string) => string`. The function would return the string to display and would be called here:

https://github.com/Royal-Navy/design-system/blob/99229de27392dce7446dc6158259f0570ba16b5c/packages/react-component-library/src/components/RangeSlider/Handle.tsx#L44

This would enable the domain to be kept to [0, 100] with a step of 1 in our exemplar app but still display the secondary value the percentage represents (with `hasPercentage` being used to also show the percentage).

(If I've missed anything let me know.)

jpveooys:
## Related issue

Fixes #1561

## Overview

This adds a new `formatValue` prop which can be used to customise how values are formatted in handle labels.

## Reason

This can be used for use cases where `displayUnit` doesn't suffice:

- unit conversions (i.e. internal state is in a different unit/scale to what is displayed)
- unit prefixes
- custom number formatting (thousand separators etc.)

It might be useful to do the same for formatting the percentage, but I haven't done that at this stage.

## Work carried out

- [x] Added new `formatValue` prop
- [x] Exported relevant types
- [x] Added test
- [x] Added story
- [x] Updated docs


packages/react-component-library/src/components/RangeSlider/Slider.tsx
```diff
@@ -10,9 +10,15 @@ import {
...
+import {
+  Handle,
+  RangeSliderValueFormatter,
+  Track,
+  ThresholdTrack,
+  Tick,
+} from '.'
 
-interface RangeSliderProps
+export interface RangeSliderProps
```
jpveooys, line 21: Have exported this as it's useful for client apps.

packages/react-component-library/src/components/RangeSlider/Slider.tsx
```diff
@@ -43,18 +50,27 @@ export const RangeSlider: React.FC<RangeSliderProps> = ({
...
-  displayUnit,
+  displayUnit = '',
+  formatValue,
   ...rest
 }) => {
   const [sliderValues, setSliderValues] = useState(values)
 
-  const onUpdateHandler = (newValues: ReadonlyArray<number>): void => {
-    setSliderValues(newValues)
+  const onUpdateHandler = useCallback(
```
jpveooys, line 59: `useCallback` might help with performance here as `Slider` is a `PureComponent` (in our exemplar app the thresholds are updated dynamically with the values, so optimisation would be beneficial).

m7kvqbe1 (approved):
LGTM nice call with the memoization stuff - have you noticed it being slow/laggy in an app?

jpveooys:
> LGTM nice call with the memoization stuff - have you noticed it being slow/laggy in an app?

It was OK in modern browsers but I noticed some slowness in older browsers – quite laggy in dev mode, a little bit of lag in prod mode (but an acceptable amount I think).

PR #1589, opened 2020-10-21
ci(CodeOwners): Update to use correct syntax

m7kvqbe1:
CodeOwners feature was not working correctly because syntax was slightly wrong.


thyhjwb6 (approved):
oh 🤣

PR #1593, opened 2020-10-26
feat(ContextMenu): Match design to Sketch component

hxltrhuxze:
- Updates the ContextMenu to match the new Right Click designs in Sketch.
- Swaps the component width in favour of a min- & max-width to give more room to longer items when needed
- ensuring overflowing text doesn't wrap, but instead truncates with an ellipsis.

Screenshot 2020-10-26 at 16 32 32


m7kvqbe1 (approved):
👍

PR #1603, opened 2020-10-29
fix(SidebarE): Rename and export

m7kvqbe1:
## Overview

Component names were conflicting with legacy implementation and export was disabled.


thyhjwb6 (approved):
👍

PR #1604, opened 2020-10-30
Add ability to hide/show Popover

### LINKED ISSUE #1566: Popover - add ability to activate component using onClick instead of onHover
### Description
We are currently using the Popover component to display information when hovering over an icon. Instead of it automatically appearing onHover, it would be better if we can make use of a click event instead. 

Is there another component with similar styling that will allow us to carry out the desired UX or does this fit as a feature enhancement?

thyhjwb6:
## Related issue
Closes #1566

## Overview
Adds the ability to hide/show `Popover` when clicking rather than just hovering.

## Reason
Required by applications.

## Work carried out
- [x] Refactor existing
- [x] Add click handlers

### DIFF
--- a/packages/react-component-library/src/components/Popover/Popover.tsx
+++ b/packages/react-component-library/src/components/Popover/Popover.tsx
@@ -2,13 +2,17 @@ import React, { useRef, useState } from 'react'
 import classNames from 'classnames'
 import TetherComponent from 'react-tether'
 
-import { POPOVER_CLOSE_DELAY, POPOVER_PLACEMENT, POPOVER_PLACEMENTS } from './constants'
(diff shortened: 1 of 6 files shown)


packages/react-component-library/src/components/Popover/Popover.tsx
```diff
@@ -80,28 +125,13 @@ export const Popover: React.FC<PopoverProps> = ({
    */
   return (
     <>
-      {/*
-      // @ts-ignore */}
+      {/* @ts-ignore */}
```
m7kvqbe1, line 85: Does this create a type error now? 

Previously the comment had to be in that weird format for the `@ts-ignore` to be respected.

m7kvqbe1 (approved):
LGTM - just a quick comment about the ts comment

PR #1618, opened 2020-11-02
feat(Popover): Use styled-components

### LINKED ISSUE #1605: Convert Popover to styled-components

thyhjwb6:
## Related issue
Closes #1605 

## Overview
Converts `Popover` to use styled-components.

## Reason
Migrating all components.

## Work carried out
- [x] Clean up old references
- [x] Convert to sc

## Developer notes
- There was an old reference to styled-components in the docs site which was causing issues with referencing jest-styled-components
- For now checking the `opacity` is enough to replace the tests that were checking `is-visible`. A longer term plan is being thought about as maybe checking style properties is not ideal


m7kvqbe1 (approved):
👍

PR #1626, opened 2020-11-03
refactor(RangeSlider): Migrate to styled-components

### LINKED ISSUE #1620: Convert RangeSlider to styled-components

m7kvqbe1:
## Related issue

Closes #1620

## Overview

Migrate styles from SASS to styled-components.

## Reason

>We need to migrate all existing components to `styled-components`.

## Work carried out

- [x] Migrate to `styled-components`
- [x] Rework some tests


packages/react-component-library/src/components/RangeSlider/Handle.tsx
```diff
@@ -14,6 +17,97 @@ interface HandleProps {
   thresholds?: number[]
 }
 
+interface StyledHandleProps {
+  $isActive?: boolean
+  thresholdColor?: ThresholdColor
```
thyhjwb6: I am just trying to develop an understanding of how we use transient props on styled-components. Would it make more sense if we default to use transient props for all styled-components and then only make it non transient (normal) if it needs to be drilled down? If this is the case then it would imply `thresholdColor` would be transient `$transientColor` because it does not need drilling down to `div`.

packages/react-component-library/src/components/RangeSlider/ThresholdTrack.tsx
```diff
@@ -1,21 +1,51 @@
...
+interface ChunkProps {
   left: number
   width: number
   maxWidth: number
   testId: string
+  thresholdColor?: ThresholdColor
 }
 
+interface StyledChunkProps {
+  thresholdColor?: ThresholdColor
```
thyhjwb6: Again here I wonder if this could be transient too.

packages/react-component-library/src/components/RangeSlider/Tick.tsx
```diff
@@ -30,17 +76,15 @@ export const Tick: React.FC<TickProps> = ({
...
-  const tickClasses = classNames('rn-rangeslider__tick', {
-    'is-active': isActive(values, tickValue),
-    ...useThresholdClasses(percent, thresholds, 'tick'),
-  })
-
   return (
     <div data-testid="rangeslider-tick">
-      <div className={tickClasses} style={{ left: `${tick.percent}%` }} />
+      <StyledTick
+        style={{ left: `${tick.percent}%` }}
```
thyhjwb6: Could we do something to make this part of `StyledTick`?

packages/react-component-library/src/components/RangeSlider/RangeSlider.test.tsx
```diff
@@ -156,30 +158,35 @@ describe('RangeSlider', () => {
...
+            bubbles: true,
+            cancelable: true,
+          })
+        )
 
-    it('should apply the stateful `is-reversed` class', () => {
-      expect(wrapper.queryByTestId('rangeslider').classList).toContain(
-        'is-reversed'
-      )
+        fireEvent.keyDown(wrapper.getByTestId('rangeslider-handle'), {
```
thyhjwb6, line 167: The `@testing-library/user-event` seems to make the use typing neater. Could be worth looking at? We have used it in `NumberInput`.

thyhjwb6 (changes requested):
Looks good, a few comments 👍

packages/react-component-library/src/components/RangeSlider/Handle.tsx
m7kvqbe1: Yeah ok that sounds like a sensible convention. 

So everything unique to the StyledComponent that doesn't need to be drilled to the wrapped component should be marked as transient.

packages/react-component-library/src/components/RangeSlider/Tick.tsx
m7kvqbe1: I've slightly changed the naming to make it less confusing but the wrapper div is a requirement.

packages/react-component-library/src/components/RangeSlider/RangeSlider.test.tsx
m7kvqbe1: I've changed the clicks to se userEvent but couldn't get the type API to play nice - seems to be more aimed at inputs:

https://github.com/testing-library/user-event#typeelement-text-options

packages/react-component-library/src/components/RangeSlider/Handle.tsx
thyhjwb6: It feels like that makes sense rather than having special cases for transient props. Was trying to find other documentation on a convention like this.

packages/react-component-library/src/components/RangeSlider/Tick.tsx
thyhjwb6: Sorry if this was not clear - I was referring to the `style` as it feels strange that we would have a styled-component and then a `style` prop here.

packages/react-component-library/src/components/RangeSlider/RangeSlider.test.tsx
thyhjwb6: Okay fair enough. I now realise that the `userEvent.click` looks nice so I will use that in other places too.

thyhjwb6:
Looks good 👍 

I left a comment about the `style` attribute.

thyhjwb6 (approved):
👍

PR #1641, opened 2020-11-06
fix(ReactComponentLibrary): Prevent `isInvalid` prop drill to DOM node

### LINKED ISSUE #1591: Warning emitted to console for Formik TextInputs
### Steps to reproduce

1. Check out `master`
2. Visit http://localhost:6006/?path=/story/textinput--formik locally
3. Check the console

The following warning is logged:

> Warning: React does not recognize the `isInvalid` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `isinvalid` instead. If you accidentally passed it from a parent component, remove it from the DOM element.

m7kvqbe1:
## Related issue

Closes #1591

## Overview

Prevent prop from being incorrectly applied to DOM node.

## Reason

>Warning emitted to console for Formik TextInputs

## Work carried out

- [x] Fix up some interfaces
- [x] Do not drill prop to DOM node

### DIFF
--- a/packages/react-component-library/src/components/Checkbox/Checkbox.tsx
+++ b/packages/react-component-library/src/components/Checkbox/Checkbox.tsx
@@ -3,8 +3,11 @@ import classNames from 'classnames'
 import { v4 as uuidv4 } from 'uuid'
 
 import { ComponentWithClass } from '../../common/ComponentWithClass'
+import { InputValidationProps } from '../../common/InputValidationProps'
 
-export interface CheckboxProps extends ComponentWithClass {
+export interface CheckboxProps
+  extends ComponentWithClass,
+    InputValidationProps {
   id?: string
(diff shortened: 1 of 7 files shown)


jpveooys (approved):
Looks good to me!

thyhjwb6 (approved):
👍

thyhjwb6:
Maybe we should think about adding tests for these to prevent future regressions. Not urgent but something to think about.

m7kvqbe1:
> Maybe we should think about adding tests for these to prevent future regressions. Not urgent but something to think about.

Yeah good shout - I think a test for every form component where we drill arbitrary DOM attributes with ...rest.

We'd need to reference it against a list of valid attributes. Maybe that can be done using built in DOM node TS types.

edit: I have created at ticket for this

PR #1655, opened 2020-11-10
Fix timeline row accessibility

### LINKED ISSUE #1446: (Spike): Investigate accessibility unit tests failing for Timeline component
### Steps to reproduce
Performing accssibility test on Timeline component using `jest-axe`. Please see below steps to reproduce.

1. BookingTimeline

2. BookingTimelineTest

3. AccessibilityTestErrorMsg

### Screenshot
*Add a screenshot*

thyhjwb6:
## Related issue
Fixes #1446 

## Overview
When all events are outside of the range then rows are rendered but no events are rendered. This change checks the rendered cells and if there are not any then it shows the `No events` hidden cell.

## Reason
This causes accessibility issues because there is a row without cells.

## Work carried out
- [x] Refactor existing `noData`
- [x] Fix missing cell

### DIFF
--- a/packages/react-component-library/src/components/Timeline/hooks/useTimelineRowContent.ts
+++ b/packages/react-component-library/src/components/Timeline/hooks/useTimelineRowContent.ts
@@ -1,0 +1,26 @@
+import { useEffect, useRef, useState, MutableRefObject } from 'react'
+
+export function useTimelineRowContent(
+  isHeader: boolean,
+  deps?: ReadonlyArray<any>
+): {
+  noCells: boolean
+  rowContentRef: MutableRefObject<HTMLDivElement>
+} {
+  const rowContentRef = useRef(null)
+  const [noCells, setNoCells] = useState<boolean>(false)
+
+  if (!isHeader) {
+    useEffect(() => {
+      const cells = rowContentRef.current.querySelectorAll('[role="cell"]')
+      if (!cells.length) {
+        setNoCells(true)
+      }
+    }, deps)
+  }
+
+  return {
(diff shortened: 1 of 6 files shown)


packages/react-component-library/src/components/Timeline/hooks/useTimelineRowContent.ts
```diff
@@ -0,0 +1,25 @@
...
+  const rowContentRef = useRef(null)
+  const [noCells, setNoCells] = useState<boolean>(false)
+
+  if (!isHeader) {
+    useEffect(() => {
+      const cells = rowContentRef.current.querySelectorAll('[role="cell"]')
+      if (!cells.length) {
+        setNoCells(true)
+      }
+    })
```
thyhjwb6: I will add `[children]` here so that it's not firing all of the time.

m7kvqbe1 (commented):
Nice - I think this is an elegant solution It is showing this a11y warning though in Storybook:

Screenshot 2020-11-12 at 15 11 51

m7kvqbe1:
Oh no wait - it seems to be ok now and I can't reproduce it.

Might have been storybook cache or something.

edit: If I quickly flip between stories it flags it sometimes it seems.

thyhjwb6:
Having reproduced this it seems to be a quirk when moving between Storybook examples and not recognising that the cell has been loaded. As the visual regression and unit tests pass in CI it feels like we can accept this as a Storybook quirk.

m7kvqbe1 (approved):

PR #1670, opened 2020-11-13
Fix NumberInput borders

### LINKED ISSUE #1640: Border issues with number steppers
This cropped up in QA testing in our exemplar project.

The borders on number steppers have a couple of issues:

- The curved border looks very heavy in the corners - it's not smooth as you'd expect
- There's an overlap with the red and blue borders

I've grabbed a photo of the issue (the product is in a different environment). Let me know if you have any questions or want to see a demo.

IMG_20201105_152602

thyhjwb6:
## Related issue
Fixes #1640 

## Overview
Fixes the `NumberInput` border.

## Reason
There are border issues:
- corners do not look sharp
- overlap from up/down buttons

## Work carried out
- [x] Refactor code
- [x] Apply fix

## Screenshot
Screenshot 2020-11-13 at 15 59 39

### DIFF
--- a/packages/react-component-library/src/components/DatePicker/DatePickerInput.tsx
+++ b/packages/react-component-library/src/components/DatePicker/DatePickerInput.tsx
@@ -4,6 +4,7 @@ import { selectors } from '@royalnavy/design-tokens'
 
 import { ComponentWithClass } from '../../common/ComponentWithClass'
 import { DropdownIndicatorIcon } from '../Dropdown/DropdownIndicatorIcon'
+import { getOuterWrapperBorder } from '../../styled-components'
 
 export interface DatePickerInputProps extends ComponentWithClass {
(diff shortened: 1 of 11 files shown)


m7kvqbe1:
I think the border needs to match `TextInput`. Looks like the thickness might be slightly too low:

Screenshot 2020-11-16 at 13 31 05

Does this change also need to be applied to the up and down button focus states:

Screenshot 2020-11-16 at 13 29 47

m7kvqbe1 (changes requested):
I think the border needs to match `TextInput`. Looks like the thickness might be slightly too low:

Screenshot 2020-11-16 at 13 31 05

Does this change also need to be applied to the up and down button focus states:

Screenshot 2020-11-16 at 13 29 47

m7kvqbe1 (approved):

PR #1678, opened 2020-11-18
feat(Checkbox): Expose input checked prop

### LINKED ISSUE #1398: Checkbox component should expose input checked prop
### Description

Currently, there is an `isChecked` prop which maps to the input `defaultChecked` prop. However, the input's `checked` prop should be exposed.

This is for a few reasons:

1. Currently, there is no easy way to properly synchonrise a `Checkbox` with external state changes. 
  
   For example, imagine a select all `Checkbox` with individual `Checkbox`es underneath, for example. The individual check boxes need to be updated when the select all check box is toggled. (The current `isChecked` doesn't allow this as it can only be used once as it maps to the input's `defaultChecked`.)

2. Formik has built-in support for check box groups when a Formik `Field` is passed `type="checkbox"`. 

    This applies in particular to `initialValues` and `setFieldValue()` as the field value is an array which needs to be mapped to individual booleans for each field. Formik works out the value of and sets the `checked` prop itself when `type="checkbox"` is used.

    This currently happens to work as Formik's `Field` isn't properly type-checked and the `Checkbox` component forwards unknown props:
   
    https://github.com/Royal-Navy/design-system/blob/97533aa3d488a0d2c0db505b729f900d8db8848a/packages/react-component-library/src/components/Checkbox/Checkbox.tsx#L30

3. An input's `defaultChecked` is intended for uncontrolled components (where state is managed in the DOM rather than in React): https://reactjs.org/docs/uncontrolled-components.html

My suggestion would be that both the input's `defaultChecked` and `checked` props should be exposed. This could be either with the original names, or as `isChecked` and `isDefaultChecked`. If the latter, apps may need to do extra work when integrating the component with third-party form libraries.

It'd also be helpful if [the two Formik stories](https://github.com/Royal-Navy/design-system/blob/master/packages/react-component-library/src/components/Checkbox/Checkbox.stories.tsx) were both updated to show `initialValues` working correctly.

m7kvqbe1:
## Related issue

Closes #1398

## Overview

Explicitly expose `checked` and `defaultChecked` props and mark `isChecked` as deprecated.

## Reason

>Currently, there is an isChecked prop which maps to the input defaultChecked prop. However, the input's checked prop should be exposed.

## Work carried out

- [x] Explicitly expose new props
- [x] Mark old prop as deprecated
- [x] Add automated tests
- [x] Fix some Formik stories
- [x] Update docs-site documentation

### DIFF
--- a/packages/docs-site/src/library/pages/components/form/checkbox.md
+++ b/packages/docs-site/src/library/pages/components/form/checkbox.md
@@ -133,7 +133,21 @@ with the field:
     Description: 'Provide the ID for a field and tie it to its label. If one is not provided then a unique id will be created',
   },
   {
-    Name: 'isChecked',
+    Name: 'checked',
+    Type: 'boolean',
+    Required: 'False',
+    Default: 'false',
+    Description: 'Whether the checkbox is checked or not.',
+  },
+  {
+    Name: 'defaultChecked',
+    Type: 'boolean',
+    Required: 'False',
+    Default: 'false',
+    Description: 'Whether the checkbox is checked by default or not.',
+  },
+  {
+    Name: 'isChecked (Deprecated)',
     Type: 'boolean',
     Required: 'False',
     Default: 'false',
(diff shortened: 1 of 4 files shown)


jpveooys (approved):
:+1:

PR #1683, opened 2020-11-18
feat(Select): Convert to styled-components

### LINKED ISSUE #1663: Convert Select to styled-components

thyhjwb6:
## Related issue
Closes #1663 

## Overview
Converts the existing `Select` to styled-components.

## Reason
Migrating all components.

## Work carried out
- [x] Migrate existing CSS

## Developer notes
This has been thoroughly tested but would benefit from further testing as it was not a simple lift and shift.


packages/react-component-library/src/components/Select/Option.tsx
```diff
@@ -15,17 +15,45 @@ export interface SelectOptionWithBadgeType {
...
+    color: ${color('neutral', 'white')};
+  }
+
+  &&& {
+    padding: ${spacing('4')};
+    margin: 0;
+  }
+`
+
+const StyledBlah = styled.span`
```
m7kvqbe1: Should StyledBlah be called something else?

packages/react-component-library/src/components/Select/Select.tsx
```diff
@@ -20,13 +23,75 @@ export interface SelectProps
...
+    margin-top: -${spacing('px')};
+    border-top-left-radius: 0;
+    border-top-right-radius: 0;
+    box-shadow: 0 0 0 1px ${color('action', '600')},
+      -2px 2px 0 2px ${color('action', '100')},
+      2px 2px 0 2px ${color('action', '100')};
+    border: 1px solid ${color('action', '600')};
+    border-top: 0;
+    background: ${color('neutral', 'white')};
+    &::after {
```
m7kvqbe1, line 68: Could we add a line break here before the pseudo element block.

m7kvqbe1 (changes requested):
LGTM - just a couple of tiny comments.

m7kvqbe1 (approved):

PR #1686, opened 2020-11-19
feat(Pagination): Expose `initialPage` prop

### LINKED ISSUE #1617: Pagination component - unable to inject current page into the component
On refresh, there is no way to inject the current page into the component so it always starts at page one. This means that if the user is on page 5 and refreshes their browser, the pagination will show page 1 as highlighted even though we will show page 5 data.

thyhjwb6:
## Related issue
Fixes #1617 

## Overview
Exposes `initialPage` for selecting the active page when loading the component.

## Reason
Consumers currently have no way of selecting the initial page when the component is loaded.

## Work carried out
- [x] Expose `initialPage` prop


m7kvqbe1 (approved):
👍

Issue #1565, opened 2020-10-16
z-indexes


kixdcrwtkqgiyrxs:
- In general, z-indexes should be incremented by bigger jumps such as 100, this is so you shouldn't run into issues when you introduce a component needing to sit above and below existing components

- z-index in 'Popover' is incorrect, as overlapping still occurs from 'today marker' in timeline component

- changing z-index above could cause issues in timeline side panel (just forwarning)

- z-index of 'today marker' and dotted lines indicating a new month in timeline, as custom events will want 'today marker' to sit ontop of events, whilst dotted line should sit under them

- z-index on sidebar needs increasing, it is higher than elements in timeline (could be changed to ~9000)

PR #1696, opened 2020-11-23
fix(Timeline): Change z-indexes to fix stacking of Timeline elements

hxltrhuxze:
- Added Masthead, Timeline, and Sidebar global z-index tokens
- Added missing colon to the `StyledTimelineMonth` component z-index rule
- Scoped the Timeline to its own stacking context
- Bumped SidebarE's z-index to the new design token
- Bump the z-index of the Today marker to be higher than all other components in the Timeline

This should fix the immediate issues in #1565, however we will need a z-index review at some point to ensure we can prevent these bugs happening again.

### DIFF
--- a/packages/design-tokens/src/tokens/zindex.json
+++ b/packages/design-tokens/src/tokens/zindex.json
@@ -6,17 +6,26 @@
     "modal": {
       "value": "8000"
     },
-    "overlay": {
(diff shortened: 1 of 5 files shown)


m7kvqbe1 (approved):
LGTM 👍

PR #1743, opened 2020-12-02
feat(DesignTokens): Expose tagged literal `bp` breakpoint selector

### LINKED ISSUE #1705: Improve design-tokens breakpoint getter (tagged template literal)
https://github.com/Royal-Navy/design-system/issues/1602#issuecomment-726093501

m7kvqbe1:
## Related issue

Closes #1705

## Overview

Create and expose `bp` tagged template literal breakpoint selector from design tokens package.

## Reason

>Allows for the short and consistent creation of breakpoints within `styled-components`.

## Work carried out

- [x] Create and expose `bp` selector
- [x] Preserve `breakpoint` selector and add deprecation warning
- [x] Add automated tests
- [x] Update documentation
- [x] Setup Prettier and ESLint


packages/design-tokens/package.json
```diff
@@ -33,6 +34,7 @@
     "@types/lodash": "^4.14.161",
     "babel-loader": "^8.1.0",
     "clean-webpack-plugin": "^3.0.0",
+    "jest": "^26.6.3",
```
thyhjwb6: Do you think it might be worth syncing all `jest` dependencies across the packages?

packages/design-tokens/src/__tests__/selectors.test.tsx
```diff
@@ -0,0 +1,101 @@
+import React from 'react'
+import styled from 'styled-components'
+import { render, RenderResult } from '@testing-library/react'
+import '@testing-library/jest-dom/extend-expect'
+import 'jest-styled-components'
+
+import { selectors } from '..'
+
+const { bp } = selectors
```
thyhjwb6: I understand it was a tricky decision to find a name that does not clash with existing naming. I wonder if `breakpoint` was originally the right name as in isolation I don't understand it although within the context of CSS it makes sense.

Maybe that opens the opportunity for another name?

packages/design-tokens/package.json
```diff
@@ -35,6 +41,7 @@
     "@types/lodash": "^4.14.161",
     "babel-loader": "^8.1.0",
     "clean-webpack-plugin": "^3.0.0",
+    "eslint": "^7.14.0",
```
thyhjwb6: Could be worth syncing `eslint`

thyhjwb6 (commented):
Looks good, a couple of comments 👍

packages/design-tokens/src/__tests__/selectors.test.tsx
m7kvqbe1: I figured that `bp` is a nice shorthand - which is somewhat the objective of the selector.

When the next major is released we could preserve the 'breakpoint' selector but use the new implementation.

I'm not sure what else to do to avoid a breaking change - it would be nice to get this out into the wild before the next major.

Very much open to suggestions on naming. 

Maybe something like `@media()` is better?

m7kvqbe1: We have agreed upon `mediaQuery` and `mq` for selectors associated with this new getter. 

`breakpoint` will be marked as deprecated.

thyhjwb6 (approved):
Looks great 👍

PR #1744, opened 2020-12-03
refactor(Select): Leverage more of `components` from react-select

### LINKED ISSUE #1709: Refactor `Select` to leverage more of the concept of components from `react-select`
[`react-select` has the concept of `components`](https://react-select.com/components) which we are using but we can further integrate this which will reduce the complexity of the styled-components implementation.

thyhjwb6:
## Related issue
Closes #1709 

## Overview
Refactor existing code to further utilise `react-select` `components`. This change also makes further use of the `partials` pattern for styled-components.

## Reason
Existing code could be simplified and easier to understand with consistent patterns.

## Work carried out
- [x] Refactor to use react-select components
- [x] Refactor to use partials

### DIFF
--- a/packages/react-component-library/package.json
+++ b/packages/react-component-library/package.json
@@ -83,7 +83,7 @@
     "@types/node": "^14.0.26",
     "@types/react": "^16.9.19",
     "@types/react-dom": "^17.0.0",
-    "@types/react-select": "^3.0.10",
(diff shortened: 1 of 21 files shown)


m7kvqbe1 (approved):
LGTM 👍 

I think there might be some type changes after a rebase because of a dependency upgrade yesterday.

PR #1776, opened 2020-12-11
Feature/update lint config

antictor:
The current lint config does not support Typescript 4 and Create React App 4/Typescript based projects as it uses an out of date version of @typescript-eslint/parser.

This change updates the dependencies in this package to bring them up to date.

A few new rules are introduced that are not compatible with the React component library or report false positives in React so the rules have been updated to address those issues.

Finally one file reported a false positive for a new rule 'no-shadow'. This seems common at this point and found in other projects too.

### DIFF
--- a/packages/eslint-config-react/index.js
+++ b/packages/eslint-config-react/index.js
@@ -17,6 +17,7 @@ module.exports = {
   },
   plugins: ['@typescript-eslint', 'prettier', 'jest'],
   rules: {
+    'arrow-body-style': 0,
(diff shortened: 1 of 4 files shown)


m7kvqbe1 (changes requested):
Thank you for this - these changes look good to me.

Could you please reword your commit messages to include the following prefix: 

`feat(ESLintConfigReact): My commit message`

m7kvqbe1 (approved):
👍

PR #1777, opened 2020-12-14
feat(ContextMenu): Prevent padding without icons

### LINKED ISSUE #1756: ContextMenu - When no icons are given to the ContextMenuItem, we don't want padding
### Description
Currently when you want to have a list of ContextMenuItems with no icons passed into them, a padding for the icons is still applied.

We would like it so that when no icons have been applied to all the ContextMenuItems in a ContextMenu, then the padding for the icons does not exist or configurable via a prop.

m7kvqbe1:
## Related issue

Closes #1756

## Overview

Break out into partials and prevent padding of text when no icons are supplied.

## Reason

>Currently when you want to have a list of ContextMenuItems with no icons passed into them, a padding for the icons is still applied.

## Work carried out

- [x] Prevent padding without icons
- [x] Refactor to use partials pattern

## Screenshot

Screenshot 2020-12-14 at 09 30 01

### DIFF
--- a/packages/react-component-library/src/components/ContextMenu/ContextMenu.stories.tsx
+++ b/packages/react-component-library/src/components/ContextMenu/ContextMenu.stories.tsx
@@ -57,3 +57,44 @@ export const Default = () => {
 }
 
 Default.storyName = 'Default'
+
+export const NoIcons = () => {
+  const ref = useRef()
+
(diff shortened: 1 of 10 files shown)


thyhjwb6 (approved):
👍

PR #1791, opened 2020-12-15
feat(Popover): Add prop for altering close delay

### LINKED ISSUE #1759: Popover - add a delay prop so the delay on hiding and showing can be customised
### Description
We would like the ability to customise the delay in hiding and showing the popover.

thyhjwb6:
## Related issue
Closes #1759 

## Overview
Once the mouse leaves the target element then there is a delay before the `Popover` closes. This defaults to `1000` and can now be modified via the `closeDelay` prop.

## Reason
Downstream applications need more control over the delay.

## Work carried out
- [x] Expose prop


m7kvqbe1 (approved):
LGTM

PR #1792, opened 2020-12-16
refactor(Alert): Migrate to `styled-components`

### LINKED ISSUE #1790: Migrate `Alert` to styled-components

m7kvqbe1:
## Related issue

Closes #1790

## Overview

Migrate `Alert` to styled-components.

## Reason

Migrating all components.

## Developer notes

Discovered issue with `mq` selector when using nested interpolations: https://github.com/Royal-Navy/design-system/issues/1794


packages/react-component-library/src/components/Alert/Alert.tsx
```diff
@@ -24,15 +30,17 @@ const VARIANT_ICON_MAP = {
   ),
 }
 
+export type AlertVariant =
```
thyhjwb6: I do not think this is a bad name although it appears to be inconsistent with other naming where we have appended `Type` to the end.

thyhjwb6 (approved):
Looks good 👍 

Just a minor observation about some naming

PR #1795, opened 2020-12-18
Add hide and show events to context menu

antictor:
This change allows an app to be notified when a context menu is shown or hidden.

If the event handlers are optional so do not change existing behaviour.

In the case of MSP this feature is required so the associated element can have it's display state changed when the context menu is displayed.

### DIFF
--- a/packages/react-component-library/src/components/ContextMenu/ContextMenu.tsx
+++ b/packages/react-component-library/src/components/ContextMenu/ContextMenu.tsx
@@ -5,8 +5,22 @@ import { useClickMenu, ClickType } from '../../hooks/useClickMenu'
 import { StyledContextMenu } from './partials/StyledContextMenu'
 
 interface ContextMenuProps extends ComponentWithClass {
-  attachedToRef?: React.RefObject<HTMLElement>
+  /**
+   * attachedToRef: A ref object for the element associated with the context menu
+   */
+  attachedToRef: React.RefObject<HTMLElement>
+  /**
+   * clickType: Should the menu display when the user uses a left or right mouse click?
+   */
   clickType?: ClickType
+  /**
+   * onHide: Optional handler function to be called when the context menu is hidden
+   */
+  onHide?: () => void
+  /**
+   * onShow: Optional handler function to be called when the context menu is displayed
+   */
+  onShow?: () => void
 }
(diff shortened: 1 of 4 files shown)


packages/react-component-library/src/hooks/useClickMenu.ts
```diff
@@ -22,14 +34,17 @@ export const useClickMenu = (
     if (e.target === attachedToRef.current) {
       e.preventDefault()
       setOpen(true)
+      if (onShow) onShow()
       return
     }
 
     setOpen(false)
+    if (onHide) onHide()
```
m7kvqbe1, line 42: Maybe worth including an event object arg when invoking these callbacks?

packages/react-component-library/src/components/ContextMenu/ContextMenu.tsx
```diff
@@ -5,17 +5,38 @@ import { useClickMenu, ClickType } from '../../hooks/useClickMenu'
 import { StyledContextMenu } from './partials/StyledContextMenu'
 
 interface ContextMenuProps extends ComponentWithClass {
-  attachedToRef?: React.RefObject<HTMLElement>
+  /**
```
m7kvqbe1, line 8: Did you find this interface confusing without the comments? 

Would you like to see more of these comments elsewhere?

m7kvqbe1 (changes requested):
Thank you for the PR!

It looks like there is some red coming back from CI that need little bit of attention.

A couple of tests would also be nice (if you are strapped for time we can add these).

I've also left a couple of small comments.

packages/react-component-library/src/components/ContextMenu/ContextMenu.tsx
antictor: It's fine without them, but having comments across the board helps remove ambiguity and it's just a good thing to do :)

packages/react-component-library/src/hooks/useClickMenu.ts
antictor: Can do if you think it's worth it though the only thing of use is the target and you wouldn't use it

m7kvqbe1 (approved):
👍

PR #1807, opened 2021-01-04
refactor(Avatar): Migrate to `styled-components`

### LINKED ISSUE #1806: Migrat `Avatar` to styled-components

m7kvqbe1:
## Related issue

Closes #1806

## Overview

Migrate `Avatar` to styled-components.

## Reason

Migrating all components.


packages/react-component-library/src/components/Avatar/partials/StyledAvatar.tsx
```diff
@@ -0,0 +1,35 @@
...
+  font-weight: bold;
+  text-decoration: none;
+
+  &:hover {
+    background: ${color('action', '500')};
+    color: ${color('neutral', 'white')};
+    cursor: pointer;
+  }
+
+  ${({ $dark }) =>
```
thyhjwb6, line 29: I am wondering whether this pattern is consistent with other variants. ie should we have `$variant` and pass through `dark` which then uses a lookup?

Appreciate it is just one at this stage.

thyhjwb6 (approved):
Looks good 👍 

Minor comment around pattern but as this is not publicly exposed we can change it if the need arises.

packages/react-component-library/src/components/Avatar/partials/StyledAvatar.tsx
m7kvqbe1: >Appreciate it is just one at this stage.

I felt like this was the simplest implementation at this stage.

PR #1827, opened 2021-01-05
test(Popover): Fix broken test resulting in infinite loop

m7kvqbe1:
## Overview

Jest was getting stuck in an infinite loop due to the use of waitFor and fake timers.

Now invoking clearAllTimers in an afterEach.

## Reason

Screenshot 2021-01-05 at 16 57 49


thyhjwb6 (approved):
👍

PR #1854, opened 2021-01-12
Migrate Modal and Dialog to styled-components

### LINKED ISSUE #1844: Migrate `Dialog` to styled-components

thyhjwb6:
## Related issue
Closes #1844 

## Overview
Migrates the `Modal` and `Dialog` to styled-components.

## Reason
Migrating all components.

## Work carried out
- [x] Migrate `Modal`
- [x] Migrate `Dialog`
- [x] Fix `font-size` in `mq`

## Developer notes
Setting `font-size` in `mq` meant that component font sizes would be overwritten with a percentage of the base font size. A good example of this being an issue is a heading that uses a `mq`. More thought needs to be given to how we document responsive font sizes.


packages/design-tokens/src/getters.ts
```diff
@@ -74,7 +74,6 @@ export function getMediaQuery(options: {
     if (breakpointGTE && !breakpointLT) {
       return css`
         @media ${media} (min-width: ${breakpointGTE}) {
-          font-size: ${baseFontSize};
```
m7kvqbe1, line 77: I think there are actually some tests asserting for this that need updating.

It reminds me actually that they need to be run as part of CI. I'll add a note to the board.

m7kvqbe1 (changes requested):
LGTM - just left a comment about some tests in design-tokens package.

m7kvqbe1 (approved):

PR #1913, opened 2021-01-22
Add ContextMenu position

### LINKED ISSUE #1787: ContextMenu - ability to choose where it opens (e.g above/below)
### Description
Currently we use the context menu on a table, however if the user opens the menu on the bottom row of the table on the screen the context menu is cut off and cannot be used. We would like the ability as a prop to change the location of the context menu, whether that be above where the user has clicked or below.

thyhjwb6:
## Related issue
Closes #1787 

## Overview
Adds ability to show menu above or below mouse pointer through the user of a `position` prop.

## Reason
There are instances where a menu showing below is not visible on the screen.

## Work carried out
- [x] Refactor existing code
- [x] Add `position`
- [x] Update docs

## Screenshot
TODO

### DIFF
--- a/packages/docs-site/src/library/pages/components/context-menu.md
+++ b/packages/docs-site/src/library/pages/components/context-menu.md
@@ -85,6 +85,13 @@ Right clicking on this component will open the context menu in place of the defa
 ### ContextMenu Properties
 <DataTable data={[
   {
+    Name: 'attachedToRef',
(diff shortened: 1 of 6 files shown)


m7kvqbe1 (approved):
LGTM

In the future we could potentially add to this so that it dynamically positions it above or below depending on where it is relative to the edges of the viewport.

PR #1921, opened 2021-01-25
fix(Select): Tweak styling and add additional story

### LINKED ISSUE #1914: Select: Gap between menu and text input
### Steps to reproduce
Go to the Select component and open the drop down menu there is a gap that wasn't there in previous versions.

Version: 20.30.6 (Bug)
This is also viewable in the current storybook build
image

Version: 20.23.0 (Working)

image

m7kvqbe1:
## Related issue

Closes #1914

## Overview

Resolves a 1px positioning issue introduced as regression from styled-components works.

## Reason

>Select options container would sit 1px below the input. There was no visual regression state capturing the open state.

## Work carried out

- [x] Fix styling bug
- [x] Add 'Open by default' story for visual regression test

## Developer notes

We should review stories for components that have hidden display states to prevent visual regressions slipping through the net.

I have created a card for this.

### DIFF
--- a/packages/react-component-library/src/components/Select/Select.stories.tsx
+++ b/packages/react-component-library/src/components/Select/Select.stories.tsx
@@ -17,7 +17,7 @@ const options = [
   { value: 'vanilla', label: 'Vanilla', badge: 50 },
 ]
 
-const iconOptions = options.map(option => ({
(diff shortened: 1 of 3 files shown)


thyhjwb6 (approved):
👍

PR #1942, opened 2021-01-28
Add ability to spread arbitrary props on all components

### LINKED ISSUE #1834: Moving to styled components and stripping data-testid making it impossible to test
### Description
We write tests using Cypress and need properties or classes to be used as selectors to locate page elements. Moving to styled components and stripping the data-testid elements during the build means that we now struggle to write tests.

The most recent examples are the update to the masthead. Our tests need to confirm that a user is logged in and the correct avatar details are displayed but there is now way to do this. The other example is context menu, there is nothing in the rendered markup that lets is know if a context menu is shown or not.

Can we either stop stripping data-testid or allow testids to be passed in as properties.

thyhjwb6:
## Related issue
Closes #1834 

## Overview
Adds ability to drill props to all components.

## Reason
Useful for drill arbitrary props such as `data-testid`.

## Work carried out
- [x] Add to all components
- [x] Backfill tests

## Developer notes
❤️

### DIFF
--- a/packages/react-component-library/src/components/Alert/Alert.test.tsx
+++ b/packages/react-component-library/src/components/Alert/Alert.test.tsx
@@ -7,7 +7,7 @@ import { Alert, ALERT_VARIANT } from '.'
 describe('Alert', () => {
   let wrapper: RenderResult
 
-  describe('when only the description is specified', () => {
+  describe('when minimal props are provided', () => {
     beforeEach(() => {
       wrapper = render(<Alert>Description</Alert>)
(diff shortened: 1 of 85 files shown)


m7kvqbe1 (approved):
Nice changes - LGTM!

Stepped through every commit - not a single comment believe it or not!

Is it worth us squashing the 'test + spread' bits into a single commit post-review (`feat(ReactComponentLibrary): Enable arbitrary prop drilling`)?

PR #1980, opened 2021-02-05
Add ability to set TabSet tabs to full width

### LINKED ISSUE #1907: TabSet - Full width tab set
### Description
We currently override css classes associated with the TabSet component in order to achieve a scrollable two tab full width set.

This is our css overrides
```
.rn-tab-set {
  height: 100%;
}

.rn-tab-set__head {
  // fix for older browsers
  min-height: 64px;
}

.rn-tab-set__body {
  height: 100%;
  padding: 1.5rem 0 !important;
  overflow-y: auto;
}

.rn-tab-set__tab-item {
  width: 50%;
}

.rn-tab-set__tab {
  width: 100%;

  &.is-active {
    background-color: color('neutral', '400');
    border-color: color('neutral', '200');
    color: #fff !important;
  }
}
```

This custom css is to allow us to achieve two tab buttons that are both 50% width each and with a scrollable tab body. We have also overriden the color for the is-active however we do not expect this to be done to the design system too, however would like the .is-active class to exist beyond the styled components migration if possible so we can continue to do this ourselves.

The react code that we use is as follows:
```
<TabSet>
  <Tab title="title 1">
    <Our own custom component />
  </Tab>
  <Tab title="title 2">
    <Our own custom component />
  </Tab>
</TabSet>
```

Could a simple change or addition to the TabSet component be made that allows the number of Tabs within it to be 100%/x width and with a tab body that is scrollable please.

thyhjwb6:
## Related issue
Closes #1907 

## Overview
Adds capability to `TabSet`:
- use `isFullWidth` on `TabSet` so that tabs use full available width
- will overflow the y axis when the content is bigger than the parent

## Reason
This was [requested](https://github.com/Royal-Navy/design-system/issues/1907) by a downstream consumer.

## Work carried out
- [x] Add `isFullWidth`
- [x] Add scrolling on y axis 

## Screenshot
Screenshot 2021-02-08 at 15 06 38

## Developer notes
Some comments have been added to the original request.

### DIFF
--- a/packages/docs-site/src/library/pages/components/tab-set.md
+++ b/packages/docs-site/src/library/pages/components/tab-set.md
@@ -124,6 +124,13 @@ The TabSet (and companion Tab) component allows the user to switch between diffe
     Description: 'Called when a tab is clicked.',
   },
   {
+    Name: 'isFullWidth',
+    Type: 'boolean',
+    Required: 'False',
+    Default: 'False',
+    Description: 'Sets the TabSet and tabs to fill the full width',
+  },
+  {
     Name: 'isScrollable',
     Type: 'boolean',
     Required: 'False',
--- a/packages/react-component-library/src/components/TabSet/TabItem.tsx
+++ b/packages/react-component-library/src/components/TabSet/TabItem.tsx
@@ -10,36 +10,46 @@ interface TabItemProps {
   isActive: boolean
   onClick: () => void
   onKeyDown: (event: KeyboardEvent<HTMLLIElement>) => void
+  isFullWidth?: boolean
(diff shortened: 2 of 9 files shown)


m7kvqbe1 (approved):
LGTM 👍

m7kvqbe1 (changes requested):
Notice a slight visual regression in the scrollable version - there is a bottom border missing from the tabs:

https://www.chromatic.com/test?appId=5e25c277526d380020b5e418&id=601d7411292e030021a2bf00

Other than this looks great!

m7kvqbe1 (approved):

PR #2005, opened 2021-02-11
fix(TabNav): Add export to expose TabNav component

antictor:
TabNav is missing in the exports for react components so projects are unable to use it


m7kvqbe1 (approved):
👍

PR #2035, opened 2021-02-22
fix(RangeSlider): Drill missing `mode` prop

### LINKED ISSUE #2007: Some RangeSlider props have stopped working
Some props such as `mode` have stopped working in the RangeSlider. These were some of the props being passed through to the Slider component in the underlying library.

### Steps to reproduce

1. Go to http://localhost:6006/?path=/story/rangeslider-examples--multiple-handles in Storybook on `master` or tag `2.32.0`. This story has `mode` set to 2 which should ['keep the sliders from crossing and separated by a step'](https://docs.royalnavy.io/components/range-slider).

2. Drag the first slider handle to the right.

3. Observe that it doesn't stop at the second handle:
   
    range-slider-after

4. Check out tag `2.31.1`.

5. Visit the same story.

6. Observe it behaves as expected:

    range-slider-old

### Minimal reproduction

I assume the existing story is sufficient but let me know if you need any more information.

### Discussion

The reason seems to be this change which moved the `...rest` from `StyledSlider` to `StyledRangeSlider`:

https://github.com/Royal-Navy/design-system/commit/e42ad6fb35ac6945f005937f8420eb3ce509f74f#diff-8464b72d73293757619c47967dc01cf028acfa777a3f95a6834dddb4f0b778f8R159-L173

That change was to allow arbitrary props to be attached to the root DOM element, but the `...rest` was previously being used to forward some props [inherited](https://github.com/Royal-Navy/design-system/blob/d1e2caac0ab313feb2d428953c2e81e993283f5f/packages/react-component-library/src/components/RangeSlider/Slider.tsx#L23) from react-compound-slider's `SliderProps` to the `Slider` component.

m7kvqbe1:
## Related issue

Closes #2007

## Overview

This was a regression introduced by the recent arbitrary prop drilling change.

## Reason

>This was causing the multiple handle story to break.

## Work carried out

- [x] Explicitly drill missing `mode` prop

### DIFF
--- a/packages/react-component-library/src/components/RangeSlider/Slider.tsx
+++ b/packages/react-component-library/src/components/RangeSlider/Slider.tsx
@@ -130,6 +130,7 @@ export const RangeSlider: React.FC<RangeSliderProps> = ({
   hasPercentage,
   displayUnit = '',
   formatValue,
+  mode,
   ...rest
 }) => {
   const [sliderValues, setSliderValues] = useState(values)
@@ -169,6 +170,7 @@ export const RangeSlider: React.FC<RangeSliderProps> = ({
         disabled={isDisabled}
         values={values}
         step={step}
+        mode={mode}
         onChange={onChange}
         onUpdate={onUpdateHandler}
       >


jpveooys (approved):
But would be good to look at the `RangeSliderProps` type as it looks like it's inheriting a few other things from `SliderProps` (maybe it shouldn't be...)

m7kvqbe1:
> But would be good to look at the `RangeSliderProps` type as it looks like it's inheriting a few other things from `SliderProps` (maybe it shouldn't be...)

I had a quick look at this and it looks like we were already explicitly drilling everything apart from `mode`.

PR #2058, opened 2021-03-03
ci(GHA): Fix check commits job

### LINKED ISSUE #1710: CI scripts `check-fixup` and `commitlint` are giving false positives in GH
The scripts are giving false positives. They work locally so maybe there is an issue with the parameters being passed in. Would also be good if they did not fail silently.

- [ ] Fix `check-fixup`
- [ ] Fix `commitlint`
- [ ] Remove silent failure

elglup:
## Related issue

closes #1710 

## Overview

Fix check commits job which runs the commintlint & check-fixup scripts.

## Work carried out
Added new method to the check commits job to pass in the pull request URL to the commintlint & check-fixup scripts.  The previous method used was passing an empty string as we are no longer using the 'pull request' trigger and therefore unable to access the 'github.event.pull_request._links.html.href' context variable.

- [x] Edited build_and_test_feature.yml 
- [x] Edited build_and_test_master.yml

### DIFF
--- a/.github/workflows/build_and_test_feature.yml
+++ b/.github/workflows/build_and_test_feature.yml
@@ -65,14 +65,8 @@ jobs:
             path: '**/node_modules'
             key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock') }}
 
-        - name: Attach workspace
-          uses: actions/download-artifact@v2
-          with:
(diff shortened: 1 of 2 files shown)


thyhjwb6 (approved):
That'll do 😎

PR #2065, opened 2021-03-05
Fix Timeline presentation if range prop is provided

### LINKED ISSUE #2067: Set correct maxWidth if range prop is provided

thyhjwb6:
## Related issue
Fixes #2067 

## Overview
The default interval size of `1` was being used to calculate the `maxWidth` but if a `range` is set then it should use that.

## Reason
Downstream consumers expect six months to be wider.

## Work carried out
- [x] Use range rather than default 1

## Screenshot
Screenshot 2021-03-05 at 08 18 48


m7kvqbe1 (approved):
👍

PR #2066, opened 2021-03-05
fix(Timeline): Show all days if `endDate` supplied

### LINKED ISSUE #2061: Timeline endDate prop no longer working
### Steps to reproduce

1. Run Storybook locally and go to http://localhost:6006/?path=/story/timeline--bound-by-fixed-dates

The displayed timeline should end on 15 February as this is set as the `endDate`. Instead, it ends on 12 February.

This regressed in 2.37.0. It only happens when the end date is more than a month after the start date.

It should display the entire period from the start date to the end date, as in older versions, and also ignore `range` when `endDate` is set.

#### 2.36.3

image

#### 2.37.0 and master

image

thyhjwb6:
## Related issue
Fixes #2061 

## Overview
If an `endDate` is supplied then it uses that at the default zoom level.

## Reason
The default zoom level was only showing one month but it needs to show all days if the `endDate` is supplied.

## Work carried out
- [x] Use `endDate`

## Screenshot
Screenshot 2021-03-05 at 08 23 54

### DIFF
--- a/packages/react-component-library/src/components/Timeline/Timeline.tsx
+++ b/packages/react-component-library/src/components/Timeline/Timeline.tsx
@@ -1,6 +1,5 @@
 import React from 'react'
 import classNames from 'classnames'
-import { find } from 'lodash'
 
 import { ComponentWithClass } from '../../common/ComponentWithClass'
 import { TimelineProvider } from './context'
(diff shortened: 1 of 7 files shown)


m7kvqbe1 (approved):
Makes sense... I think. 

I need to spend some more time with this code at some point to refamiliarise.

👍

PR #2069, opened 2021-03-05
feat(Timeline): Add ability to customise row CSS

### LINKED ISSUE #1998: Timeline - Option to override the row styling (not just header)
### Description
On Timeline component: 
https://storybook.royalnavy.io/?path=/story/timeline--with-custom-event-content

We would like to see guidelines of the rows in the timeline, so a thin border around the rows essentially. We currently only have the ability to provide a function to renderRowHeader but would be great if there was a way we could put these guidelines onto the whole row somehow.

thyhjwb6:
## Related issue
Closes #1998 

## Overview
Adds the ability to customise the CSS of a `TimelineRow`, including header and content.

## Reason
Required by downstream applications as `Timeline` is a framework.

## Work carried out
- [x] Add customisation

## Screenshot
Screenshot 2021-03-11 at 13 07 51

### DIFF
--- a/packages/react-component-library/src/components/Timeline/Timeline.stories.tsx
+++ b/packages/react-component-library/src/components/Timeline/Timeline.stories.tsx
@@ -1,7 +1,12 @@
 import React from 'react'
+import { css, CSSProp } from 'styled-components'
 import { format } from 'date-fns'
 import { Meta } from '@storybook/react/types-6-0'
-import { ColorDanger500 } from '@royalnavy/design-tokens'
+import {
+  ColorDanger500,
+  ColorNeutral100,
+  ColorNeutral200,
+} from '@royalnavy/design-tokens'
 
 import {
   Timeline,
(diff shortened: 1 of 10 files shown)


packages/react-component-library/src/components/Timeline/Timeline.stories.tsx
```diff
@@ -414,6 +419,55 @@ export const WithCustomColumns = () => {
...
+      startDate={new Date(2020, 3, 1)}
+      today={new Date(2020, 3, 15)}
+    >
+      <TimelineTodayMarker />
+      <TimelineMonths />
+      <TimelineWeeks />
+      <TimelineDays />
+      <TimelineRows>
+        <TimelineRow
+          css={rowCss}
```
jpveooys, line 449: The general concept looks good to me :+1: 

Could you check if the root `css` prop on TimelineRow is working when the component library is used in another project? 

I gave it a go and the content and header ones were fine, but the row height was unchanged (using this same example).

m7kvqbe1 (commented):
Happy with the pattern but it probably needs documenting in the timeline framework docs.

Are there other places we need to apply this to now or just rows to start?

packages/react-component-library/src/components/Timeline/Timeline.stories.tsx
jpveooys: Just retested – this is fixed now :slightly_smiling_face:

thyhjwb6:
> probably needs documenting in the timeline framework docs

added docs in a `fixup`

> Are there other places we need to apply this to now or just rows to start?

There is a note on the board which I will convert to an issue now. Good to get this through first and then repeat the pattern 👍

m7kvqbe1 (approved):
💚

PR #2125, opened 2021-03-19
Fix setting of startDate in Timeline

### LINKED ISSUE #2117: Timeline startDate context not being updated
Timeline startDate context not being updated when the user click next or previous.
We use the context in our application to reset the horizontal scroll on the timeline. 

In our application we used the startDate to detect if the user has click on the next or previous. 
In previous version the startDate would update.
Could this be added back?

thyhjwb6:
## Related issue
Closes #2117 

## Overview
Fixes the setting of `startDate` and `endDate` if the buttons to navigate left or right are pressed.

## Reason
When navigating, a calculation happens with the scales but the `startDate` and `endDate` are not updated but this is being used by downstream applications.

## Work carried out
- [x] Refactor tests
- [x] Set field


m7kvqbe1 (approved):
LGTM

PR #2137, opened 2021-03-24
fix(Timeline): Fix width of day

### LINKED ISSUE #2100: Timeline unitWidth no longer precise
### Steps to reproduce

1. Render the following example:
    
    ```
    const ExampleTimeline = () => (
      <Timeline
        startDate={new Date(2020, 9, 5)}
        endDate={new Date(2021, 1, 1)}
        unitWidth={53}
      >
        <TimelineTodayMarker />
        <TimelineMonths />
        <TimelineWeeks />
        <TimelineDays />
        <TimelineRows>{}</TimelineRows>
      </Timeline>
    )
    ```

2. Inspect the width of a day in the timeline.
3. Observe that it is not exactly 53px as requested (but is slightly off). The actual width appears to vary depending on the dates displayed.

This is a regression as it was previously exact in version `2.36.3`. It causes problems with the positioning and sizing of custom elements that expect a consistent and fixed day width. (For example, our project uses a repeating background to show dividers between days, but these are no longer aligned correctly.)

### Minimal reproduction

Repro example: https://github.com/jpveooys/design-system-repros/blob/timeline-day-width/src/pages/Home/Home.tsx

#### 2.40.0

image

#### 2.36.3

image

thyhjwb6:
## Related issue
Fixes #2100 

## Overview
Fixes the day width with a failing test for London/Europe timezone.

## Reason
This was previously fixed but there was not a failing test around the timezone issue.

## Work carried out
- [x] Add failing test
- [x] Add fix

### DIFF
--- a/packages/react-component-library/src/components/Timeline/context/timeline_scales.ts
+++ b/packages/react-component-library/src/components/Timeline/context/timeline_scales.ts
@@ -4,7 +4,6 @@ import {
   addWeeks,
   addYears,
   differenceInDays,
-  differenceInHours,
 } from 'date-fns'
 import { find } from 'lodash'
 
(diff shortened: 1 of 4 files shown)


packages/react-component-library/src/components/Timeline/context/timeline_scales.ts
```diff
@@ -95,7 +94,8 @@ function mapScaleOption(
     ...rest
   }: ScaleConfigOptionType): TimelineScaleOption => {
     const to = addDays(calculateDate(startDate, intervalSize), -1)
-    const numberOfHours = differenceInHours(to, startDate) + HOURS_IN_DAY
+    const numberOfHours =
+      differenceInDays(to, startDate) * HOURS_IN_DAY + HOURS_IN_DAY
```
jpveooys: What about:
```suggestion
      (differenceInDays(to, startDate) + 1) * HOURS_IN_DAY
```

jpveooys (approved):
:+1:

PR #2138, opened 2021-03-25
fix(NumberInput): Make number unit unselectable

### LINKED ISSUE #2087: NumberInput: Unit is selectable and makes it difficult to edit value
When the NumberInput is used with a unit and label the unit is selectable as text. 

This makes it difficult to interact with and focus the input with a mouse. Same applies for the label.

m7kvqbe1:
## Related issue

Closes #2087

## Overview

Make unit unselectable and refactor to use partials pattern.

## Reason

>When the NumberInput is used with a unit and label the unit is selectable as text.
>
>This makes it difficult to interact with and focus the input with a mouse. Same applies for the label.

## Work carried out

- [x] Make unit element unselectable
- [x] Refactor to use partials pattern


thyhjwb6 (approved):
👍

PR #2139, opened 2021-03-26
Fix date picker arrow click to close

### LINKED ISSUE #2085: DatePicker: The arrow to the right of the input does not close the picker
When in an opened state clicking the arrow to the right of the input does not close the picker.

thyhjwb6:
## Related issue
Fixes #2085 

## Overview
Fixes the `DatePicker` so that it closes when the arrow is clicked.

## Reason
Arrow is being clicked and the `DatePicker` closes and then reopens immediately.

When clicking in the document the `DatePicker` should close except for when a date is selected and the input button is clicked. When the input button is clicked the `DatePicker` will close with a separate piece of code. Without this fix, the `DatePicker` is closed and then immediately the `focus` event fires which reopens it again.

## Work carried out
- [x] Move to `partials`
- [x] Use multiple refs when document is clicked


m7kvqbe1 (approved):
Nice fix!

PR #2151, opened 2021-03-30
Add the ability for users to type the date into `DatePicker`

### LINKED ISSUE #1904: DatePicker: The ability for users to type the date
### Description
Currently the date picker does not allow users to type a date in. It would be useful for users who have to select dates 2/3 years in advance which could be 24/36 clicks away at the moment. 

Would this be possible for power users?

thyhjwb6:
## Related issue
Closes #1904 

## Overview
Adds the ability to type the date into the `DatePicker`.

## Reason
Required by downstream applications.

## Work carried out
- [x] Tidy up existing code
- [x] Implement change

### DIFF
--- a/packages/react-component-library/src/components/DatePicker/DatePickerInput.tsx
+++ b/packages/react-component-library/src/components/DatePicker/DatePickerInput.tsx
@@ -1,0 +1,56 @@
+import React from 'react'
+import differenceInMinutes from 'date-fns/differenceInMinutes'
+import { format } from 'date-fns'
+
+import { ComponentWithClass } from '../../common/ComponentWithClass'
+import { StyledInput } from './partials/StyledInput'
+
+export interface DatePickerInputProps extends ComponentWithClass {
+  id: string
+  isDisabled: boolean
+  format: string
+  from: Date
+  onFocus: React.FocusEventHandler<HTMLInputElement>
+  to: Date
+}
+
+function transformDates(
+  startDate: Date,
+  endDate: Date,
+  datePickerFormat: string
+) {
+  if (startDate && endDate && differenceInMinutes(endDate, startDate) > 0) {
+    return `${format(startDate, datePickerFormat)} - ${format(
+      endDate,
+      datePickerFormat
+    )}`
+  }
+
+  if (startDate) {
+    return format(startDate, datePickerFormat)
+  }
+
+  return ''
+}
+
+export const DatePickerInput: React.FC<DatePickerInputProps> = ({
+  from,
+  to,
+  format: datePickerFormat,
+  isDisabled,
+  ...rest
+}) => {
+  return (
+    <StyledInput
+      aria-label="Choose date"
+      data-testid="datepicker-input"
+      disabled={isDisabled}
+      readOnly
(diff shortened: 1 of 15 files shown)


packages/react-component-library/src/components/DatePicker/DatePickerInput.tsx
```diff
@@ -0,0 +1,101 @@
+import React from 'react'
+import { isString } from 'lodash'
+import { addHours, isMatch, parse } from 'date-fns'
+import { DayPickerProps, ModifiersUtils } from 'react-day-picker'
+
+import { ComponentWithClass } from '../../common/ComponentWithClass'
+import { StyledInput } from './partials/StyledInput'
+import { useInputValue } from './useInputValue'
+
+const VALUE_PATTERN = /^[0-9]{1,2}\/[0-9]{1,2}\/[0-9]{4}$/
```
m7kvqbe1: Could we call this `INPUT_MASK` maybe instead of `VALUE_PATTERN`?

m7kvqbe1 (changes requested):
This is nice! 

Couple of things I noticed from a UX point of view: 

- Once you've typed a full valid date it doesn't update the picker until the input loses focus. Could it update it as soon as a full valid date is entered instead?

- If the picker is open and I type a valid date and then hit return, it will update the date and close the picker. Maybe the picker should stay open?

packages/react-component-library/src/components/DatePicker/DatePickerInput.tsx
k9dh3zij: - **Update picker as soon as full valid date is entered:** That would be ideal, if not too much dev effort. I think it would also partially address the next point.

- **Hitting return closes the picker:** The reason behind this behaviour is that, when using a keyboard and after having typed a date, the user shouldn't have to tab through the picker elements in order to move to the next field. Am I right that if we can implement the first point, we will address your concern here, as the picker will update automatically, before even pressing Return?

k9dh3zij:
> * Once you've typed a full valid date it doesn't update the picker until the input loses focus. Could it update it as soon as a full valid date is entered instead?
> * If the picker is open and I type a valid date and then hit return, it will update the date and close the picker. Maybe the picker should stay open?

- **Update picker as soon as full valid date is entered:** That would be ideal, if not too much dev effort. I think it would also partially address the next point.

- **Hitting return closes the picker:** The reason behind this behaviour is that, when using a keyboard and after having typed a date, the user shouldn't have to tab through the picker elements in order to move to the next field. Am I right that if we can implement the first point, we will address your concern here, as the picker will update automatically, before even pressing Return?

m7kvqbe1 (changes requested):
Definitely almost there! I just noticed a couple of things (sorry):

- When I focus the input value the picker disappears, could it stay open?



- I noticed on some of the other stories there seems to be a styling conflict between the placeholder and label when a picker has a null value. In this instance I guess the label should always be in the focussed state style?

Screenshot 2021-04-20 at 10 45 11

m7kvqbe1 (approved):
Awesome - this is lovely now!

PR #2270, opened 2021-04-22
fix(Select): Resolve input styling issue

### LINKED ISSUE #2261: Select styling is out when a value is selected - can't consistently repro - maybe override specificity issue
Screenshot 2021-04-21 at 11 03 17

m7kvqbe1:
## Related issue

Closes #2261

## Overview

Resolve a persistent styling issue with the Select component that isn't always reproducible. 

## Reason

>The select input was positioning the label, input and value incorrectly.

## Work carried out

- [x] Resolve styling issue


thyhjwb6 (approved):
👍

PR #2271, opened 2021-04-23
Fix pagination error hiding input

### LINKED ISSUE #2189: Pagination - Check page number error obscures page number input
With the new condensed Pagination. If you type a page number that does not exist into the input box and hit enter. A error popup is shown titled 'Check page number'. However for us, this popup covers the page number input and stops the user being able to change the page number, rendering the Pagination frozen.

It is to do with the top css applied to it. Have noticed on Storybook it seems to be working though, but on our application is does not, might be to do with the top css being calculated incorrectly?

thyhjwb6:
## Related issue
Fixes #2189 

## Overview
Pagination error hides the input in some circumstances. This change positions the error from the `bottom` rather than the `top`.

## Reason
Use cannot type into the input.

## Work carried out
- [x] Refactor existing code
- [x] Fix issue

## Screenshot
Screenshot 2021-04-23 at 15 26 08

### DIFF
--- a/packages/react-component-library/src/components/Pagination/Pagination.test.tsx
+++ b/packages/react-component-library/src/components/Pagination/Pagination.test.tsx
@@ -169,6 +169,13 @@ describe('Pagination', () => {
             expect(wrapper.getByText('Check page number')).toBeVisible()
           })
 
+          it('should display the error in the correct position', () => {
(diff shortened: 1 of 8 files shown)


m7kvqbe1 (approved):
👍

PR #2291, opened 2021-04-26
fix(Timeline): Increase threshold to display day

### LINKED ISSUE #2181: TimelineDay hides if dayWidth is under 29px
For our app we set the dayWidth is set to 22px. 
I've notice the timeline does not display the days if the dayWith is less than 29px.

Is there anyway we can get round this?

https://github.com/Royal-Navy/design-system/blob/cb0a96d92f90983ea3e10f00f2f7af6af04612c2/packages/react-component-library/src/components/Timeline/TimelineDay.tsx#L8

thyhjwb6:
## Related issue
Closes #2181 

## Overview
Increases the width threshold for displaying days.

## Reason
A downstream app is using `unitWidth` of `22` and they would like to see days.

## Work carried out
- [x] Add fix

## Screenshot
Screenshot 2021-04-26 at 12 19 17

### DIFF
--- a/packages/react-component-library/src/components/Timeline/TimelineDay.tsx
+++ b/packages/react-component-library/src/components/Timeline/TimelineDay.tsx
@@ -5,7 +5,7 @@ import { DATE_DAY_FORMAT } from './constants'
 import { StyledDay } from './partials/StyledDay'
 import { StyledDayTitle } from './partials/StyledDayTitle'
 
-const DAY_DISPLAY_THRESHOLD = 29
+const DAY_DISPLAY_THRESHOLD = 20
 
 interface TimelineDayProps {
   date: Date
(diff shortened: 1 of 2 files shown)


packages/react-component-library/src/components/Timeline/TimelineDay.tsx
```diff
@@ -5,7 +5,7 @@ import { DATE_DAY_FORMAT } from './constants'
 import { StyledDay } from './partials/StyledDay'
 import { StyledDayTitle } from './partials/StyledDayTitle'
 
-const DAY_DISPLAY_THRESHOLD = 29
+const DAY_DISPLAY_THRESHOLD = 20
```
c9qsm62x, line 8: 👍

m7kvqbe1 (approved):

PR #2332, opened 2021-05-05
Add `isFullWidth` to `Timeline` to present at full width

### LINKED ISSUE #2120: Timeline - Range prop introduces strange behaviour on different zoom levels
When using the Timeline with a range prop (example range={3}), the behaviour that is expected is that the amount of months shown on default zoom level is 3. Then on a zoom in and out, the amount of x (years, weeks etc) should be 3 times the normal amount.

Instead what happens is the width of the individual columns are 3 times as wide not the amount of x (years, weeks etc).

This has been shown to one of the DS devs.

thyhjwb6:
## Related issue
Closes #2120 

## Overview
With an `isFullWidth` prop the `Timeline` will expand to the full width of the viewport or parent.

## Reason
The `range` prop is being used to fill the full width with unexpected results.

## Work carried out
- [x] Refactor to use a `TimelineGrid`
- [x] Implement change

## Developer notes
- Full test coverage of window resizing is not in this PR. A [separate issue](https://github.com/Royal-Navy/design-system/issues/2335) has been created.
- The rendering of months needed to change as we need to know what is the biggest month that can be presented and then render all months the same. Because the window is being resized and unit sizes are constantly changing, it meant that some months were being rendered differently
- The initial load of the calendar feels magical as it is now [hidden in `useTimelineWidth`](https://github.com/Royal-Navy/design-system/pull/2332/files#diff-0bf12df71e7ff0b7533a25b059fce664d6fc508ac76b098d718252a02ae73b87R32). A thought was to abstract out a `initialLoad` `function` which is called in `TimelineGrid`.

### DIFF
--- a/packages/react-component-library/src/components/Timeline/TimelineGrid.tsx
+++ b/packages/react-component-library/src/components/Timeline/TimelineGrid.tsx
@@ -1,0 +1,99 @@
+import React from 'react'
+import classNames from 'classnames'
+
+import { ComponentWithClass } from '../../common/ComponentWithClass'
+import { extractChildren, timelineChildrenType } from './helpers/children'
+import { StyledHeader } from './partials/StyledHeader'
+import { StyledInner } from './partials/StyledInner'
+import { StyledTimeline } from './partials/StyledTimeline'
+import { TimelineDays } from './TimelineDays'
+import { TimelineHours } from './TimelineHours'
+import { TimelineMonths } from './TimelineMonths'
+import { TimelineRows, TimelineRowsProps } from './TimelineRows'
+import { TimelineSide } from './TimelineSide'
+import { TimelineTodayMarker } from './TimelineTodayMarker'
+import { TimelineToolbar } from './TimelineToolbar'
+import { TimelineWeekColumns } from './TimelineWeekColumns'
+import { TimelineWeeks } from './TimelineWeeks'
+
+export interface ComponentNameProps extends ComponentWithClass {
+  children: timelineChildrenType | timelineChildrenType[]
+  hasSide: boolean
+  hideScaling: boolean
+  hideToolbar: boolean
+}
+
+function getRenderColumns(
+  children: timelineChildrenType | timelineChildrenType[]
+) {
+  const rowsChildren = extractChildren(children, [TimelineRows])
+
+  return (rowsChildren[0] as React.ReactElement<TimelineRowsProps>).props
+    .renderColumns
+}
+
+export const TimelineGrid: React.FC<ComponentNameProps> = ({
+  children,
+  className,
+  hasSide,
+  hideScaling,
+  hideToolbar,
+  ...rest
+}) => {
+  const renderColumns = getRenderColumns(children)
+
+  const rootChildren = extractChildren(
+    children,
+    [
+      TimelineDays,
+      TimelineHours,
+      TimelineMonths,
+      TimelineRows,
+      TimelineSide,
+      TimelineTodayMarker,
+      TimelineWeeks,
+    ],
+    true
+  )
+
+  const headChildren = extractChildren(children, [
+    TimelineDays,
+    TimelineHours,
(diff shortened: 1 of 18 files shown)


packages/react-component-library/src/components/Timeline/hooks/useTimelineWidth.ts
```diff
@@ -0,0 +1,45 @@
...
+import { TIMELINE_ACTIONS } from '../context/types'
+import { TIMELINE_ROW_HEADER_WIDTH } from '../constants'
+
+export function useTimelineWidth(
+  hasSide: boolean,
+  isFullWidth: boolean
+): {
+  timelineRef: MutableRefObject<HTMLDivElement>
+} {
+  const timelineRef = useRef<HTMLDivElement>()
```
m7kvqbe1, line 14: Does it make sense to supply the ref to the hook instead of the hook itself having ownership over it? 

Just thinking about future hooks/functionality that possibly also need to work with a ref.

packages/react-component-library/src/components/Timeline/hooks/useTimelineWidth.ts
```diff
@@ -0,0 +1,45 @@
...
+  function dispatchWidthChange() {
+    dispatch({
+      type: TIMELINE_ACTIONS.CHANGE_WIDTH,
+      width: isFullWidth ? getWidth() : null,
+    })
+  }
+
+  /* eslint-disable consistent-return */
+  useEffect(() => {
+    dispatchWidthChange()
```
m7kvqbe1, line 32: It is arguably hidden away, but I think i'd expect the hook to initialise what it needs to implicitly anyway so I think it's probably ok imo.

packages/react-component-library/src/components/Timeline/context/reducer.ts
```diff
@@ -89,20 +92,60 @@ export function buildCalendar({
...
-      startDate: scale.from,
-      endDate: state.options.endDate ? scale.to : null,
-    },
+  switch (action.type) {
+    case TIMELINE_ACTIONS.CHANGE_WIDTH:
+      /* eslint-disable no-case-declarations */
+      const scaleOptions = initialiseScaleOptions(state.options, action.width)
+      const currentScaleIndex =
+        state.currentScaleIndex ||
+        findIndex(scaleOptions, ({ isDefault }) => isDefault)
```
m7kvqbe1: Is there a reason to use the lodash helper over the native implementation?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex

packages/react-component-library/src/components/Timeline/TimelineGrid.tsx
```diff
@@ -0,0 +1,103 @@
...
+  hideToolbar: boolean
+}
+
+function getRenderColumns(
+  children: timelineChildrenType | timelineChildrenType[]
+) {
+  const rowsChildren = extractChildren(children, [TimelineRows])
+
+  if (!rowsChildren.length) {
+    return null
```
m7kvqbe1: This line is flagged by coverage report is there a test or something that covers this no rows state?

m7kvqbe1 (changes requested):
🎉 

Looks great and feels really nice!

Have you tested it cross browser (mainly thinking about the window resize stuff that snapshots won't cover)?

m7kvqbe1:
Needs a rebase to make the snapshot generation work again. I had to revert a recent dependency upgrade.

packages/react-component-library/src/components/Timeline/hooks/useTimelineWidth.ts
thyhjwb6: After looking at some options we will leave this as-is but keep thinking about how we will share refs with consumers.

m7kvqbe1 (approved):
👍

PR #2361, opened 2021-05-14
fix(DatePicker): Do not style `--outside` days

### LINKED ISSUE #2253: DatePicker range selection styles blank days
Screenshot 2021-04-16 at 20 16 25

The empty boxes that represent days in the previous/next month should not include any additional styling.

m7kvqbe1:
## Related issue

Closes #2253

## Overview

Add missing qualifier to CSS selector to handle edge case.

## Reason

Outside days would sometimes be styled in a month block.

## Work carried out

- [x] Fix broken CSS selector

## Screenshot

Screenshot 2021-05-14 at 11 13 28
Screenshot 2021-05-14 at 11 12 44

## Developer notes

I'd add a test to cover this but it's tricky because we're overriding internal library styles using CSS selectors.


thyhjwb6 (approved):
👍

PR #2363, opened 2021-05-14
fix(SidebarE): Use explicit refs with `Transition`

### LINKED ISSUE #2334: findDOMNode was passed an instance of Transition which is inside StrictMode
### Steps to reproduce
1. Run your React app in StrictMode

### Minimal reproduction

Please provide a reference to a repository that includes a minimal reproduction of the issue.

We recommend spawning a new application using the [Royal Navy CRA Template](https://www.npmjs.com/package/cra-template-royalnavy):

```bash
// npx
npx create-react-app my-app --template royalnavy

// yarn create
yarn create react-app my-app --template royalnavy
```

### Screenshot
`Screenshot 2021-05-05 at 10 13 44

m7kvqbe1:
## Related issue

Closes #2334

## Overview

Stops a deprecation warning for `findDOMNode` usage being thrown in in React strict mode.

## Reason

>React would throw a deprecation warning for `findDOMNode` usage when run in strict mode.

## Work carried out

- [x] Refactor to use explicit refs with `Transition`

### DIFF
--- a/packages/react-component-library/jest/setupTests.js
+++ b/packages/react-component-library/jest/setupTests.js
@@ -6,7 +6,7 @@ import { format } from 'util'
 const originalConsoleError = global.console.error
 
 global.console.error = (...args) => {
-  const reactFailures = [/Failed prop type/, /Warning: /]
+  const reactFailures = [/Failed prop type/, /Warning: /, /StrictMode/]
 
   if (reactFailures.some((p) => p.test(args[0]))) {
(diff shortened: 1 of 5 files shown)


thyhjwb6 (commented):
I wonder if any of this would be testable. What prevents the `ref` from being removed in the future?

thyhjwb6 (approved):
👍

PR #2374, opened 2021-05-18
ci(GHA): Add notifications workflow

### LINKED ISSUE #2299: Create Awaiting Triage GitHub slack bot

elglup:
## Related issue

closes #2299

## Overview

Add notifications workflow

## Reason

Automate team tasks

## Work carried out

- [x] Created Slack app + webhook
- [x] Added GHA workflow to check for triage & stale issues then post slack message in team channel Mon-Fr am

### DIFF
--- a/.github/workflows/project_notify.yml
+++ b/.github/workflows/project_notify.yml
@@ -1,0 +1,43 @@
+name: Project Automation - notify
+
+on:
+  schedule:
+    - cron: '0 8 * * 1-5'
+
+jobs:
+     Notify:
+      runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


m7kvqbe1 (commented):
Does it make sense to move the bash into a file of it's own instead of it being inline in the YAML?

elglup:
> Does it make sense to move the bash into a file of it's own instead of it being inline in the YAML?

Not sure its worth it for such a small amount? Personally I like the workflow being self contained also.

m7kvqbe1 (approved):

PR #2379, opened 2021-05-21
fix(Modal): Set `max-height` and `overflow-y`

### LINKED ISSUE #2269: Dialog sizing bug
When users have a low screen resolution, or if dialogues contain a lot of content a dialogue's size can often exceed that of the available space. In this case the content is not visible, and things like the action buttons are off screen.

Ideally there would be a mechanism to mitigate this. An obvious candidate is an internal scroll bar on the dialogue.

There could also be a rule to prevent dialogues exceeding the size of the available screen space.

I have some images of this - I can't share them here but when this ticket is allocated please get in touch.

Thanks.

thyhjwb6:
## Related issue
Fixes #2269 

## Overview
Adds a `max-height` and `overflow-y: scroll` to `Modal`.

## Work carried out
- [x] Apply fix

### DIFF
--- a/packages/react-component-library/src/components/Modal/Modal.test.tsx
+++ b/packages/react-component-library/src/components/Modal/Modal.test.tsx
@@ -79,6 +79,17 @@ describe('Modal', () => {
         )
       })
 
+      it('should scroll on lower resolution screens', () => {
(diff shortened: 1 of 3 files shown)


m7kvqbe1 (approved):
👍

PR #2407, opened 2021-06-01
fix(Select): Allow clear from outside component

### LINKED ISSUE #2372: Unable to 'clear' Select input
### Steps to reproduce
1. Create a page with a select input with some options
2. Create a stateful variable to store the field value
3. Add a button that sets the value back to '' or undefined
4. Render the page and select a value from the select
5. Click the button to clear the field

The field value changes and inspecting the select component shows it's value is not undefined or '' but the field still displays the previously selected value

```
const [value, setValue] = useState<string | undefined>()
const onChange = (event: any) => setValue(event.target.value)
const onClick = () => setValue(undefined)

return (
  <>
    <div><Select options={options} value={value} onChange={onChange} label="test"/></div>
    <div><Button type="button onClick={onClick}/>Reset</Button></div>
  </>
}
```

thyhjwb6:
## Related issue
Closes #2372 

## Overview
Needs to handle the side effect of the value changing.

## Work carried out
- [x] Fix

### DIFF
--- a/packages/react-component-library/src/components/Select/Select.test.tsx
+++ b/packages/react-component-library/src/components/Select/Select.test.tsx
@@ -1,9 +1,15 @@
-import React from 'react'
+import React, { useState } from 'react'
 import '@testing-library/jest-dom/extend-expect'
-import { fireEvent, render, RenderResult } from '@testing-library/react'
+import {
+  fireEvent,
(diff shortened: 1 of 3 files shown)


m7kvqbe1 (approved):
LGTM

PR #2409, opened 2021-06-02
test(ReactComponentLibrary): Resolve `Jest` + `Cypress` clashing types

m7kvqbe1:
## Overview

You now need to explicitly import Cypress objects instead of relying on globals (Jest doesn't support this).

https://github.com/bahmutov/local-cypress

"The objects are still attached to the window object during the test, but the types are no longer global, you need to explicitly import them from this package."


m7kvqbe1:
Also fixes a flagged `yarn audit` issue.

thyhjwb6 (approved):
👍

PR #2416, opened 2021-06-04
Fix `DatePicker` showing error during range selection

### LINKED ISSUE #2300: DatePicker shows incorrect red border after clicking in text input
### Steps to reproduce

1. Go to https://storybook.royalnavy.io/?path=/docs/date-picker--default
2. Click in the date picker text input area (not on the down arrow)
3. Click on a date from the calendar-based picker below

A red border incorrectly appears around the control.

### Demo

date-picker-invalid

thyhjwb6:
## Related issue
Fixes #2300 

## Overview
Removes checking the date when the input blurs because this is not required when a range is selected as it was not keyed.

## Work carried out
- [x] Refactor existing test
- [x] Fix issue


m7kvqbe1 (approved):
👍

PR #2447, opened 2021-06-09
ci(Cypress): Add fix for missing binary issue

### LINKED ISSUE #2427: Cypress e2e job failing intermittently
https://github.com/Royal-Navy/design-system/runs/2768152949

```
The cypress npm package is installed, but the Cypress binary is missing.

We expected the binary to be installed here: /home/runner/.cache/Cypress/7.4.0/Cypress/Cypress

Reasons it may be missing:

- You're caching 'node_modules' but are not caching this path: /home/runner/.cache/Cypress
- You ran 'npm install' at an earlier build step but did not persist: /home/runner/.cache/Cypress

Properly caching the binary will fix this error and avoid downloading and unzipping Cypress.

Alternatively, you can run 'cypress install' to download the binary again.

https://on.cypress.io/not-installed-ci-error```

elglup:
## Related issue

closes #2427 

## Overview

Fix for Cypress e2e job failing intermittently

### DIFF
--- a/.github/workflows/build_and_test_feature.yml
+++ b/.github/workflows/build_and_test_feature.yml
@@ -23,7 +23,7 @@ jobs:
           uses: actions/cache@v2
           with:
             path: ~/.cache/Cypress
-            key: cypress-${{ runner.os }}-cypress2-${{ hashFiles('**/yarn.lock') }}
+            key: cypress-${{ runner.os }}-cypress-${{ hashFiles('**/yarn.lock') }}
             restore-keys: |
              cypress-${{ runner.os }}-cypress-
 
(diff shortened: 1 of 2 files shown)


m7kvqbe1 (approved):
nice! was it a typo?

elglup:
> nice! was it a typo?

No :>} I reset the cache key but that wasn't the root cause.
When the yarn.lock changes, occasionally it causes the yarn cache to be refreshed but missing the cypress binary (which is separate to the cypress npm package).  I couldn't find a way to avoid this happening and don't want to stop caching node modules,  so to get round this I have added a conditional statement that checks and installs the cypress binary if it is not cached. This should put an end to this pesky issue!

PR #2448, opened 2021-06-09
feat(Select): Add ability to remove clear button

### LINKED ISSUE #2386: Select - Clear button should be optional
### Description
https://storybook.royalnavy.io/?path=/docs/select--default

The 'x' button that clears the select should be optional, we would like a way to turn it off, as we currently have mandatory fields using Select which we do not wish to be cleared.

thyhjwb6:
## Related issue
Closes #2386 

## Overview
Allows `isClearable` to be passed as a parameter.

## Work carried out
- [x] Allow `isClearable`

## Screenshot
Screenshot 2021-06-10 at 09 35 21


m7kvqbe1 (approved):
LGTM 👍

PR #2461, opened 2021-06-15
fix(DatePicker): Close single picker on selection

### LINKED ISSUE #2424: Date picker won't get closed by itself on date selection
### Steps to reproduce
1. On a date picker after clicking a date for selecting it the date picker won't get closed by itself.
2. We need to click somewhere outside to get the date picker closed

It is not very clear it terms of user journey to select a date.

m7kvqbe1:
## Related issue

Closes #2424

## Overview

Hide the `DatePicker` on date selection (when not in `isRange` mode).

## Reason

>After some discussion this appears to be the commonly expected behaviour.

## Work carried out

- [x] Adjust automated tests
- [x] Modify behaviour


packages/react-component-library/src/components/DatePicker/DatePicker.tsx
```diff
@@ -185,6 +185,14 @@ export const DatePicker: React.FC<DatePickerProps> = ({
     }
   }
 
+  useEffect(() => {
```
thyhjwb6: When I have been touching state I have been thinking about if it's a good time to abstract to a hook which feels like a pattern that has been working well. In this situation I wonder if there's an opportunity for a hook `useSelection` which manages `state` and how that would affect this code. It appears to be all manageable at this stage but I wonder if we add something else and the code starts becoming harder to manage.

thyhjwb6 (approved):
nice 👍 

Observation about opportunity to refactor to a hook, maybe.

m7kvqbe1:
https://60cb1b909179a3b64f8815f6--pedantic-brahmagupta-8e0f34.netlify.app/

PR #2475, opened 2021-06-21
feat(DesignTokens): Enable styled-components theming

### LINKED ISSUE #2351: Add styled-components context / theme switching ability to React Component Library
Add styled-components context / theme switching ability to React Component Library

This is related to https://github.com/Royal-Navy/design-system/issues/2344

m7kvqbe1:
## Related issue

Closes #2351

## Overview

Add the ability to inject themed token sets and leverage `styled-theming`.

## Reason

>Add styled-components context / theme switching ability to React Component Library.

## Work carried out

- [x] Remove some dead code in token getters
- [x] Add ability to switch token sets
- [x] Add `styled-theming` pattern example
- [x] Document

## Screenshot

Screenshot 2021-06-21 at 13 23 50

## Developer notes

>By default the `GlobalStyleProvider` will use the `lightTheme` exported by the `@royalnavy/design-tokens` package. You can create your own themes by injecting your own custom token set, via the `theme` prop.
>
>We recommend also reading the following blog post on [`styled-theming`](https://jamie.build/styled-theming.html). Using this pattern you can selectively theme individual components. Inverting responsibility for the implementation of the theme to the component itself.
>
>When utilising this pattern remember to extend a base token set:
>
>```javascript
><GlobalStyleProvider theme={{ ...lightTheme, customThemeMode: 'foo' }} />
>```


packages/design-tokens/src/getters.ts
```diff
@@ -20,16 +13,24 @@ import {
   TypographySize,
   ZIndexGroup,
   StyledComponentsInterpolation,
+  Theme,
 } from './types'
 
-export function getBreakpoint(size: BreakpointSize): Breakpoint | undefined {
+function getTheme(theme?: Theme): Theme {
+  return theme?.coloursTokens ? theme : defaultTheme
```
thyhjwb6: It could be good to use `color` spelling although I appreciate we use `colour` in other places.

packages/design-tokens/src/getters.ts
```diff
@@ -20,16 +13,24 @@ import {
...
+}
+
+export function getBreakpoint(
+  size: BreakpointSize,
+  theme?: Theme
+): Breakpoint | undefined {
   const breakpoint = get(
```
thyhjwb6, line 29: I wonder how this could be written so that we are not going `getTheme(theme).breakpointsTokens` twice.. and also  `breakpoint[${size}].x.value` twice.

Along the lines of:

```
const { breakpointsTokens } = getTheme(theme)
```

packages/design-tokens/src/getters.ts
```diff
@@ -53,12 +57,12 @@ export function getMediaQuery(options: {
   }
 
   const breakpointGTE = get(
-    breakpointsTokens,
+    getTheme(theme).breakpointsTokens,
```
thyhjwb6: Same here really

thyhjwb6 (commented):
I think it looks good.. just wondering if we can tweak some syntax slightly.

I also wonder to what extent this is covered by tests.

packages/design-tokens/src/getters.ts
m7kvqbe1: I agree and the selector uses `color` but I have a feeling I couldn't easily change @hxltrhuxze original spelling in the tokens because of the css-framework. I'll investigate.

thyhjwb6 (changes requested):
I can set a lot of the functions in `getter.ts` to return `undefined` and the tests still pass so it feels like we could do with more covering this which I anticipate is evolving `selectors.test.ts`.

thyhjwb6 (approved):
👍

PR #2562, opened 2021-06-30
fix(Storybook): Add more specific actions

### LINKED ISSUE #2473: Fix Storybook passing in `onKeyDown` which affects performance
`onKeyDown` is being passed into `TextInputE` (and `TextInput` + potentially more) for Storybook examples when it has not been specified, giving the impression that the component is laggy. We need to figure out where it is coming from (possibly docs add on) and remove it.

thyhjwb6:
## Related issue
Fixes #2473 

## Overview
The example configuration was adding an `action` for each callback which was having performance issues when changing the `TextInput` and `TextInputE`

## Work carried out
- [x] Apply fix


m7kvqbe1 (approved):
Aha nice spot! Naughty wildcard!!

PR #2577, opened 2021-07-08
Ci/dependabot

elglup:
## Related issue

#2493

## Overview
Resolve permissions issue for Dependabot PR's which is preventing visual regression tests from running.


thyhjwb6 (approved):
👍

PR #2588, opened 2021-07-13
Fix items not showing on last day of `Timeline`

### LINKED ISSUE #2561: Timeline doesn't consider times after 00:00 on last day as visible
### Steps to reproduce

1. Run minimal reproduction, which renders a Timeline starting on `2021-01-01` with a range of one month
2. Observe that the today marker is set to show on `2021-01-31 12:00` but it isn't rendered
2. Observe that the event starting at `2021-01-31 06:00` in row two isn't rendered, but the event starting at `2021-01-31 00:00` in row one is rendered
3. Observe from the console output from `CustomComponent` that `startsAfterEnd` is true for `2021-01-31 12:00`, and false for `2021-01-31 00:00`

### Minimal reproduction

https://github.com/jpveooys/design-system-repros/tree/timeline-last-day
(See https://github.com/jpveooys/design-system-repros/blob/timeline-last-day/src/pages/Home/Home.tsx.)

### Screenshot

image

thyhjwb6:
## Related issue
Closes #2561 

## Overview
The last day is visible so if an item has a time on the last day then it should be shown.

## Work carried out
- [x] Tidy code
- [x] Apply fix

## Screenshot
Screenshot 2021-07-13 at 11 35 02

### DIFF
--- a/packages/react-component-library/src/components/Timeline/hooks/useTimelinePosition.ts
+++ b/packages/react-component-library/src/components/Timeline/hooks/useTimelinePosition.ts
@@ -1,6 +1,6 @@
 import { useContext } from 'react'
 
-import { differenceInCalendarDays, isBefore, isAfter } from 'date-fns'
+import { differenceInCalendarDays, isAfter, isBefore } from 'date-fns'
 
 import { TimelineContext } from '../context'
 import { formatPx } from '../helpers'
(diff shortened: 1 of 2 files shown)


packages/react-component-library/src/components/Timeline/hooks/useTimelinePosition.ts
```diff
@@ -37,16 +42,19 @@ export function useTimelinePosition(
...
-  const lastDateDisplayed = days[days.length - 1].date
+  const { from: firstDateDisplayed, to: lastDateDisplayed } = currentScaleOption
 
-  const startsBeforeStart = isBefore(new Date(startDate), firstDateDisplayed)
-  const startsAfterEnd = isAfter(new Date(startDate), lastDateDisplayed)
-  const endsBeforeStart = isBefore(new Date(endDate), firstDateDisplayed)
-  const endsAfterEnd = isAfter(new Date(endDate), lastDateDisplayed)
+  const startsBeforeStart = isBefore(startDate, firstDateDisplayed)
+  const startsAfterEnd =
+    isAfter(startDate, lastDateDisplayed) &&
```
jpveooys:
```tsx
  const startsAfterEnd = isAfter(startDate, endOfDay(lastDateDisplayed))
  // ...
  const endsAfterEnd = isAfter(endDate, endOfDay(lastDateDisplayed))

```
might be another option.

jpveooys (approved):
Looks good to me

m7kvqbe1 (approved):
LGTM

PR #2590, opened 2021-07-14
fix(Helpers): Handle non existent `window` object gracefully

m7kvqbe1:
## Overview

Handle non-existent `window` object gracefully.

## Reason

Sometimes the `window` object is non-existent in downstream builds - for example NextJS static site generation.


thyhjwb6 (approved):
👍

PR #2696, opened 2021-10-14
fix(DatePicker): Allow update from side effect

### LINKED ISSUE #2631: DatePicker startDate value does not change when updated externally
### Steps to reproduce
1. Consume the `DatePicker` component from the `@royalnavy/react-component-library` package.
2. Pass in the `startDate` value.
3. Change the `startDate` value externally (normally this is due to wanting to open an edit form and pre-populate the components with existing data).
4. Notice that the `DatePicker` value does not change.
5. Set the `key` property on the `DatePicker` component to the date value -> it will now update as expected.

### Minimal reproduction

This has been reproduced in [CodeSandbox](https://codesandbox.io/s/peiwf).

### Expected outcome

The `startDate` value of the `DatePicker` component should be updated when the value is changed externally.

### Actual outcome

The `startDate` value of the `DatePicker` component does not change, unless using the `key` workaround.

thyhjwb6:
## Related issue
Closes #2631 

## Overview
Updates `startDate` and `endDate` stored in state when updated externally.

## Reason
`startDate` and `endDate` could be updated by a side effect.

> DatePicker startDate value does not change when updated externally

## Work carried out
- [x] Handle side effect

### DIFF
--- a/packages/react-component-library/src/components/DatePicker/DatePicker.test.tsx
+++ b/packages/react-component-library/src/components/DatePicker/DatePicker.test.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useState } from 'react'
 import '@testing-library/jest-dom/extend-expect'
 import { ColorDanger600, ColorSuccess600 } from '@defencedigital/design-tokens'
 import {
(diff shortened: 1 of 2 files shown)


m7kvqbe1 (approved):
🎉

Issue #2454, opened 2021-06-11
React hook ESLint rules aren't running


jpveooys:
The `react-hooks/rules-of-hooks` and `react-hooks/exhaustive-deps` ESLint rules aren't running in our app. I've created this as a bug as I believe they were running at some point (but I could be wrong so feel free to change this to a feature request).

The reason they aren't running appears to be that [`airbnb/hooks`](https://github.com/airbnb/javascript/tree/master/packages/eslint-config-airbnb#eslint-config-airbnbhooks) is missing from `extends` in `packages/eslint-config-react/index.js`.

### Steps to reproduce

1. Run `yarn lint` in `packages/react-component-library`
2. Observe there are 0 errors
3. Add `airbnb/hooks` to `extends` in `.eslintrc.js`
4. Run `yarn lint` in `packages/react-component-library` again
5. Observe there are now ~10 errors

thyhjwb6:
After investigation this is a larger piece of work and is a feature request so should be a lower priority than other bugs. For now it is possible to work around this by extending the configuration in the application.

PR #2702, opened 2021-10-15
Fix `Timeline` start date update from side effect

### LINKED ISSUE #2470: Timeline component updates  if the props StartDate is changed
### Description
At the moment the startDate has has an effect when it is set on the initial render. Would it be possible that the timeline updates when the startDate props has changed?

thyhjwb6:
## Related issue
Closes #2470 

## Overview
It is possible that `startDate` is updated as a side effect and therefore the `Timeline` needs to update. This change dispatches an event which is handled by the reducer.

## Work carried out
- [x] Tidy up existing code
- [x] Handle `startDate`

### DIFF
--- a/packages/react-component-library/src/components/Timeline/context/index.tsx
+++ b/packages/react-component-library/src/components/Timeline/context/index.tsx
@@ -1,8 +1,13 @@
-import React, { createContext, useReducer } from 'react'
+import React, { createContext, useEffect, useReducer } from 'react'
 
 import { initialState } from './state'
 import { reducer } from './reducer'
-import { TimelineContextDefault, TimelineProviderProps } from './types'
+import {
+  TIMELINE_ACTIONS,
+  TimelineContextDefault,
+  TimelineProviderProps,
+} from './types'
+import { initialiseScaleOptions } from './timeline_scales'
 
 const timelineContextDefaults: TimelineContextDefault = {
   hasSide: false,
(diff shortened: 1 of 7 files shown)


jpveooys (commented):
Looks OK to me :+1: 

But, it does something to the 'By fixed dates' story when navigating back a month:

fixed-dates

(It starts navigating back by longer and longer periods of times...)

Navigating back and forward doesn't make too much sense when using fixed dates but might be worth checking why the behaviour's changed (especially as it is in Storybook).

thyhjwb6:
> But, it does something to the 'By fixed dates' story when navigating back a month:

Nice spot! Fix and supporting test added

packages/react-component-library/src/components/Timeline/context/index.tsx
```diff
@@ -24,6 +29,23 @@ export const TimelineProvider: React.FC<TimelineProviderProps> = ({
...
+      {
+        ...options,
+        endDate: state.getNewEndDate(),
+        hoursBlockSize: state.currentScaleOption.hoursBlockSize,
+      },
+      state.width
+    )
+
+    dispatch({ scaleOptions, type: TIMELINE_ACTIONS.CHANGE_START_DATE })
+  }, [options.startDate])
```
jpveooys, line 47: Strictly speaking, the `useEffect()` hook call is missing dependencies on `options` and `state`. This will get picked up once #2454 is in.

I think most likely we'd need to use [`usePrevious`](https://reactjs.org/docs/hooks-faq.html#how-to-get-the-previous-props-or-state) to check if the startDate has actually changed (as with the other hook dependencies in you won't know which one has changed). But happy to leave that to #2454 as there are probably other instances of the same thing about.

jpveooys (approved):

packages/react-component-library/src/components/Timeline/context/index.tsx
thyhjwb6: 💯  having investigated #2454 it is a larger piece of work which we should look at holistically.

PR #2717, opened 2021-10-19
feat(Forms): Add `Formik` and `react-hook-form` examples

### LINKED ISSUE #2488: Add e2e example using `react-hook-form` and `formik` with all new v3 components
Begin demonstrating the use of the new components with the added benefit of us developing an understanding of how the components fit together.

- [x] Create an example
  - [x] Include `TextInputE`
  - [x] Include `TextAreaE`
  - [x] Include `RadioE`
  - [x] Include `CheckboxE`
- [x] Add Cypress test

We can continue developing the examples as more components become available.

### LINKED ISSUE #2740: CheckboxE and RadioE components don't function correctly in Firefox
### Steps to reproduce

1. Go to https://storybook.design-system.digital.mod.uk/?path=/docs/checkboxe--default
2. Try to click on the checkbox
3. Go to https://storybook.design-system.digital.mod.uk/?path=/docs/radioe--default#formik-group
4. Try to click on one of the radio buttons

Most of the time the click doesn't toggle state, though occasionally it does.

#### CheckboxE

checkboxe

#### RadioE

radioe

m7kvqbe1:
## Related issue

Closes #2488 

Closes https://github.com/defencedigital/mod-uk-design-system/issues/2740

## Overview

Create and test working examples of `3.x.x` forms utilising a multitude of different companion libraries.

## Reason

>We need to ensure that form components are interoperable with a multitude of different downstream libraries and use cases. We also need to ensure that the discrete components themselves look and behave correctly when combined together into a web form.

## Work carried out

- [x] Ensure that refs are forwarded to underlying HTML inputs
- [x] Include `Formik` example in Storybook
- [x] Include `react-hook-form` example in Storybook 
- [x] Create suite of e2e Cypress tests
- [x] Add some missing return types to helpers
- [x] Resolve style bug causing Firefox `CheckboxE` and `RadioE` to be unselectable

## Screenshot

Screenshot 2021-10-25 at 10 41 25
Screenshot 2021-10-25 at 10 41 46


packages/react-component-library/cypress/specs/forms/index.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+              before(() => {
+                cy.get(selectors.form.submit).click()
+              })
+
+              it('should set the submit Button state to disabled', () => {
+                cy.get(selectors.form.submit).should('have.attr', 'disabled')
+              })
+
+              it('should supply form the field values', () => {
+                if (name === 'Formik') {
```
thyhjwb6: I wonder if we could make this more generic and add the expected values to the `examples` array? This would avoid changes needed to be made in two places if an example is added or removed.

packages/react-component-library/src/helpers.ts
```diff
@@ -49,7 +49,7 @@ function withKey(
   element: React.ReactElement,
   prefix: string,
   suffix: string | number
-) {
+): React.ReactElement | null {
```
thyhjwb6: Because `ReactElement` is nullable I don't think the `null` is required.

thyhjwb6 (commented):
Looks good, I really like running the same tests against two different form implementations.

jpveooys:
Are you getting some errors like this when running `yarn install`/`yarn build` on this branch?

```
$ tsc
cypress/helpers.ts:3:26 - error TS7006: Parameter 'color' implicitly has an 'any' type.

3 export function hexToRgb(color) {
                           ~~~~~

cypress/plugins/index.ts:15:26 - error TS7006: Parameter 'pn' implicitly has an 'any' type.

15 export default function (pn, config) {
                            ~~

```

Could be related to a dependency update somewhere...

m7kvqbe1:
> Are you getting some errors like this when running `yarn install`/`yarn build` on this branch?
> 
> ```
> $ tsc
> cypress/helpers.ts:3:26 - error TS7006: Parameter 'color' implicitly has an 'any' type.
> 
> 3 export function hexToRgb(color) {
>                            ~~~~~
> 
> cypress/plugins/index.ts:15:26 - error TS7006: Parameter 'pn' implicitly has an 'any' type.
> 
> 15 export default function (pn, config) {
>                             ~~
> ```
> 
> Could be related to a dependency update somewhere...

~~These typings have now been added (https://github.com/defencedigital/mod-uk-design-system/pull/2717/commits/80f1a7ef7723362d19cc904bcceb10acc13afe3c) - they were being flagged because the cypress directory has been added to the `.tsconfig` include array as part of this PR.~~

EDIT: Looking at this further - i'm not sure why I originally added cypress directory to the `.tsconfig` include array. Typing it all seems to be more trouble than it's worth at this stage - so i've removed it again.

jpveooys:
@m7kvqbe1 

I'm still getting some type errors unfortunately (a few are in `src/components`) – entire list is here: https://gist.github.com/jpveooys/6ae4c9ffd51cdc67232cd09dd447993f

Edit: Update the list as of `012d8bc`

jpveooys:
Is it expected that, for React Hook Form, the initial value of exampleCheckbox is `false`, but it changes to `[]` after toggling a checkbox twice?

react-hook-form

**Edit**: Looks like the fix is to set `defaultValues` when calling `useForm()` (similar to Formik's `initialValues`, so that the initial value of `exampleCheckbox` is `[]`): https://react-hook-form.com/api/useform

(It does also say 'It is encouraged that you set defaultValues for all inputs to non-undefined values such as the empty string or null.')

packages/react-component-library/package.json
```diff
@@ -179,12 +179,14 @@
     "autoprefixer": "^10.2.5",
     "classnames": "^2.2.6",
     "date-fns": "^2.9.0",
-    "formik": "^2.1.2",
+    "formik": "^2.2.9",
     "polished": "^4.0.3",
     "postcss-flexbugs-fixes": "^5.0.2",
     "react-compound-slider": "^3.3.1",
     "react-day-picker": "^7.4.8",
+    "react-hook-form": "^7.17.5",
```
jpveooys: Should this be a dev dependency, or possibly [an optional peer dependency](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#peerdependenciesmeta)?

jpveooys (commented):
Looks good overall :+1: Just those couple of questions/comments.

Perhaps we could create another issue to add some examples where the form is prepopulated with some data (i.e. the editing case)?

jpveooys (approved):
Nice one :+1:

m7kvqbe1:
Downgrading to `6.15.4` because IE11 support has been dropped in `7.x.x`.

😭

thyhjwb6 (commented):
👍 

It can be useful to leave squashing `fixup` commits until it is ready to be merged as it's difficult to see what has changed without reviewing the code again.

PR #2720, opened 2021-10-19
Timeline Performance gains 

c9qsm62x:
## Overview

In applications where there is a lot of manipulation to the timeline with many events the performance degrades. 

I first notice that the timeline rerenders the whole tree if the parent component rerenders. 
See the profiling screenshot. 

Screenshot 2021-10-19 at 21 20 27
This takes on average 200ms to rerender on a 2.2ghz  6 core i7  MacBook Pro  which is decent but does take significantly longer on our low powered devices (seconds) 

You man also notice that we applied some memoization components to our app which also did not help. 
The reason why is because the context is always a new object on every rerender. So any component that use the `TimelineContext` will rerender because the context state had change. 

With the changes made in this PR and the same optimisations applied in the previous test the render time dropped to <9 ms 
image

## Reason

[Why did you do what you did? - feel free to copy from the ticket if the reason is there]

## Work carried out

[A list of work you have done, use markdown checklist format, if you leave any boxes unchecked, be sure to set this PR as a WIP]

**Example.**

- [x] A bit of work I completed
- [ ] A bit of work I did not complete

## Screenshot

[If the work is UI related then paste a screenshot of the update here.]

## Developer notes

[Sometimes, extra notes are needed to add clarity to a PR, add them here]

### DIFF
--- a/packages/react-component-library/src/components/Timeline/__tests__/Timeline.test.tsx
+++ b/packages/react-component-library/src/components/Timeline/__tests__/Timeline.test.tsx
@@ -24,6 +24,7 @@ import {
   TimelineHours,
   TimelineMonths,
   TimelineRow,
+  TimelineRowProps,
   TimelineRows,
   TimelineSide,
   TimelineTodayMarker,
@@ -63,7 +64,7 @@ describe('Timeline', () => {
           className="test-class-name"
         >
           <TimelineMonths />
-          <TimelineRows>{}</TimelineRows>
+          <TimelineRows>{ }</TimelineRows>
         </Timeline>
       )
     })
@@ -82,7 +83,7 @@ describe('Timeline', () => {
           <TimelineWeeks />
           <TimelineDays />
           <TimelineHours />
-          <TimelineRows>{}</TimelineRows>
+          <TimelineRows>{ }</TimelineRows>
         </Timeline>
       )
 
@@ -373,7 +374,7 @@ describe('Timeline', () => {
             <TimelineMonths />
             <TimelineWeeks />
             <TimelineDays />
-            <TimelineRows>{}</TimelineRows>
+            <TimelineRows>{ }</TimelineRows>
           </Timeline>
         )
       })
@@ -508,7 +509,7 @@ describe('Timeline', () => {
           <TimelineMonths />
           <TimelineWeeks />
           <TimelineDays />
-          <TimelineRows>{}</TimelineRows>
+          <TimelineRows>{ }</TimelineRows>
         </Timeline>
       )
     })
@@ -803,7 +804,7 @@ describe('Timeline', () => {
               />
             )}
           />
-          <TimelineRows>{}</TimelineRows>
+          <TimelineRows>{ }</TimelineRows>
         </Timeline>
       )
     })
@@ -872,7 +873,7 @@ describe('Timeline', () => {
               />
             )}
           />
-          <TimelineRows>{}</TimelineRows>
+          <TimelineRows>{ }</TimelineRows>
         </Timeline>
       )
     })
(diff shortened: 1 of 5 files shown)


packages/react-component-library/src/components/Timeline/context/index.tsx
```diff
@@ -1,4 +1,6 @@
-import React, { createContext, useEffect, useReducer } from 'react'
+
```
jpveooys: Looks like a blank line has crept in here.

jpveooys (commented):
Looks good to me!

Unfortunately, most of the CI builds aren't running. It might be because you aren't a member and the PR is from a fork, and `build_and_test_feature.yml` isn't set to run on `pull_request` events.

@elglup Is this something you can help with?

thyhjwb6 (commented):
Nice change 🔥 

Is there an easy way to cover this with a test?

c9qsm62x:
> Is there an easy way to cover this with a test?

hmm I'm not sure if there is a elegant way to test that a rerender was skipped using react-testing-library 
I would argue that a regression test is enough because the functionality is identical.

thyhjwb6:
> hmm I'm not sure if there is a elegant way to test that a rerender was skipped using react-testing-library I would argue that a regression test is enough because the functionality is identical.

I wonder if there's a way to use the custom `render` which can then call a spy and that is asserted? Not looking to make this difficult but would be good to get the lines covered.

packages/react-component-library/src/components/Timeline/__tests__/Timeline.test.tsx
```diff
@@ -2336,4 +2337,68 @@ describe('Timeline', () => {
...
+      wrapper = render(
+        <TimelineWithUpdate>
+          <TimeLineRows />
+        </TimelineWithUpdate>
+      )
+    })
+
+    it('should not rerender children if props have not changed', async () => {
+      expect(wrapper.getByText('Render: 1')).toBeInTheDocument()
+      wrapper.getByText('Force update').click()
```
jpveooys, line 2400: Minor – something like:

```tsx
      expect(EventSpy).toBeCalledTimes(1)
      EventSpy.mockClear()
      
      wrapper.getByText('Force update').click()
      expect(await wrapper.findByText('Render: 2')).toBeInTheDocument()
      expect(EventSpy).not.toHaveBeenCalled()
```

might be a bit more explicit and clearer.

packages/react-component-library/src/components/Timeline/__tests__/Timeline.test.tsx
```diff
@@ -2336,4 +2337,68 @@ describe('Timeline', () => {
...
+              <TimelineDays />
+              <TimelineRows>
+                {children}
+              </TimelineRows>
+            </Timeline>
+          </>
+        )
+      }
+
+      const TimeLineRows = () => (
```
jpveooys: Minor:
```suggestion
      const TimelineRows = () => (
```

jpveooys (approved):
Nice work :+1: Putting aside the CI problems (that still need resolving), it looks good to me.

packages/react-component-library/src/components/Timeline/__tests__/Timeline.test.tsx
c9qsm62x: Just realised the name clash with the  design systems `TimelineRows` will rename to `TimelineRowItems`

thyhjwb6 (approved):
Awesome 👍 

Like the suggestion from @jpveooys 

(squashing the commits could make sense)

m7kvqbe1:
I'm closing this PR in favour of https://github.com/defencedigital/mod-uk-design-system/pull/2739. 

This was done to get CI running.

PR #2733, opened 2021-10-22
Add ButtonE component

### LINKED ISSUE #2440: Copy existing `Button` to `ButtonE`
> Buttons are clickable components that allow users to trigger actions. The button label signifies the action that will occur when the user interacts with that button.

- [ ] Copy component (sizes: forms, small)
- [ ] Apply CSS changes
  - [ ] Spacing
  - [ ] With icons
  - [ ] Primary
    - [ ] Default
    - [ ] Active
    - [ ] Pressed
    - [ ] Disabled 
  - [ ] Secondary
    - [ ] Default
    - [ ] Active
    - [ ] Pressed
    - [ ] Disabled 
  - [ ] Tertiary
  - [ ] Danger
    - [ ] Default
    - [ ] Active
    - [ ] Pressed
    - [ ] Disabled 

[Design spec](https://docs.google.com/document/d/1yXF0wUUvvsH88j9bn-PGc9VD3zr75tAC78uuXfFCoNs/edit#heading=h.cjtl4mn5pziy)

jpveooys:
## Related issue

Closes #2440

## Overview

This adds a new ButtonE component, based on the existing Button component but updated to be in line with the forms v3 spec.

## Link to preview

https://buttone.netlify.app/?path=/docs/button-experimental--default

## Reason

To implement forms v3 side-by-side with existing v2 components.

## Work carried out

- [x] Fork the Button component into a new ButtonE component
- [x] Update the styling of ButtonE
- [x] Add a new isLoading prop and feature to ButtonE

## Developer notes

Few bits about the component to note:

- Danger is now a variant instead of a colour
- The number of available button sizes is reduce to two (forms and small)
- Icon-only buttons are now properly implemented
- Tertiary buttons don't have a disabled state; this is enforce using types

The first commit is a straight copy of the existing component with renaming etc. and the following commits are the changes that have been made to the component, so you may find it easier to review the commits individually.


packages/react-component-library/src/components/ButtonE/ButtonE.stories.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+export default {
+  component: ButtonE,
+  title: 'Button (Experimental)',
+  parameters: { actions: { argTypesRegex: '^on.*' } },
+} as Meta
+
+const Template: Story<ButtonEProps> = (args) => <ButtonE {...args} />
+
+export const Default = Template.bind({})
```
jpveooys, line 20: I've used this pattern to simplify the stories a fair bit: https://storybook.js.org/docs/react/writing-stories/args#story-args

packages/react-component-library/src/components/ButtonE/ButtonE.tsx
```diff
@@ -0,0 +1,150 @@
...
+  /**
+   * Type of component to display (style varies accordingly).
+   */
+  variant?:
+    | typeof BUTTON_E_VARIANT.PRIMARY
+    | typeof BUTTON_E_VARIANT.SECONDARY
+    | typeof BUTTON_E_VARIANT.DANGER
+}
+
+interface NonDisableableButtonEProps extends BaseButtonEProps {
```
jpveooys: This is where I've split the types to stop `isDisabled` and `isLoading` from being set to true for tertiary buttons.

Would be interested if anyone has any views about this.

m7kvqbe1: Looks good to me - my only criticism is that within the generated Storybook docs it's not obvious the resulting restricted behaviour is intentional.

packages/react-component-library/src/components/ButtonE/ButtonE.stories.tsx
m7kvqbe1: Nice!

m7kvqbe1 (commented):
Looks good!

It would be good to update the about to be merged Forms stories to use this new experimental implementation.

packages/react-component-library/src/components/ButtonE/ButtonE.tsx
jpveooys: Yep, it's not too clear there. I had a look for a way to disable or restrict the `isDisabled` and `isLoading` controls if `tertiary` is selected but I didn't find a way to do that.

But the docs could be updated to say they aren't supported for tertiary buttons at least.

m7kvqbe1:
Probably also worth waiting to get eyes on from @k9dh3zij to review from a design perspective.

packages/react-component-library/src/components/ButtonE/ButtonE.tsx
thyhjwb6: You may find it useful to look at examples in the codebase where we have used unified types with `never`.

packages/react-component-library/src/components/ButtonE/ButtonE.tsx
```diff
@@ -55,40 +46,65 @@ export interface ButtonEProps extends ComponentWithClass {
...
   onClick,
-  size = BUTTON_E_SIZE.REGULAR,
+  size = BUTTON_E_SIZE.FORMS,
+  title,
   type = 'button',
-  variant,
+  variant = BUTTON_E_VARIANT.PRIMARY,
   ...rest
 }) => {
+  if (!children && !title && icon) {
```
thyhjwb6: We can avoid this by using the unified types and `never`.

packages/react-component-library/src/components/ButtonE/constants.ts
```diff
@@ -1,17 +1,12 @@
 const BUTTON_E_SIZE = {
   SMALL: 'small',
-  REGULAR: 'regular',
-  LARGE: 'large',
-  XLARGE: 'xlarge',
+  FORMS: 'forms',
```
thyhjwb6, line 3: I am wondering if we need a conversation about the variant names to use as we would want it to be consistent.

Is `forms` actually regular? Then for `small` we previously used an `isCondensed` flag on `NumberInput`. Happy either way but it needs to be consistent.

thyhjwb6 (commented):
Looks good 👍 

Main feedback is around use of unified types and `never`.

Will be good to get @k9dh3zij to give this the once over too. We usually deploy a preview to Netlify.

packages/react-component-library/src/components/ButtonE/ButtonE.tsx
jpveooys: @thyhjwb6 Not too sure what difference you're suggesting here?

packages/react-component-library/src/components/ButtonE/ButtonE.tsx
jpveooys: Could add more types. but I went for this as we won't be able to check for empty strings using types.

packages/react-component-library/src/components/ButtonE/constants.ts
jpveooys: Yep we should make sure we keep it consistent for the different components, but also with the design docs.

The design spec seems to use forms size and small size for the most part – so could have a chat with @k9dh3zij as well.

jpveooys:
@thyhjwb6 I've reworked the types as suggested. Note that it can generate confusing type errors so it may not be a net improvement for library consumers:

image

m7kvqbe1 (approved):

k9dh3zij:
Thank you @jpveooys, this is brilliant! 🤩 

Only one note from me:  

- The Tertiary style should be underlined by default and have a tab focus outline (just like the "Your feedback" link works in this one: https://5e25c277526d380020b5e418-yrusslegbj.chromatic.com/?path=/docs/phase-banner--custom-link )

image

Thank you!

packages/react-component-library/src/components/ButtonE/constants.ts
k9dh3zij: There may be small inconsistencies in the docs as we battled back and forth a few times with size names, but I'll review with our content editor. We previously settled on 'forms size' and 'small size', but happy to consider alternatives if they make more sense. 

We should be consistent with what we call these in code and in the docs, where possible, so maybe one to have a quick meeting about before we start getting inconsistent?

jpveooys:
@k9dh3zij Thanks for the review! I've pushed a fix for the tertiary underline bug (crept in when I was tidying up :grimacing: )

On the tertiary button focus outline, the one for the feedback link is actually coming from the browser (it looks different in different browsers). In that case it's the browser styles for a hyperlink, but in this case it's a button. Shall we have a quick chat about it when you're free?

jpveooys:
@k9dh3zij I've pushed updates based on the updated tertiary button spec and also added more stories for icons. Updated link: https://buttone.netlify.app/?path=/docs/button-experimental--default

(Have a look at all the buttons if you can, in case I've inadvertently changed something on one of the other ones...)

thyhjwb6 (approved):
@jpveooys 

> I've reworked the types as suggested

I get a friendlier message and based on consistency with existing patterns it makes sense.

Screenshot 2021-10-28 at 15 37 03

The interfaces look neat 👍

jpveooys:
@thyhjwb6 

> I get a friendlier message

Yep the types are simpler now as tertiary buttons now have disabled and loading states, so there are only two interfaces (was four before).

k9dh3zij:
@jpveooys  Thank you, tertiary button looks great! 

One thing I noticed (on all variants):

When clicking a button (especially if clicking a few times in quick succession), the shadow seems to change size and the style change effect doesn't feel 'snappy' (see video)

I think this is caused by an animation? Can we have no animation effects for transitions to and from the `pressed` state please? That's all from me, everything else is great!

https://user-images.githubusercontent.com/73531035/139288325-e5748de9-bd5f-4947-bcb6-189d422e2346.mov

jpveooys:
@k9dh3zij Good spot on the shadow glitch!

The lag actually seems to be coming from Storybook (at least here). Let me create a simple site outside of Storybook and we can review on that tomorrow :slightly_smiling_face:

thyhjwb6:
@jpveooys 

> The lag actually seems to be coming from Storybook

I tested this too and created a separate ticket #2760

jpveooys:
@k9dh3zij 

Here's a demo site outside of Storybook with the shadow glitch fixed (transitions are still on):

https://red-playground.netlify.app/

Have a look and see if that's better – if not we can have a chat about it later :slightly_smiling_face:

k9dh3zij:
Thanks so much @jpveooys 

Shadow glitch is gone and happy with the animations reduced to 75ms. 👍

As discussed:

1. If I tab through to a button and then press Enter, focus disappears. It should not. We should also remove the relevant test please.

2. In that test environment, all buttons seem to be using the default system font rather than Lato.

Many thanks!

jpveooys:
Thanks @k9dh3zij !

I've just done 1. – this is updated on Storybook and the test site.

On 2., I've just checked it was a problem with my test site (it had another font set on the body). (Should also be correct on the test site now.)

jpveooys:
@m7kvqbe1 Do you want to have another look at this? There have been a few changes since you reviewed it (mainly to the tertiary button, which now has the disabled and loading states and a different hover and click appearance).

k9dh3zij:
All good from my side, thank you very much @jpveooys 👍 🙏

thyhjwb6 (approved):
👍

Issue #1128, opened 2020-06-17
Add support for floating point numbers to `NumberInputE`


thyhjwb6:
The `NumberInputE` does not currently support floating point numbers.

If `step` is `0.1` then it will not work as expected.

If `1.5` is entered as the value then it will not work as expected.

PR #2734, opened 2021-10-22
Add `NumberInputE` v3 component

### LINKED ISSUE #2434: Copy existing `NumberInput` to `NumberInputE`
> Number inputs let users enter new or change existing numbers, either by editing the number directly or using the input controls.

- [x] Copy component
- [x] Apply CSS changes
  - [x] Update prefix
  - [x] Update suffix
  - [x] Update buttons

[Design spec](https://docs.google.com/document/d/1yXF0wUUvvsH88j9bn-PGc9VD3zr75tAC78uuXfFCoNs/edit#heading=h.9dpux7wjjbeb)

thyhjwb6:
## Related issue
Closes #2434 

## Overview
Adds the `NumberInputE` v3 component.

## Link to preview
https://numberinpute.netlify.app/?path=/story/number-input-experimental--default

## Reason
Preparing components for v3.

## Work carried out
- [x] Add component
- [x] Add condensed styles
- [x] Update component to match spec

## Developer notes
- thoughts on `isCondensed` vs `small`?
- #2748 has been created for adding this to the test sandbox


packages/react-component-library/src/components/NumberInputE/Button.tsx
```diff
@@ -0,0 +1,43 @@
...
+  onClick: (event: React.MouseEvent<HTMLButtonElement>) => void
+  type: ButtonType
+}
+
+const svgLookup = {
+  [NUMBER_INPUT_BUTTON_TYPE.DECREASE]: <IconRemove />,
+  [NUMBER_INPUT_BUTTON_TYPE.INCREASE]: <IconAdd />,
+}
+
+export const Button: React.FC<ButtonProps> = ({
```
jpveooys: Would be nice to use a more specific name for this as it's easy to confuse it for the main `Button` component when it's being used (in other files).

Some ideas:

- `NumberInputButton`
- `InlineButton`

packages/react-component-library/src/components/NumberInputE/Button.tsx
```diff
@@ -0,0 +1,43 @@
...
+  | typeof NUMBER_INPUT_BUTTON_TYPE.INCREASE
+
+interface ButtonProps {
+  isCondensed: boolean
+  isDisabled: boolean
+  onClick: (event: React.MouseEvent<HTMLButtonElement>) => void
+  type: ButtonType
+}
+
+const svgLookup = {
```
jpveooys: `svg` seems like a not too relevant implementation detail of the icons here. How about `iconLookup`?

```suggestion
const iconLookup = {
```

packages/react-component-library/src/components/NumberInputE/NumberInputE.stories.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+import {
+  NumberInputE,
+  NumberInputProps,
+  NumberInputWithIconProps,
+  NumberInputWithPrefixProps,
+  NumberInputWithSuffixProps,
+} from './NumberInputE'
+
+const chromaticIgnore = { chromatic: { disable: true } }
```
jpveooys: Looks like canvas isn't used now – is this still needed?

packages/react-component-library/src/components/NumberInputE/NumberInputE.test.tsx
```diff
@@ -664,4 +552,55 @@ describe('NumberInput', () => {
...
+        <NumberInputE
+          isCondensed
+          label="Label"
+          name="number-input"
+          onChange={onChangeSpy}
+        />
+      )
+    })
+
+    it('should display the label', () => {
```
jpveooys: Minor: Not sure we need to follow `it` by `should` all the time – could be `    it('displays the label', () => {`.

packages/react-component-library/src/components/NumberInputE/NumberInputE.test.tsx
```diff
@@ -0,0 +1,606 @@
...
+      beforeEach(() => {
+        wrapper.getByTestId('number-input-increase').click()
+      })
+
+      it('should hide the label', () => {
+        expect(wrapper.getByTestId('number-input-label')).not.toBeVisible()
+      })
+    })
+
+    describe('when', () => {
```
jpveooys: Looks like text for `describe()` is incomplete here.

packages/react-component-library/src/components/NumberInputE/partials/StyledButton.tsx
```diff
@@ -1,32 +1,66 @@
...
-  width: 42px;
+  flex-grow: 1;
+  justify-content: center;
+  height: 28px;
+  width: 28px;
   align-items: center;
 
   background: transparent;
-  margin: 0;
+  margin: 8px;
```
jpveooys: Use `spacing('4')`?

packages/react-component-library/src/components/NumberInputE/partials/StyledButton.tsx
```diff
@@ -1,32 +1,66 @@
...
+      margin: 6px;
+      border-radius: 3px;
+
+      & > svg {
+        height: 12px;
+        width: 12px;
+      }
+    `}
+
+  ${({ $isDisabled }) =>
```
jpveooys: Should be able to use the disabled pseudo-class here to avoid needing to set both `$isDisabled` and `disabled` on `StyledButton`:

```
&:disabled {
  ...
}
```

(Alternatively, `disabled` is automatically made available so  `${({ disabled }) =>` should also work.)

packages/react-component-library/src/components/TextInputE/partials/StyledInput.tsx
```diff
@@ -16,23 +16,36 @@ function removeAutoFillBackground() {
...
   box-sizing: border-box;
   width: 100%;
   margin: 0;
   border: 0;
   background: none;
-  padding: 22px 12px 6px 11px;
+
+  padding: ${({ $hasLabel, $isCondensed }) => {
+    if ($isCondensed || !$hasLabel) {
+      return '6px 12px 6px 11px'
```
jpveooys: Should we be using `spacing()` here?

packages/react-component-library/src/components/TextInputE/partials/StyledLabel.tsx
```diff
@@ -6,16 +6,32 @@ import {
...
 
 export const StyledLabel = styled(StyledLabelBase)<StyledLabelProps>`
-  transform: translate(11px, ${getYPosition()}) scale(1);
+  ${({ $isCondensed }) => css`
+    transform: translate(11px, ${getYPosition($isCondensed)}) scale(1);
+  `}
+
+  ${({ $hasContent, $hasFocus, $isCondensed }) => {
+    if (!$hasContent && !$hasFocus) {
+      return {}
```
jpveooys: `return null` should work here (`return {}` looks a bit out of place).

jpveooys (commented):
Nice work :+1:

I might need to do another pass as there's a fair bit here but wanted to submit these comments.

> thoughts on isCondensed vs small?

On 'small' vs 'condensed', small is something I would look for, and if I saw it I'd know what it meant. I don't think I'd look for condensed when thinking about size, and I'm not sure I'd know exactly what to expect if I saw `isCondensed` (it more has connotations of 'squished' to me rather than 'small').

On another topic, the design spec has this:

> As an option, you can specify that a Number input should also accept and display decimal numbers (-0.1, 0.25, 0.333).
> 
> In this case, ensure you have also defined an appropriate step.

Do we have this functionality when typing (not using the steppers)?

packages/react-component-library/src/components/NumberInputE/NumberInputE.stories.tsx
thyhjwb6: Really well spotted!

thyhjwb6:
Thank you for the feedback @jpveooys:
- introduced `COMPONENT_SIZE` and `ComponentSizeType` for size
- the current `NumberInput` does not support decimal numbers. There is a ticket which will be done as an incremental change #1128

k9dh3zij:
Number inputs, yay! 😁  Thank you @thyhjwb6 , here is my feedback:

1. Corner radius looks smaller than design. Please use same values as TextInputE (was this fixed as I was writing this comment?! 😮)

2. Prefix and suffix font colour should be #3e5667 (Neutral 400)

3. Tab with keyboard to plus or minus button > press Enter > focus moves elsewhere, which shouldn't happen.

4. When the field gets/loses focus, the fade animation appears to happen in 'two steps'. I assume that's because the animation effects/timings are slightly different between border and shadow. Can we make sure they are the same please? Let's go for whichever is the shortest one, as they're already on the slow side. (I think this may also be the case for TextInputE - if fixing it here doesn't fix it everywhere, we should raise bugs for all other instances.)

5. Border and shadow design for the `active` state of `small` (isCondensed) size should be different to that of `forms` design. Please see spec under TextField `small` size

image

packages/react-component-library/src/components/NumberInputE/partials/StyledInlineButtons.tsx
```diff
@@ -0,0 +1,19 @@
...
+  $isDisabled: boolean
+}
+
+export const StyledInlineButtons = styled.div<StyledInlineButtonsProps>`
+  position: relative;
+  display: flex;
+  justify-content: center;
+  text-align: center;
+
+  &:disabled {
```
jpveooys: This one should still be using `$isDisabled` I think (no disabled state on a `div`).

thyhjwb6: Oops! Didn't meant to leave this in 😂

packages/react-component-library/src/components/NumberInputE/Input.tsx
```diff
@@ -0,0 +1,62 @@
...
+  size,
+  value,
+  ...rest
+}) => {
+  const hasLabel = !!(label && label.length)
+
+  return (
+    <StyledInputWrapper>
+      {hasLabel && (
+        <StyledLabel
```
m7kvqbe1, line 39: Not for now - all for the code re-use but maybe we should abstract some of these bits out into primitives?

packages/react-component-library/src/components/NumberInputE/NumberInputE.stories.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+export default {
+  component: NumberInputE,
+  title: 'Number Input (Experimental)',
+  parameters: {
+    actions: { argTypesRegex: '^on.*' },
+  },
+} as Meta
+
+export const Default: Story<NumberInputProps> = (props) => (
```
m7kvqbe1: Have you seen the `Template.bind` pattern in @jpveooys `ButtonE` PR (https://github.com/defencedigital/mod-uk-design-system/pull/2733)?

packages/react-component-library/src/components/NumberInputE/NumberInputE.stories.tsx
```diff
@@ -0,0 +1,99 @@
+import React from 'react'
```
m7kvqbe1, line 1: Is it worth including react-hook-form and Formik usage examples in individual component story files or is this superfluous?

packages/react-component-library/src/components/NumberInputE/NumberInputE.test.tsx
```diff
@@ -0,0 +1,606 @@
...
+  waitFor,
+} from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
+
+import { Button, COMPONENT_SIZE } from '../..'
+import { NumberInputE } from '.'
+
+describe('NumberInputE', () => {
+  let wrapper: RenderResult
+  let onChangeSpy: (event: any) => void
```
m7kvqbe1: Sneaky `any` here that could be gotten rid of.

packages/react-component-library/src/components/NumberInputE/NumberInputE.tsx
```diff
@@ -0,0 +1,278 @@
...
+   */
+  name: string
+  /**
+   * Optional handler called when the `onBlur` event is emitted.
+   */
+  onBlur?: (event: React.FormEvent) => void
+  /**
+   * Handler called when the value selected by the component changes.
+   */
+  onChange: (event: any) => void
```
m7kvqbe1: `any`

packages/react-component-library/src/components/TextInputE/partials/StyledInput.tsx
m7kvqbe1: Also is there a constant for `small`?

m7kvqbe1 (changes requested):
Looks awesome!!! Just minor comments really.

One UX issue I noticed is when trying to increment / decrement using keyboard controls hitting return on on of the buttons has the unwanted effect of tabbing you to the next element:

packages/react-component-library/src/components/NumberInputE/Input.tsx
thyhjwb6: Yeah it seems to be evolving that way but would be too much for this PR.

packages/react-component-library/src/components/NumberInputE/NumberInputE.stories.tsx
thyhjwb6: I think that we have our "sandbox" examples of how to put the components into a form and individual component stories we could keep clean.

packages/react-component-library/src/components/TextInputE/partials/StyledInput.tsx
thyhjwb6: @jpveooys @m7kvqbe1:
- there was a `small` constant added in latest fixups
- `6px 12px 6px 11px` has been chosen as precision is required at px level, hence `22px 12px 6px 11px` in original code for the `forms` version

thyhjwb6:
@k9dh3zij responses:
1. 👍 
2. 👍 
3. As agreed #2759 
4. This appears to be an issue with Storybook as I created a separate app with the component and it looks fine (see screenshot). As this rings a bell, a separate issue has been raised #2760



5. 👍 

https://numberinpute.netlify.app/

jpveooys:
@thyhjwb6 Couple of failing tests on the 'Build & Test Feature / Test_react-component-library (pull_request)' build if you haven't seen them already

jpveooys (approved):

k9dh3zij:
Thanks for the fixes @thyhjwb6. 🙂
New feedback:

6. Component is now missing the light blue shadow when active (also not visible in the gif you shared above)

7. The user input (the `1000` in the example below) appears 1px lower than the middle of the input box. For reference, the `£` prefix is aligned correctly. Tested on Chrome and Safari on Mac.

image

 8. Not sure if this is relevant for here or if we need to change 'centrally', but `Footnote` style should be as follows:
 
image
Font: Lato Regular
Colour: #3e5667 (Neutral 400)
Size: 12px

Thank you!

thyhjwb6:
@k9dh3zij 

8. gif updated 😂 
9. 👍 
10. 👍 

https://numberinpute.netlify.app/

k9dh3zij:
Thanks for updating the gif @thyhjwb6 😂
All styles look great, thank you 👍

Last one from me, sorry I wrote it on a post-it but forgot to include it earlier:

Typing an alpha character is not accepted - that's great.
However, if the cursor is not at the end of the number when I type an alpha character, all digits to the right of the cursor are deleted. It shouldn't delete any digits.

https://user-images.githubusercontent.com/73531035/139445642-3cf94bb3-0294-4d91-a068-cddc95abde73.mov

thyhjwb6:
@k9dh3zij well spotted! As this is an issue with the existing `NumberInput` too, my recommendation is that we create a separate issue that is treated as a priority. The benefit is that we can get this piece of work through and the issue is then a smaller piece in isolation.

Update: Added #2767

m7kvqbe1 (approved):
LGTM 🎉

PR #2754, opened 2021-10-27
Add `SwitchE` component

### LINKED ISSUE #2435: Copy existing `Switch` to `SwitchE`
> Switch components allow users to choose one out of a maximum of five options.

- [x] Copy component (sizes: forms, small) 
- [x] Change to use JSX composition based API
- [x] Apply CSS changes

[Design spec](https://docs.google.com/document/d/1yXF0wUUvvsH88j9bn-PGc9VD3zr75tAC78uuXfFCoNs/edit#heading=h.cps8kldu6d64)

m7kvqbe1:
## Related issue

Closes #2435

## Overview

Create new experimental SwitchE component based on design spec.

## Link to preview

https://61815c8f62fa950d7e84d5c7--switche.netlify.app/?path=/docs/switche-experimental--default

## Reason

>Implement new SwitchE component and reach feature parity with legacy.

## Work carried out

- [x] Reach parity with legacy
- [x] Refactor to use make composable via JSX
- [x] Add support for invalid and disabled states
- [x] Add to forms example stories and Cypress tests
- [x] Enable keyboard control 

## Screenshot

Screenshot 2021-11-02 at 12 19 18
Screenshot 2021-11-02 at 12 19 22

## Developer notes

Worth stepping through the commits to get the benefit of the diff.


packages/react-component-library/src/components/SwitchE/SwitchE.tsx
```diff
@@ -0,0 +1,110 @@
...
+      data-testid="switch-wrapper"
+    >
+      {label && (
+        <StyledLegend data-testid="switch-legend">{label}</StyledLegend>
+      )}
+      <StyledContainer>
+        {React.Children.map(
+          children,
+          (child: React.ReactElement<SwitchOptionProps>) => {
+            warnIfOverwriting(child.props, 'name', SwitchEOption.name)
```
m7kvqbe1: Do we think this is the best way to deal with overwrites with 'private' props?

If a sub-component prop is for internal use only should we maybe denote so with an underscore prefix to make it obvious in the generated Storybook docs?

packages/react-component-library/src/components/SwitchE/SwitchE.stories.tsx
```diff
@@ -0,0 +1,103 @@
...
+const Template: Story<SwitchProps> = (props) => (
+  <SwitchE {...props}>
+    <SwitchEOption label="Day" value="1" />
+    <SwitchEOption label="Week" value="2" />
+    <SwitchEOption label="Month" value="3" />
+    <SwitchEOption label="Year" value="4" />
+  </SwitchE>
+)
+
+export const Default = Template.bind({})
```
jpveooys, line 24: :+1:

packages/react-component-library/src/components/SwitchE/SwitchE.tsx
```diff
@@ -0,0 +1,110 @@
...
+  label,
+  onChange,
+  size = SWITCHE_SIZE.FORMS,
+  isDisabled,
+  isInvalid,
+  children,
+  ...rest
+}) => {
+  const [active, setActive] = useState<string | undefined>()
+  const id = uuidv4()
```
jpveooys: Are we able to put this through `useMemo` so it doesn't change on every render (will reduce performance)?

packages/react-component-library/src/components/SwitchE/SwitchE.tsx
jpveooys: Hmm... Could a context be another way of achieving the same thing? Would it be better? :thinking:

packages/react-component-library/src/components/SwitchE/SwitchEOption.tsx
```diff
@@ -0,0 +1,75 @@
...
+  value,
+  isDisabled,
+  name,
+  id,
+  isActive,
+  onChange,
+}) => {
+  return (
+    <StyledSwitchOption
+      key={getKey('switch-option', label)}
```
jpveooys: Think we should be setting the keys inside the `.map()` really: https://reactjs.org/docs/lists-and-keys.html#extracting-components-with-keys

jpveooys (commented):
Looks nice :+1: 

Couple of other points:

- are the disabled hover styles correct?

- it would be good to conditional logic like this work:

  ```jsx
  () => {
    const showYear = false  // could be e.g. null or undefined also
    return (
      <SwitchE name="moo">
        <SwitchEOption label="Day" value="1" />
        <SwitchEOption label="Week" value="2" />
        <SwitchEOption label="Month" value="3" />
        {showYear && <SwitchEOption label="Year" value="4" />}
      </SwitchE>
    )
  }
  ```
  Not sure what the best way would be though :thinking: (Currently it complains about the boolean value.)

packages/react-component-library/src/components/SwitchE/SwitchEOption.tsx
m7kvqbe1: Oops yes that's left over from the refactor! Nice spot!

packages/react-component-library/src/components/SwitchE/SwitchE.tsx
m7kvqbe1: I'll leave it as is for now as we use this pattern elsewhere (Breadcrumbs come to mind). I'll add a note to the board to investigate.

m7kvqbe1:
re: Disabled hover styles - not sure - maybe it should have no hover? I'll refer to @k9dh3zij

packages/react-component-library/src/components/SwitchE/SwitchE.tsx
jpveooys: :+1: 

I do think it'll probably be cleaner so would be good to look at at some point :slightly_smiling_face:

jpveooys (approved):
Looks good (pending that question about the disabled hover styling).

(I am also wondering about keyboard navigation – but that seems to be a problem with the old switch component as well.)

packages/react-component-library/src/components/SwitchE/SwitchE.tsx
```diff
@@ -0,0 +1,119 @@
...
+      {...rest}
+      data-testid="switch-wrapper"
+    >
+      {label && (
+        <StyledLegend data-testid="switch-legend">{label}</StyledLegend>
+      )}
+      <StyledContainer>
+        {React.Children.map(children, (child: SwitchEChildType) => {
+          if (!React.isValidElement(child)) {
+            return null
```
thyhjwb6, line 86: Is this testable?

thyhjwb6 (approved):
Looks good 👍

k9dh3zij:
Thanks @m7kvqbe1, beautiful! 😁  Feedback:

1. Currently, when I click on an option, the `hover` style remains visible, until I move my mouse off of that option. Can we make it so that when I click on an option, the `selected` state becomes visible? (this should persist until the mouse pointer moves off of that option)

2. The currently active option shouldn't be a link (it's already active, it won't do anything). Therefore, it doesn't have a hover state either.

Demo of both # 1 and # 2:
 
Current:

https://user-images.githubusercontent.com/73531035/139296372-e214b884-6533-4aba-a7e0-28161ffac5fc.mov

Proposed:

https://user-images.githubusercontent.com/73531035/139295879-f9f19374-e66d-4b5e-b6c9-64928faa8791.mov

3. The `invalid` state border (3px wide) should not start outside the component default border. Pixel 1 or 3 of the error border should be the same as the component border itself.

image

4. Can we please reset the `invalid` state border, after the user selects an option?

5. Legend bottom margin should be 6px (currently 4px)

Current:
image

Spec:
image

6. Disabled controls should have no hover/focus state.

As discussed, let's chat about the focus styles tomorrow.
Thanks!

k9dh3zij:
@m7kvqbe1  Thanks for your patience. Further to the feedback above, as discussed we're introducing a keyboard focused state. This is consistent with CheckboxE and RadioE. I've updated the spec, as below:

image

**Shadow**
Line width: 3px
Colour: #ddf4ff (Action 100)
Radius top left, top right, bottom right and bottom left corners of Switch group only: 17

**Border (active)**
Line width: 3px
Colour: #3a8fdd (Action 500)
Radius top left, top right, bottom right and bottom left corners of Switch group only: 15


Thank you!

m7kvqbe1:
@k9dh3zij 

These styling changes have now been made. I have re-deployed the Netlify preview.

Note that control of the `isInvalid` state of the component is devolved to the consuming form code as such I can't force the error state to disappear upon option selection.

k9dh3zij:
Thanks for fixing these and adding keyboard focus states @m7kvqbe1 Feedback:

1. Border and shadow design for the active state of small size should be different to that of forms size design. Please see spec under TextField small size

image

2. When an option is highlighted with the keyboard, can we make it so if I press space it has the same effect as pressing Enter please? (same as how Buttons, Radios etc work)

3. I think the error border z-index needs fixing so the red border is in front of everything else **other than** the keyboard highlight

image

image

m7kvqbe1:
@k9dh3zij 

Screenshot 2021-11-02 at 15 40 41
Screenshot 2021-11-02 at 15 40 29

k9dh3zij:
Thanks @m7kvqbe1 

As discussed, can we change the keyboard tab behaviour so it works like Radio buttons do? (ie: tab into component, move with arrows, tab off the component).

That's all from me, thank you!

k9dh3zij:
👍 

@m7kvqbe1  As discussed, let's merge with current keyboard tab behaviour and we'll review further whether we should make consistent with other components (such as Radio / Checkboxes) and we can raise a separate issue then. Thanks!

PR #2758, opened 2021-10-28
Enable rule of hook linting rules

jpveooys:
## Related issue

#2454

## Overview

This enables the `react-hooks/rules-of-hooks` and `react-hooks/exhaustive-deps` ESLint rules and fixes the resulting errors across the code base.

## Link to preview

https://rules-of-hooks.netlify.app/

## Reason

These are standard React linting rules that should be running to prevent bugs.

## Work carried out

- [x] Fix rule of hook linting errors in DatePicker
- [x] Ditto, for NumberInput
- [x] Ditto, for useDocumentClick and related components
- [x] Ditto, for the Timeline
- [x] Add 'airbnb/hooks' to the ESLint configuration

## Developer notes

A full list of fixed errors can be found here:

https://gist.github.com/jpveooys/fb7251f6495d5dd1851f13d01bfc980a

All tests are passing, but I'd suggest doing some manual testing to double-check there haven't been any regressions.

These components have had changes:

- Date picker
- Number input
- Timeline

Additionally, the useDocumentClick changes affect the following components:

- Date picker
- Popover (in click mode)
- Masthead search
- Masthead user menu
- Sidebar (deprecated) notifications
- Sidebar (experimental) sub-navigation, notifications, user menu

I'd recommend looking at the commits individually.

You may find it easier to review with changes to white space hidden:

image

A new [`usePrevious`](https://reactjs.org/docs/hooks-faq.html#how-to-get-the-previous-props-or-state) hook has been added. This can be used to get the previous value of any variable (typically a prop).

Various uses of [`useCallback`](https://reactjs.org/docs/hooks-reference.html#usecallback) have been added as well. This stops inline functions from being redefined with each render, which can cause performance problems (would cause effects to run more frequently than needed in this case, can cause unnecessary re-renders in others).

## Related issues to raise

A couple of existing bugs were found while testing this:

- https://github.com/defencedigital/mod-uk-design-system/issues/2766
- https://github.com/defencedigital/mod-uk-design-system/issues/2765

### DIFF
--- a/packages/eslint-config-react/index.js
+++ b/packages/eslint-config-react/index.js
@@ -6,6 +6,7 @@ module.exports = {
   },
   extends: [
     'airbnb',
+    'airbnb/hooks',
     'plugin:@typescript-eslint/recommended',
     'prettier',
     'plugin:jest/recommended',
@@ -19,16 +20,6 @@ module.exports = {
     'arrow-body-style': 0,
     camelCase: 0,
     curly: ['error', 'all'],
-    'import/extensions': [
-      'error',
-      'ignorePackages',
-      {
-        js: 'never',
-        jsx: 'never',
-        ts: 'never',
-        tsx: 'never',
-      },
-    ],
     'import/no-extraneous-dependencies': 0,
     'import/no-unresolved': 0,
     'import/prefer-default-export': 0,
--- a/packages/react-component-library/src/components/Timeline/context/reducer.ts
+++ b/packages/react-component-library/src/components/Timeline/context/reducer.ts
@@ -18,7 +18,7 @@ import {
 } from './types'
 
 import { WEEK_START } from '../constants'
-import { initialiseScaleOptions } from './timeline_scales'
+import { buildScaleOptions } from './timeline_scales'
 
 const HOURS_IN_DAY = 24
 
@@ -103,9 +103,8 @@ export function reducer(
 ): TimelineState | never {
   switch (action.type) {
     case TIMELINE_ACTIONS.CHANGE_WIDTH:
-    case TIMELINE_ACTIONS.INITIALISE:
-      /* eslint-disable no-case-declarations */
-      const scaleOptions = initialiseScaleOptions(state.options, action.width)
+    case TIMELINE_ACTIONS.INITIALISE: {
+      const scaleOptions = buildScaleOptions(state.options, action.width)
       const currentScaleIndex =
         state.currentScaleIndex ||
         scaleOptions.findIndex(({ isDefault }) => isDefault)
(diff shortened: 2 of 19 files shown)


packages/eslint-config-react/index.js
```diff
@@ -19,16 +20,6 @@ module.exports = {
     'arrow-body-style': 0,
     camelCase: 0,
     curly: ['error', 'all'],
-    'import/extensions': [
```
jpveooys, line 22: This was being set again further down, I just removed this one.

packages/react-component-library/src/components/Timeline/context/reducer.ts
```diff
@@ -103,9 +103,8 @@ export function reducer(
 ): TimelineState | never {
   switch (action.type) {
     case TIMELINE_ACTIONS.CHANGE_WIDTH:
-    case TIMELINE_ACTIONS.INITIALISE:
-      /* eslint-disable no-case-declarations */
-      const scaleOptions = initialiseScaleOptions(state.options, action.width)
+    case TIMELINE_ACTIONS.INITIALISE: {
```
jpveooys, line 106: The linting rule was correct here – it's because variables end up being visible in other cases. Wrapping in braces fixes it.

packages/react-component-library/src/components/Timeline/context/timeline_scales.ts
```diff
@@ -125,7 +125,7 @@ function getMaxWidth(
   return unitWidthTotal
 }
 
-function initialiseScaleOptions(
+function buildScaleOptions(
```
jpveooys, line 128: Renamed this as I found the old name confusing (initialise made it sound like it has a side effect, when it just builds a data structure).

packages/react-component-library/src/components/Timeline/hooks/useTimelineZoom.ts
```diff
@@ -13,20 +12,10 @@ export function useTimelineZoom(): {
 } {
   const {
     dispatch,
-    state: { currentScaleIndex, options, scaleOptions, width },
+    state: { currentScaleIndex, scaleOptions },
   } = useContext(TimelineContext)
-  const [timelineScaleOptions, setTimelineScaleOptions] = useState(scaleOptions)
```
jpveooys, line 18: This was trying to keep its own copy of the scale options. The ones in the context are kept up-to-date though, so rather than trying to fix this logic I just changed it to use the scale options directly from the context which seems to work fine as far as I can see.

thyhjwb6: Ah nice, the code was changed recently and this makes it much cleaner 👍

packages/react-component-library/src/components/Timeline/hooks/useTimelineRowContent.ts
```diff
@@ -1,23 +1,29 @@
-import { useEffect, useRef, useState, MutableRefObject } from 'react'
+import React, { useEffect, useRef, useState, MutableRefObject } from 'react'
+import { TimelineEventsProps } from '../TimelineEvents'
```
thyhjwb6, line 3: Just needs an extra like break here

thyhjwb6 (approved):
Nice, looks good 👍

m7kvqbe1 (approved):
LGTM

PR #2762, opened 2021-10-29
Test Cypress console output

### LINKED ISSUE #2471: Add tests to interact with component and check for console output
To be done after: #2472

Very straightforward task to add:

```
Cypress.on('window:before:load', (win) => {
  cy.spy(win.console, 'error')
  cy.spy(win.console, 'warn')
})

afterEach(() => {
  cy.window().then((win) => {
    expect(win.console.error).to.have.callCount(0)
    expect(win.console.warn).to.have.callCount(0)
  })
})
```

to [support/index.ts](https://github.com/Royal-Navy/design-system/blob/master/packages/react-component-library/cypress/support/index.ts)

The final piece of this could be to remove `useInputValue.ts` and this change will provide the coverage.

thyhjwb6:
## Related issue
Closes #2471 

## Overview
Checks the console output for errors each time a command is run. As we add to the Cypress test suite we will have more coverage and less chance that unhandled errors split through. 

## Reason
Unhandled errors were slipping through.

## Work carried out
- [x] Add check

### DIFF
--- a/packages/react-component-library/cypress/support/consoleErrors.ts
+++ b/packages/react-component-library/cypress/support/consoleErrors.ts
@@ -1,0 +1,21 @@
+import { Cypress, cy } from 'local-cypress'
+
+function check() {
+  let stub
+
+  Cypress.on('window:before:load', (win) => {
+    stub = cy.stub(win.console, 'error')
+  })
+
+  Cypress.on('command:end', () => {
+    if (stub && stub.called) {
+      chai
+        .expect(stub, 'There should be no console errors')
+        .to.have.callCount(0)
+    }
+  })
+}
+
+export default {
+  check,
+}
--- a/packages/react-component-library/cypress/support/index.ts
+++ b/packages/react-component-library/cypress/support/index.ts
@@ -1,25 +1,4 @@
-// ***********************************************************
-// This example support/index.js is processed and
-// loaded automatically before your test files.
-//
-// This is a great place to put global configuration and
-// behavior that modifies Cypress.
-//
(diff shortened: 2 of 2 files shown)


packages/react-component-library/cypress/support/consoleErrors.ts
```diff
@@ -0,0 +1,19 @@
...
+function check() {
+  let stub
+
+  Cypress.on('window:before:load', (win) => {
+    stub = cy.stub(win.console, 'error')
+  })
+
+  Cypress.on('command:end', () => {
+    if (stub) {
+      chai.expect(stub, 'There are console errors').to.have.callCount(0)
```
jpveooys: I wonder if a message like 'There should be no console errors' would be clearer in the Cypress log?

image

packages/react-component-library/cypress/support/consoleErrors.ts
```diff
@@ -0,0 +1,19 @@
...
+
+function check() {
+  let stub
+
+  Cypress.on('window:before:load', (win) => {
+    stub = cy.stub(win.console, 'error')
+  })
+
+  Cypress.on('command:end', () => {
+    if (stub) {
```
jpveooys: It does seem a little bit spammy when there aren't any errors. Could add in ` && stub.called` here so it doesn't log anything on success (a bit of a cheat of course).

jpveooys (commented):
Nice :+1: 

Just a couple of small comments.

jpveooys (approved):

PR #2768, opened 2021-11-01
Fix Modal bugs in IE11

### LINKED ISSUE #2731: Chromatic `Modal` snapshot appears to be broken in ie 11
Screenshot 2021-10-21 at 17 46 29

jpveooys:
## Related issue

Fixes #2731

## Overview

This fixes a few problems with Modal:

- Modal wasn't rendering correctly in IE11 on Chromatic
- the primary and secondary buttons weren't being placed correctly in IE11 when there was no tertiary button
- the spacing between buttons in the mobile layout was lost recently when another bug was fixed

## Link to preview

https://5e25c277526d380020b5e418-ttsrdrgfqr.chromatic.com/

## Reason

To fix Modal bugs and get Chromatic working in IE11 for Modal.

The mobile button spacing appears to have been accidentally removed here: https://github.com/defencedigital/mod-uk-design-system/commit/aa764fa322bdda88187e0a32111f1d62f60a9ad6#diff-876bcbc5d87e1d790080d64c5def097c8fe7a506b27db3d8b783bf13c175906cL16-L18 (I just noticed it as the `margin-bottom: 0` overrides were still there).

## Work carried out

- [x] Fix glitch rendering Modal in IE11 on Chromatic
- [x] Tidy up Modal stories
- [x] Re-add spacing between buttons in mobile layout
- [x] Fix problems with button placement in IE11

### Chromatic glitch

See changes to IE snapshots here: https://www.chromatic.com/build?appId=5e25c277526d380020b5e418&number=4072.

#### Before

image

#### After

image

#### Before

image

#### After

image


packages/react-component-library/src/components/Modal/partials/StyledSecondaryButton.tsx
```diff
@@ -1,12 +1,5 @@
...
 import { StyledButton } from './StyledButton'
 
-const { mq } = selectors
-
-export const StyledSecondaryButton = styled(StyledButton)`
-  ${mq({ gte: 'xs' })`
-    margin-left: auto;
-  `}
-`
+export const StyledSecondaryButton = styled(StyledButton)``
```
m7kvqbe1: Is this partial redundant now? Could it be removed (in exchange for a different CSS selector in StyledFooter)?

m7kvqbe1 (approved):
Lovely fixes!

packages/react-component-library/src/components/Modal/partials/StyledSecondaryButton.tsx
jpveooys: Good question! This one is still being used in `StyledFooter.tsx`. But, while checking that, I found that Dialog was also trying to set mobile button padding on `StyledTertiaryButton`. That isn't needed now, and it was the last bit of styling on `StyledTertiaryButton` so that one has gone now :slightly_smiling_face:

thyhjwb6 (approved):
👍

packages/react-component-library/src/components/Modal/partials/StyledSecondaryButton.tsx
jpveooys: Good question! 

It seems so – I've swapped out the selector and removed `StyledSectoryButton` and all seems well.

Also, while checking that, I found that Dialog was also trying to set mobile button padding on `StyledTertiaryButton`. That isn't needed now, and it was the last bit of styling on `StyledTertiaryButton` so that one has gone now :slightly_smiling_face:

PR #2775, opened 2021-11-02
fix(Timeline): Use date control for stories

### LINKED ISSUE #2706: Compound Timeline Storybook docs page breaks if startDate value changed
### Steps to reproduce

1. Go to https://storybook.design-system.digital.mod.uk/?path=/docs/compound-timeline--default
2. (On the docs tab), click in the text input control for startDate:

  image

3. Click out of the text box

The page becomes blank and a number of errors are logged in the console.

The same happens if the reset button just above the input is clicked.

The cause may be this message in the console:

```
Starting with v2.0.0-beta.1 date-fns doesn't accept strings as date arguments. Please use `parseISO` to parse strings. See: https://git.io/fjule
```

thyhjwb6:
## Related issue
Closes #2706

## Overview
Uses the date control for date fields in `Timeline` stories.

## Reason
Control is suited to dates.

## Work carried out
- [x] Use date control
- [x] Log error when `endDate` is less than `startDate`

### Controls
Screenshot 2021-11-22 at 14 20 10

### Error
Screenshot 2021-11-22 at 14 19 57

### DIFF
--- a/packages/react-component-library/src/components/Timeline/Timeline.tsx
+++ b/packages/react-component-library/src/components/Timeline/Timeline.tsx
@@ -90,7 +90,7 @@ export const Timeline: React.FC<TimelineProps> = ({
   children,
   className,
   dayWidth,
-  hasSide,
+  hasSide = false,
   hideScaling = false,
   hideToolbar = false,
   isFullWidth = false,
--- a/packages/react-component-library/src/components/Timeline/Timeline.tsx
+++ b/packages/react-component-library/src/components/Timeline/Timeline.tsx
@@ -1,4 +1,5 @@
 import React from 'react'
+import { isAfter } from 'date-fns'
 
 import { ComponentWithClass } from '../../common/ComponentWithClass'
 import { TimelineProvider } from './context'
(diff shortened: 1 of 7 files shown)


packages/react-component-library/src/components/Timeline/__tests__/Timeline.test.tsx
```diff
@@ -2094,6 +2094,47 @@ describe('Timeline', () => {
...
 
+  describe('when `endDate` is before `startDate`', () => {
+    let consoleErrorSpy: jest.SpyInstance
+
+    beforeAll(() => {
+      timezoneMock.register('Europe/London')
+    })
+
+    beforeEach(() => {
+      consoleErrorSpy = jest.spyOn(global.console, 'error')
```
jpveooys, line 2105: Call `consoleErrorSpy.mockRestore()` in an `afterEach()`?

(Could turn on https://jestjs.io/docs/configuration#restoremocks-boolean, but looks like we'd need to fix some tests...)

jpveooys (approved):
Nice one! One small comment

packages/react-component-library/src/components/Timeline/Timeline.tsx
```diff
@@ -86,11 +87,20 @@ function getHoursBlockSize(
   )
 }
 
+function getEndDate(startDate: Date, endDate: Date) {
+  if (startDate && endDate && isAfter(startDate, endDate)) {
+    console.error('`startDate` is after `endDate`')
```
m7kvqbe1: Can we use the wrapped logger here?

 https://github.com/defencedigital/mod-uk-design-system/blob/master/packages/react-component-library/src/utils/logger.ts

m7kvqbe1 (dismissed):
LGTM just a comment about the logging

jpveooys (approved):

PR #2779, opened 2021-11-03
fix(DatePicker): Fix date validation when using custom date format

### LINKED ISSUE #2766: DatePicker: Keyboard date entry fails when using custom date format
### Steps to reproduce

1. Go to https://storybook.design-system.digital.mod.uk/?path=/story/date-picker--custom-format
2. Click in the date picker
3. (Optional) type a new date using the keyboard
3. Click out of the input

A red invalid value border appears around the input.

### Screenshot

datepicker-custom

jpveooys:
## Related issue

Fixes #2766 

## Overview

This fixes the validation of types date when a custom date format is being used.

## Link to preview

https://date-picker-custom-format.netlify.app/

## Reason

It was a bug.

## Work carried out

- [x] Add relevant tests
- [x] Amend date validation logic

### Before

datepicker-custom-before

### After

datepicker-custom-after

## Developer notes

The first problem was that a regular expression was being used to initially validate the date, and this only accounted for the default date format.

The second problem was that the `isMatch` (and `parse`) functions from `date-fns` accept years less than four-digits, and so while typing e.g. 01/01/2021, `onChange` is called four times as the year is being typed.

The solution was to try and parse the date (which is what `isMatch` does already), and then check that the year of the parsed date is not less than or equal to 999.

I also refactored some of the logic slightly to avoid having to validate and parse the date multiple times in a row.

### DIFF
--- a/packages/react-component-library/src/components/DatePicker/DatePicker.test.tsx
+++ b/packages/react-component-library/src/components/DatePicker/DatePicker.test.tsx
@@ -14,6 +14,7 @@ import { DatePicker } from '.'
 import { Button } from '../Button'
 
 const NOW = '2019-12-05T11:00:00.000Z'
+const ERROR_BORDER = `1px solid ${ColorDanger600}`
 
 function click(element: HTMLElement) {
   fireEvent.mouseDown(element)
@@ -453,7 +454,7 @@ describe('DatePicker', () => {
             it('should be in an error state', () => {
               expect(
                 wrapper.getByTestId('datepicker-outer-wrapper')
-              ).toHaveStyleRule('border', `1px solid ${ColorDanger600}`)
+              ).toHaveStyleRule('border', ERROR_BORDER)
             })
 
             describe('and the escape key is pressed', () => {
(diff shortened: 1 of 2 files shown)


packages/react-component-library/src/components/DatePicker/DatePicker.test.tsx
```diff
@@ -925,6 +932,61 @@ describe('DatePicker', () => {
...
+      it('invokes the `onChange` callback', () => {
+        const expectedDate = new Date('2016-02-03T12:00:00.000Z')
+        expect(onChange).toHaveBeenCalledTimes(1)
+        expect(onChange).toHaveBeenCalledWith({
+          startDate: expectedDate,
+          endDate: expectedDate,
+        })
+      })
+
+      it("isn't in an error state", () => {
```
m7kvqbe1, line 952: Single quotes here

m7kvqbe1 (approved):
LGTM

packages/react-component-library/src/components/DatePicker/DatePicker.test.tsx
jpveooys: Nested single quote there – Prettier will change it even if you escape it.

PR #2807, opened 2021-11-17
Change default Storybook layout and improve Story sorting

jpveooys:
## Overview

This makes a couple of changes:

- changes the default layout of stories from `fullscreen` to the default layout (`padded`) (while changing certain stories to use `fullscreen` where it makes sense)
- sorts folders in Storybook alphabetically
- changes the Meta and Story types to be imported from `@storybook/react` instead of `@storybook/react/types-6-0`

## Link to preview

https://5e25c277526d380020b5e418-kqmsqxgvha.chromatic.com/

### Layout

For many stories, a full-screen layout was problematic because it caused shadows and borders to be out of view on the Canvas tab:

image

With padding these can be properly seen:

image

I've gone through and manually applied the `fullscreen` layout to stories for components where it makes sense or wasn't getting in the way of borders or shadows.

(There shouldn't be any change to the Docs tab.)

### Sorting

The change to sorting is to stop some story folders being sorted last:

image

Instead these are sorted alphabetically:

image

There is no change to the sorting of individual stories, but note that we are currently getting hit by this in production: https://github.com/storybookjs/storybook/issues/15574

This is currently causing some individual stories to be in the wrong order in production e.g. https://storybook.design-system.digital.mod.uk/?path=/docs/range-slider--custom-value-formatter

### Imports

This change is to simply make the imports in line with [examples in the Storybook docs](https://storybook.js.org/docs/react/get-started/setup):

image

## Work carried out

- [x] Remove the default `fullscreen` layout override
- [x] Go through stories and set certain ones to use a `fullscreen` layout as appropriate
- [x] Configure story sorting
- [x] Correct type imports

### DIFF
--- a/packages/react-component-library/src/components/CheckboxE/CheckboxE.stories.tsx
+++ b/packages/react-component-library/src/components/CheckboxE/CheckboxE.stories.tsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import { Story, Meta } from '@storybook/react/types-6-0'
+import { Story, Meta } from '@storybook/react'
 import { action } from '@storybook/addon-actions'
 import { Field, Formik, Form } from 'formik'
 import * as yup from 'yup'
--- a/packages/react-component-library/src/components/CheckboxE/CheckboxE.stories.tsx
+++ b/packages/react-component-library/src/components/CheckboxE/CheckboxE.stories.tsx
@@ -11,7 +11,7 @@ import { FormikGroupE } from '../FormikGroup'
 
 export default {
   component: CheckboxE,
-  title: 'CheckboxE',
+  title: 'Checkbox (Experimental)',
   parameters: {
     actions: { argTypesRegex: '^on.*' },
   },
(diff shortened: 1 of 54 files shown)


packages/react-component-library/src/components/CheckboxE/CheckboxE.stories.tsx
```diff
@@ -1,5 +1,5 @@
 import React from 'react'
-import { Story, Meta } from '@storybook/react/types-6-0'
+import { Story, Meta } from '@storybook/react'
```
m7kvqbe1, line 2: I've just noticed the name of this set of component seems to be incorrect in the Sidebar.

Could we change this to `Checkbox (Experimental)`.

packages/react-component-library/src/components/RadioE/RadioE.stories.tsx
```diff
@@ -1,5 +1,5 @@
 import React from 'react'
```
m7kvqbe1, line 1: Same as CheckboxE - needs name changing to `Radio (Experimental)` in Storybook sidebar.

m7kvqbe1 (changes requested):
LGTM - just noticed some naming inconsistencies that could be fixed quickly in this PR.

m7kvqbe1 (approved):

PR #2808, opened 2021-11-17
fix(Storybook): Firefox Storybook docs obscures components

### LINKED ISSUE #2810: Storybook obscures hidden components (like `DatePicker`) in Firefox browser
This only happens in Firefox:

Screenshot 2021-11-17 at 11 05 19

m7kvqbe1:
## Related issue

Closes #2810

storybookjs/storybook#16714

storybookjs/storybook#8928

## Overview

Overiding some Storybook styles to fix presentation when viewed using the Firefox browser.

## Reason

>Storybook obscures components that have hidden elements (think ContextMenu or DatePicker) when viewed with the Firefox browser.

## Work carried out

- [x] Run Cypress e2e suites against Firefox in CI
- [x] Add `ContextMenu` test to catch issue
- [x] Overide some Storybook styles to fix Firefox presentation

## Screenshot

Screenshot 2021-11-17 at 11 05 03
Screenshot 2021-11-17 at 11 05 19

## Developer notes

Added slightly different assertion to account for Firefox behaviour with `DatePicker`.

Issue created pending further investigation: https://github.com/defencedigital/mod-uk-design-system/issues/2809

### DIFF
--- a/.github/workflows/build_and_test_feature.yml
+++ b/.github/workflows/build_and_test_feature.yml
@@ -3,279 +3,272 @@ name: Build & Test Feature
 on: pull_request
 
 jobs:
-     Build_icon_library:
-      runs-on: ubuntu-latest
-      steps:
-        - name: Git clone repository
-          uses: actions/checkout@v2
+  Build_icon_library:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Git clone repository
+        uses: actions/checkout@v2
 
-        - name: Cache Node modules
-          uses: actions/cache@v2
-          with:
-            path: '**/node_modules'
-            key: ${{ runner.os }}-modules-${{ secrets.CACHE_VERSION }}-${{ hashFiles('**/yarn.lock') }}
+      - name: Cache Node modules
+        uses: actions/cache@v2
+        with:
+          path: '**/node_modules'
+          key: ${{ runner.os }}-modules-${{ secrets.CACHE_VERSION }}-${{ hashFiles('**/yarn.lock') }}
 
-        - name: Cache Cypress binary
-          uses: actions/cache@v2
-          with:
-            path: ~/.cache/Cypress
-            key: cypress-${{ runner.os }}-cypress-${{ hashFiles('**/yarn.lock') }}
-            restore-keys: |
-             cypress-${{ runner.os }}-cypress-
+      - name: Cache Cypress binary
+        uses: actions/cache@v2
+        with:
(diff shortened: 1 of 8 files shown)


jpveooys:
Did you mean to link #2810 instead of #418?

.github/workflows/build_and_test_feature.yml
```diff
@@ -3,279 +3,280 @@ name: Build & Test Feature
...
+
+      - name: Confirm Cypress binary is installed
+        run: |
+          result=$(yarn cypress version)
+          if [[  `echo $result` == *"not installed"* ]];
+            then echo "cypress binary is not installed, installing binary...." && yarn cypress install;
+            else echo "cypress binary is installed, no action required";
+          fi
+
+      - name: Run e2e tests - Chrome
```
jpveooys: Would be nice to use a matrix for this if we can: https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix

m7kvqbe1: Ooh nice! I'll change it now!

jpveooys (commented):
Nice one! This was driving me insane :smile: 

There's a small glitch with the Modal and Dialog backgrounds at the top if it's easily fixable (not to worry if not):

image

packages/react-component-library/cypress/specs/ContextMenu/index.spec.ts
```diff
@@ -0,0 +1,23 @@
+import { describe, cy, it, before } from 'local-cypress'
```
jpveooys: Could sort the imported names alphabetically:

```suggestion
import { before, cy, describe, it } from 'local-cypress'
```

m7kvqbe1:
Nice spot! That has also been sorted now.

.github/workflows/build_and_test_master.yml
```diff
@@ -146,6 +146,8 @@ jobs:
      Test-e2e_react-component-library:
       runs-on: ubuntu-latest
       needs: [Build_icon_library, Lint_react-component-library]
+      matrix:
```
jpveooys: FYI it's not building at the moment – `matrix` should be nested under a `strategy` key

m7kvqbe1: 🤦‍♂️

jpveooys (approved):
Nice one :+1:

PR #2813, opened 2021-11-17
docs(Storybook): Enable dynamic source code generation

### LINKED ISSUE #2811: Storybook story source code isn't being generated correctly
### Steps to reproduce

1. Go to e.g. https://storybook.design-system.digital.mod.uk/?path=/docs/button-experimental--default
2. Click on the code button

The source code isn't generated correctly:

image

However, this is supposed to work: https://github.com/storybookjs/storybook/pull/11332

It can also be seen working in this example here:
https://next--storybookjs.netlify.app/official-storybook/?path=/docs/docs-source--loading

We need to investigate why it's not working in this project.

jpveooys:
## Related issue

Fixes #2811

## Overview

This turns on dynamic source code generation on the Docs tab in Storybook.

## Link to preview

https://5e25c277526d380020b5e418-ziursfrpsg.chromatic.com/

## Reason

To generate more useful code examples.

## Work carried out

- [x] Turn on dynamic source code generation by default
- [x] Check each docs page in Storybook
- [x] Tweak a few stories for better code snippets
- [x] Turn off source code generation for certain Select and Table stories due to incompatibility

### Before

code-before

### After

code-after

## Developer notes

There were a couple of limitations found which workarounds have been put in for:

- a couple of stories were being hit by storybookjs/storybook#16730, I've amended those stories to work around that
- it also breaks IE11 on Chromatic, so it's disabled there too

(If you want to look through the generated source more easily, you can set `parameters.docs.source.state` to `open` in `packages/react-component-library/.storybook/preview.js` open the code panels by default.)

### DIFF
--- a/packages/react-component-library/.storybook/preview.js
+++ b/packages/react-component-library/.storybook/preview.js
@@ -29,9 +29,18 @@ function clickDocsButtonOnFirstLoad() {
 window.addEventListener('load', clickDocsButtonOnFirstLoad)
 
 export const parameters = {
+  jsx: {
+    // Filter out callback props injected by the Actions addon
+    // from dynamic code snippets on the Docs tab as these are
+    // noisy and usually not needed
+    filterProps: (value) =>
+      value !== undefined && value?.name !== 'actionHandler',
+  },
   docs: {
     source: {
-      type: 'code',
+      // Avoid breaking IE11 on Chromatic as WeakSet isn't available
+      // there
+      type: typeof WeakSet === undefined ? 'code' : 'auto',
     },
   },
   options: {
--- a/packages/react-component-library/src/components/ContextMenu/ContextMenu.stories.tsx
+++ b/packages/react-component-library/src/components/ContextMenu/ContextMenu.stories.tsx
@@ -1,5 +1,6 @@
 import React, { useRef } from 'react'
 import { Story, Meta } from '@storybook/react'
+import styled from 'styled-components'
 
 import { IconEdit, IconDelete, IconAdd } from '@defencedigital/icon-library'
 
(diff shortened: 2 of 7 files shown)


jpveooys:
I think I've got a better workaround for the nested JSX problem – let me double-check and push a fixup if I do...

m7kvqbe1:
Looks good to me - it's a pitty that the dynamic source code generation doesn't handle the callbacks nicely.

I wonder if there is a way to fix that. Does `showFunctions` do anything?

https://github.com/storybookjs/storybook/blob/8d7fa4249cc73f315cfc15ebd8c6f0d574f341d5/addons/docs/src/frameworks/react/jsxDecorator.tsx#L142

Screenshot 2021-11-19 at 10 04 44

jpveooys:
It's worse with `showFunctions` on :smile: 

image

Those functions are coming from the actions addon from the looks of it. 

One option is to try and replace those with something prettier:

```
export const parameters = {
  docs: {
    transformSource: (source) =>
      source.replaceAll(/function noRefCheck\s?\(\)\s?{}/g, '() => {}'),
...
```

Which renders:

image

(In fact they are trying to do that already, but they're using `replace` instead of `replaceAll`...
https://github.com/storybookjs/storybook/blob/8d7fa4249cc73f315cfc15ebd8c6f0d574f341d5/addons/docs/src/frameworks/react/jsxDecorator.tsx#L137
)

Another option is to try and strip them out using the same approach, but it may be misleading if the props are required, and can cause problems with line wrapping. That would end up looking something like this:

image

(using `/\n\s+\w+={(function noRefCheck\s?\(\)\s?{}|\(\)\s?=>\s?{})}/g` for the regex).

jpveooys:
Just found a better option actually:

```
export const parameters = {
  jsx: {
    filterProps: (value, key) =>
      value !== undefined && value.name !== 'actionHandler',
  },
...
```

Ends up with:

image

Let me test that a bit more thoroughly and will push a fixup...

jpveooys:
I've put that in now so the injected actions prop values will be removed from the generated source.

I've also amended the stories that were being hit by https://github.com/storybookjs/storybook/issues/16730 to work around that problem without disabling dynamic source code generation for them. (The fix is to use the args pattern, or otherwise move the array with nested JSX outside of the function component body.)

(There was one Select story that had an incorrect height so I've fixed that as well.)

m7kvqbe1 (approved):
LGTM

packages/react-component-library/.storybook/preview.js
```diff
@@ -29,6 +29,10 @@ function clickDocsButtonOnFirstLoad() {
 window.addEventListener('load', clickDocsButtonOnFirstLoad)
 
 export const parameters = {
+  jsx: {
+    filterProps: (value, key) =>
```
m7kvqbe1: I wonder if this warrants a comment

jpveooys: Done!

PR #2837, opened 2021-11-25
test(DatePicker): Fix Cypress border assertions in Firefox

### LINKED ISSUE #2809: Firefox `DatePicker` error-less expected behaviour differs from Chrome in Cypress tests
Run Cypress DatePicker e2e test suite.

Browser behaviour is inconsistent (this may just be a browser quirk).

- Chrome: sets the outer wrapper border (assumed to be because of focus state) to `ColorAction600`.

- Firefox: does not set any outer wrapper border (assumed lack of focus state).

jpveooys:
## Related issue

Resolves #2809

## Overview

Update the Cypress border assertions for DatePicker and DatePickerE to work in both Firefox and Chrome.

## Reason

CSS assertions on the `border` shorthand property aren't supported in Firefox. Assertions on the constituent components have to be done instead (`border-top-color`, `border-right-color` etc.) (which works in Chrome as well).

## Work carried out

- [x] Add custom Cypress Chai assertion for border styles
- [x] Update DatePicker and DatePickerE Cypress tests

## Developer notes

I added a custom Chai assertion to do these kind of assertions more easily.

### DIFF
--- a/packages/react-component-library/cypress/specs/DatePicker/index.spec.ts
+++ b/packages/react-component-library/cypress/specs/DatePicker/index.spec.ts
@@ -1,10 +1,6 @@
 import { describe, cy, it, before } from 'local-cypress'
 import { addDays, startOfMonth, format } from 'date-fns'
-import {
-  ColorAction600,
-  ColorNeutral200,
-  ColorDanger600,
-} from '@defencedigital/design-tokens'
+import { ColorAction600, ColorNeutral200 } from '@defencedigital/design-tokens'
 
 import { DATE_FORMAT } from '../../../src/constants'
 import { hexToRgb } from '../../helpers'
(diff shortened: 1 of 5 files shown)


packages/react-component-library/cypress/specs/DatePicker/index.spec.ts
```diff
@@ -39,25 +35,13 @@ describe('DatePicker', () => {
...
 
-      it('should not be in an error state', { browser: '!firefox' }, () => {
+      it('should not be in an error state', () => {
         cy.get(selectors.datePicker.outerWrapper).should(
-          'have.css',
-          'border',
-          `1px solid ${hexToRgb(ColorNeutral200)}`
+          'has.border',
+          'color',
+          hexToRgb(ColorNeutral200)
```
jpveooys, line 42: I've only done assertions on the border colour as that seems like the relevant thing here.

thyhjwb6 (approved):
👍

m7kvqbe1 (approved):

Issue #2761, opened 2021-10-29
Review the need for font breakpoints and remove if not needed


k9dh3zij:
See video example below. When changing viewport size, buttons move about slightly. 

From @jpveooys:

There's a bit of logic in what's called the GlobalStyleProvider that makes small tweaks to the font size depending on the viewport width:
https://github.com/defencedigital/mod-uk-design-system/blob/master/packages/design-tokens/src/tokens/light/breakpoints.json

I agree it's odd (we turned it off in MSP). It's not something visible in Storybook as I don't think the GlobalStyleProvider is loaded there.

This was annoying as different people can be looking at slightly different things, and the modals get taller as the browser gets wider.

https://user-images.githubusercontent.com/73531035/139437276-ca47dba0-1dc1-4f0a-8eae-195820246d86.mov

jpveooys:
@k9dh3zij Just updated the title to be a bit more specific as GlobalStyleProvider does other stuff as well :slightly_smiling_face:

PR #2841, opened 2021-11-26
feat(DatePickerE): Initial updates based on v3 design spec

### LINKED ISSUE #2439: Initial iteration of `DatePickerE`
[Design spec](https://docs.google.com/document/d/1MJlbdqhxQvt5qGn87Ubs2_OZUusdIgHwdV8NEofAMqQ/edit#heading=h.39kk8xu)

## Tasks 

- [x] Copy existing component
- [x] Remove deprecated features
- [x] Update styling of input
- [x] Update styling of calendar body
- [x] Add selection shading on hover for date ranges
- [x] Integrate with floating box component
- [x] Update styling and positioning of floating box
- [x] Update placeholder and automatic opening behaviour
- [x] Reopen on correct date

jpveooys:
## Related issue

Resolves #2439

## Overview

This makes various updates to the DatePickerE component based on the v3 design spec.

## Link to preview

- [Storybook](https://5e25c277526d380020b5e418-bwgmozgpbk.chromatic.com/?path=/docs/date-picker-experimental--custom-format)

- [Non-Storybook example](https://red-playground.netlify.app/)
  (base font size overridden to 100% due to #2761)

## Reason

Incremental work towards the v3 release.

## Work carried out

- [x] Update the styling of the input control
- [x] Update the styling of the day picker
- [x] Update the styling and positioning of the floating box
- [x] Remove `isValid` prop
- [x] Append date format to label automatically
- [x] Reuse partials from TextInputE
- [x] Add hover shading for intermediate dates in range mode
- [x] Update placeholder and automatic opening behaviour

### Single date

date-picker-e-single

### Range

date-picker-e-range

## Developer notes

Does not include:

- updated typing behaviour for single date mode
- typing and input date highlighting for range date picker
- time and period selection
- clear button
- month and year buttons
- #2863
- #2864


packages/react-component-library/src/components/DatePickerE/DatePickerE.tsx
```diff
@@ -153,20 +161,20 @@ export const DatePickerE: React.FC<DatePickerEProps> = ({
       >
         <StyledOuterWrapper
           data-testid="datepicker-outer-wrapper"
-          $hasFocus={open}
+          $hasFocus={hasFocus && !hasError}
```
jpveooys, line 182: The `&& !hasError` is temporary here to preserve the existing behaviour (which is covered by tests). It will be reviewed as part of #2863.

packages/react-component-library/src/components/DatePickerE/partials/StyledFloatingBox.tsx
```diff
@@ -0,0 +1,23 @@
+import { selectors } from '@defencedigital/design-tokens'
+import styled from 'styled-components'
+
+import { FloatingBox } from '../../../primitives'
+import { StyledArrow } from '../../../primitives/FloatingBox/partials/StyledArrow'
+import { StyledContent } from '../../../primitives/FloatingBox/partials/StyledContent'
+
+const { color } = selectors
+
+export const StyledFloatingBox = styled(FloatingBox)`
```
jpveooys, line 10: I've left these as overrides for now as it's not yet clear if these will be directly applicable anywhere else.

jpveooys:
@k9dh3zij I've added the outer shadow to the calendar sheet. Please have a look if it's as intended, as how it looks will depend on the origin (in relation to the other borders and shadows).

k9dh3zij:
Thanks @jpveooys loving this! Feedback:

**Visual**

- The button looks slightly squashed - should be consistent with other components:
image

- When filled, the input label should be 2px higher and  the user input should be 1px higher. Please note: I think this needs to change across all experimental inputs - not sure if I missed it in previous PRs or if something changed, sorry
Visual comparison with design:
image

- As already discussed, missing Sheet drop shadow


**Keyboard controls** 

- When opening the Sheet, can we move focus to the currently selected date? (as in https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html)

- When closing the Sheet, can we return focus to the button? (as in the W3 example)

- Range date picker: when using the keyboard to select the end-date, the visual highlight between start date and currently focused end date doesn't work
image

- Can we make it so that the Sheet contains its own tab sequence. (Tab and Shift + Tab do not move focus outside the dialog)
- Further to the above, can we add Esc as a shortcut to close the Sheet please? 

That's all for now, I think 😅  Thanks!

packages/react-component-library/src/components/DatePickerE/DatePickerE.tsx
```diff
@@ -199,16 +216,17 @@ export const DatePickerE: React.FC<DatePickerEProps> = ({
...
             ref={inputButtonRef}
             type="button"
-            onClick={open ? handleOnClose : handleOnFocus}
+            onClick={open ? handleOnClose : handleOnOpen}
             disabled={isDisabled}
             data-testid="datepicker-input-button"
           >
-            <StyledSeparator />
-            <DropdownIndicatorIcon isOpen={open} />
+            <StyledIconCalendarWrapper>
```
jpveooys: Just noticed `StyledIconCalendarWrapper` is misnamed – will rename to `StyledIconEventWrapper`.

jpveooys:
@k9dh3zij 

I've made updates to address those except for the following two:

> When opening the Sheet, can we move focus to the currently selected date? (as in https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html)

There was an issue already open in the underlying library we're using (react-day-picker) regarding this. I've created a separate issue #2871 at our end with more info and responded to the issue in the underlying library.

> Can we make it so that the Sheet contains its own tab sequence. (Tab and Shift + Tab do not move focus outside the dialog)

I found a library that will add this behaviour (called a 'focus trap'). However, it has a bug where the tab key stops functioning when dates after the first of the month are focused, so I've left that change out and created a separate issue (#2870). I also need to report the bug upstream.

Regarding:

> Range date picker: when using the keyboard to select the end-date, the visual highlight between start date and currently focused end date doesn't work

I've added this but note it's not possible at the moment to clear that highlight when tabbing away with react-day-picker v7 (it will be with v8 which is currently in beta). (Have created a separate issue for this as well: #2872.)

I also updated the border radius of the sheet and the hover/pressed/focus styles of the next and previous month buttons, as they seemed wrong before to me. Have another look at those if you can :slightly_smiling_face:

packages/react-component-library/src/components/DatePickerE/DatePickerE.test.tsx
```diff
@@ -605,6 +605,32 @@ describe('DatePickerE', () => {
       expect(wrapper.getByText('December 2019')).toBeInTheDocument()
     })
 
+    describe.each([
```
thyhjwb6, line 608: 👌

thyhjwb6 (commented):
Looks good 👍

m7kvqbe1:
Are the disabled state styles correct (currently the picker button is blue and clickable):

Screenshot 2021-12-06 at 14 38 39

packages/react-component-library/src/components/DatePickerE/DatePickerE.tsx
```diff
@@ -1,28 +1,41 @@
...
+import { useFocus } from '../../hooks/useFocus'
+import { useRangeHoverOrFocusDate } from './useRangeHoverOrFocusDate'
 import { useSelection } from './useSelection'
+import { useStatefulRef } from '../../hooks/useStatefulRef'
+import { WEEKDAY_TITLES } from './constants'
+
+declare module 'react-day-picker' {
+  // eslint-disable-next-line no-shadow
+  interface DayPickerProps {
+    // This prop is currently missing from the react-day-picker types
```
m7kvqbe1, line 31: 😭

packages/react-component-library/src/components/TextInputE/TextInputE.test.tsx
```diff
@@ -41,7 +41,7 @@ describe('TextInputE', () => {
     it('should apply the `$hasContent` style rule to label', () => {
       expect(wrapper.getByTestId('text-input-label')).toHaveStyleRule(
         'transform',
-        'translate(11px,8px) scale(0.75)'
+        'translate(11px,6px) scale(0.75)'
```
m7kvqbe1, line 44: Missing whitespace after comma

m7kvqbe1 (changes requested):
LGTM - just a couple of tiny comments

packages/react-component-library/src/components/TextInputE/TextInputE.test.tsx
jpveooys: This one's a test – it comes through like that (from jsdom or whatever is normalising it...)

packages/react-component-library/src/components/DatePickerE/DatePickerE.tsx
jpveooys: Have reported it: https://github.com/gpbl/react-day-picker/issues/1317

It is already fixed for react-day-picker v8 though (in beta at the moment).

jpveooys:
> Are the disabled state styles correct (currently the picker button is blue and clickable):

Good spot! Have updated those.

m7kvqbe1 (approved):
👍

k9dh3zij:
Thanks for the updates @jpveooys looking great!

As discussed, something that we can look to improve when the react-day-picker library allows it, please:

- Disabled dates shouldn't get focus. Therefore, the state shown here should not be possible:
image

Big 👍 from me on this! :)

jpveooys:
Thanks @k9dh3zij !

It looks like that's covered by https://github.com/gpbl/react-day-picker/issues/1320 which was coincidentally opened 30 minutes ago :smile:

PR #2844, opened 2021-11-29
feat(FloatingBox): Integrate with DatePickerE

jpveooys:
## Related issue

#2439

(Split from #2841.)

## Overview

This makes it possible for FloatingBox to use an external target element, and integrates it with DatePickerE.

## Reason

Code reuse and to simplify DatePickerE.

(It wasn't possible to integrate it as it was as the styling on `StyledTarget` interferes with the width of the date picker input.)

## Work carried out

- [x] Add ability for FloatingBox to use an external target
- [x] Update DatePickerE to use FloatingBox

### DIFF
--- a/packages/react-component-library/src/components/Popover/Popover.tsx
+++ b/packages/react-component-library/src/components/Popover/Popover.tsx
@@ -15,7 +15,7 @@ import { useHideShow } from '../../hooks/useHideShow'
 export interface PopoverProps
   extends Omit<
     FloatingBoxProps,
-    'onMouseEnter' | 'onMouseLeave' | 'renderTarget'
+    'onMouseEnter' | 'onMouseLeave' | 'renderTarget' | 'targetElement'
   > {
   /**
    * JSX target element to attach the Popover to.
--- a/packages/react-component-library/src/primitives/FloatingBox/FloatingBox.tsx
+++ b/packages/react-component-library/src/primitives/FloatingBox/FloatingBox.tsx
@@ -13,17 +13,39 @@ import { StyledArrow } from './partials/StyledArrow'
 import { useFloatingElement } from '../../hooks/useFloatingElement'
 import { FLOATING_BOX_SCHEME } from './constants'
 
-export interface FloatingBoxProps extends PositionType, ComponentWithClass {
+export interface FloatingBoxBaseProps extends PositionType, ComponentWithClass {
   role?: string
   contentId?: string
   scheme?: FloatingBoxSchemeType
   onMouseEnter?: (e: React.MouseEvent) => void
   onMouseLeave?: (e: React.MouseEvent) => void
   children?: React.ReactElement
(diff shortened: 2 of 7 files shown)


packages/react-component-library/src/components/Popover/Popover.tsx
```diff
@@ -15,7 +15,7 @@ import { useHideShow } from '../../hooks/useHideShow'
 export interface PopoverProps
   extends Omit<
     FloatingBoxProps,
-    'onMouseEnter' | 'onMouseLeave' | 'renderTarget'
+    'onMouseEnter' | 'onMouseLeave' | 'renderTarget' | 'targetElement'
```
thyhjwb6, line 18: A thought here is whether this can be covered by a test. Not a showstopper though.

packages/react-component-library/src/primitives/FloatingBox/FloatingBox.tsx
```diff
@@ -49,11 +72,18 @@ export const FloatingBox: React.FC<FloatingBoxProps> = ({
     arrowElementRef,
     styles,
     attributes,
-  } = useFloatingElement(placement)
+  } = useFloatingElement(placement, undefined, targetElement)
```
thyhjwb6, line 75: Wondering if it could work well to swap the parameters on `useFloatingElement` hence not needing to pass `undefined`.

thyhjwb6 (approved):
Looks great 👍

packages/react-component-library/src/primitives/FloatingBox/FloatingBox.tsx
jpveooys: I did want to change the signature of `useFloatingElement` but we're exporting it for some reason – will have to leave it alone for now but we could change it or make it internal for v3.

packages/react-component-library/src/components/Popover/Popover.tsx
jpveooys: Did you have something specific in mind? It would be a type error to pass `targetElement` (though we have `// @ts-nocheck` on the tests for some reason :man_shrugging: )

thyhjwb6: I was thinking something along the lines of instantiating the component with the omitted props and perhaps wrapping with a `try catch` but I cannot find an easy solution which makes me think it's not worth it.

packages/react-component-library/src/primitives/FloatingBox/FloatingBox.tsx
m7kvqbe1: I think the thinking was to expose the hook for people to use in situations where FloatingBox was not appropriate.

I don't think anyone is using it at the moment though...

m7kvqbe1 (approved):
👍

PR #2845, opened 2021-11-29
feat(RangeSliderE): Base implementation with feature parity

### LINKED ISSUE #2436: Implement `RangeSliderE` (legacy feature parity)
https://docs.google.com/document/d/1MJlbdqhxQvt5qGn87Ubs2_OZUusdIgHwdV8NEofAMqQ/edit

m7kvqbe1:
## Related issue

Closes #2436

## Overview

Create new experimental `RangeSliderE` component based on design spec.

## Link to preview

https://61a8f4b587ed053ee108b405--rangeslidere.netlify.app

## Reason

>Implement new `RangeSliderE` component and reach feature parity with legacy.

## Work carried out

- [x] Create base component implementation
- [x] Use new story format
- [x] Make some performance enhancements
- [x] Fix bug when clicking to change slider value
- [x] Add `hasMarkers` prop to toggle markers
- [x] Remove deprecated `hasPercentage` feature
- [x] Style in line with design spec

## Screenshot

Screenshot 2021-11-29 at 16 01 45

## Developer notes

The design spec changes the way threshold functionality works when combined with multiple handles. 

Needs discussion with design authority to elaborate on desired behaviour.

This will be implemented as part of a seperate PR: https://github.com/defencedigital/mod-uk-design-system/issues/2849


packages/react-component-library/src/components/RangeSliderE/partials/StyledHandle.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+interface StyledHandleProps {
+  $isActive?: boolean
+  $thresholdColor?: ThresholdColor
+}
+
+const { color } = selectors
+
+export const StyledHandle = styled.div<StyledHandleProps>`
+  left: var(--left);
```
jpveooys: Is this compatible with IE11?

jpveooys (commented):
Looks very nice.

One other small comment – even if they aren't meant to overlap, I think the labels should have a background colour set (instead of being transparent):

image

packages/react-component-library/src/components/RangeSliderE/RangeSliderE.stories.tsx
```diff
@@ -0,0 +1,171 @@
...
+          enabled: false,
+        },
+      ],
+    },
+  },
+}
+
+const StyledWrapper = styled.div`
+  display: flex;
+  height: 5rem'
```
jpveooys: May need to bump the top-padding for the canvas view:

image

packages/react-component-library/src/components/RangeSliderE/partials/StyledHandle.tsx
m7kvqbe1: Hmm looks like it needs a polyfill - I will investigate because this technique gives a nice performance improvement when dragging the handles in some environments.

packages/react-component-library/src/components/RangeSliderE/HandleE.tsx
```diff
@@ -41,7 +41,7 @@ export const HandleE: React.FC<HandleEProps> = ({
       aria-valuemin={min}
       aria-valuemax={max}
       aria-valuenow={value}
-      $left={`${percent}%`}
+      style={{ '--left': `${percent}%` }}
```
thyhjwb6: As this is a new pattern I am wondering how it compares to passing `$left` as typically we may want to steer away from using a `style` prop as that is the plan for other components.

My point is if it is favourable over `$left` then there might be some situations we want to use `style` which is not obviously documented in the code.

thyhjwb6 (commented):
Code looks good 👍 

Comment about the proposed `var` pattern

packages/react-component-library/src/components/RangeSliderE/HandleE.tsx
m7kvqbe1: Hmm yeah it's only being used here over the transient prop for performance gain.

Using the css variable means we don't destroy and create loads of new instances of the styled-component as the user drags the slider handle.

Maybe I should supplement it with a comment?

thyhjwb6: Perhaps that and/or updating [Contributing docs](https://github.com/defencedigital/mod-uk-design-system/blob/master/docs/contributing.md#styled-components) as there appears to be nuance.

k9dh3zij:
Thanks @m7kvqbe1, looking great! Minor style tweaks:

1. When at the lowest / highest value, the handle should be mid-aligned to the visual start / end of the scale. It's currently a few pixels off:
image
image

2. Clicking anywhere on the slider (causing the handle to move) or grabbing and moving the number input box should cause the slider handle to get the focus. Currently it only gets the focus if I grab the handle itself or tab to it.

https://user-images.githubusercontent.com/73531035/144046648-ca1a3f48-b091-4301-a1b9-b707cda755bb.mov

3. Clicking exactly on a tickmark doesn't cause the slider to move there (notice the second click in the video above).

4. Apologies this wasn't called out in the spec, but can you ensure a 12px padding between the icons and the edges of the slider track please?
image

packages/react-component-library/src/components/RangeSliderE/partials/StyledHandle.tsx
m7kvqbe1: Polyfill added to CRA template and Storybook.

jpveooys:
@m7kvqbe1 FYI the Chromatic build failed for IE11: 

image

m7kvqbe1:
Removed CSS variables optimization because IE11 polyfills do not work with styled-components.

k9dh3zij:
Thanks for the fixes @m7kvqbe1 - looking great! 👍 

I assume this is part of #2849 but just so it doesn't get lost:

Tickmarks with thresholds currently disappear behind the filled track.
Current:
image
Spec:
image

All earlier points have been addressed, thank you! :)

jpveooys (approved):
Nice!

PR #2848, opened 2021-11-29
Add clear button to `SelectE`

### LINKED ISSUE #2846: Add `SelectE` - clear button
Split out from #2438 

Adds a clear button to `SelectE`

thyhjwb6:
## Related issue
Closes #2846 

## Overview
Adds a clear button to `SelectE`.

## Link to preview
https://selecte-clear.netlify.app/?path=/story/select-experimental--default

## Reason
When an option has been selected then it may need to be cleared.

## Work carried out
- [x] Add clear button

## Notes
- Button sizes will be picked up in a [separate ticket](https://github.com/defencedigital/mod-uk-design-system/issues/2883).

### DIFF
--- a/packages/react-component-library/src/components/SelectE/SelectE.test.tsx
+++ b/packages/react-component-library/src/components/SelectE/SelectE.test.tsx
@@ -67,6 +67,10 @@ describe('SelectE', () => {
       expect(wrapper.getByTestId('select-input')).toHaveValue('')
     })
 
+    it('does not display a clear button', () => {
+      expect(wrapper.queryAllByTestId('select-clear-button')).toHaveLength(0)
+    })
+
     it('does not display the items', () => {
       expect(wrapper.queryAllByTestId('select-option')).toHaveLength(0)
     })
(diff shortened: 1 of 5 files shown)


packages/react-component-library/src/components/SelectE/SelectE.test.tsx
```diff
@@ -135,6 +139,20 @@ describe('SelectE', () => {
...
+          beforeEach(() => {
+            wrapper.getByTestId('select-clear-button').click()
+          })
+
+          it('resets the value', () => {
+            expect(wrapper.getByTestId('select-input')).toHaveValue('')
+          })
+
+          it('calls the `onChange` callback', () => {
+            expect(onChangeSpy).toHaveBeenCalledWith(undefined)
```
jpveooys: Should we be calling it with `undefined` or `null`?

(I think I'd say null as that is what you'd expect when it's serialised into JSON; undefined might get dropped.)

Would be good to update the type of `onChange` in `SelectEProps` too (currently `onChange?: (value: string) => void`, should be `value?: string` if we're using undefined, or `value: string | null` if null).

jpveooys (commented):
Looks good! Just that small comment.

jpveooys (approved):

k9dh3zij:
👍 Clear button all good, cheers!

m7kvqbe1 (approved):
LGTM

Issue #2863, opened 2021-12-01
DatePickerE: Preserve invalid typed dates


jpveooys:
If an invalid date is typed in the text input and the input is tabbed out of, the text entered should be preserved. (Currently it resets to the previous value.)

Also review intermittently failing test (currently skipped):

```
when `startDate` and `endDate` are updated externally
```

Split from #2439.

PR #2865, opened 2021-12-01
test(DatePickerE): Fix date mocking in tests

jpveooys:
## Overview 

This fixes some failing DatePickerE tests which were caused by the date mocking not working.

## Reason

Tests were failing on master.

## Work carried out

- [x] Replace mocking of `Date.now` with [`jest.setSystemTime`](https://jestjs.io/docs/next/jest-object#jestsetsystemtimenow-number--date) in DatePickerE tests
- [x] Skip another intermittently failing test (to be looked at as part of https://github.com/defencedigital/mod-uk-design-system/issues/2863)

### DIFF
--- a/packages/react-component-library/src/components/DatePickerE/DatePickerE.test.tsx
+++ b/packages/react-component-library/src/components/DatePickerE/DatePickerE.test.tsx
@@ -29,7 +29,6 @@ describe('DatePickerE', () => {
   let onChange: (data: { startDate: Date; endDate: Date }) => void
   let onBlur: (e: React.FormEvent) => void
   let onCalendarFocus: (e: React.SyntheticEvent) => void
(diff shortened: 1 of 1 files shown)


m7kvqbe1 (approved):
👍

m7kvqbe1:
Weird that it just suddenly stopped working? Is that because of a dependency bump?

jpveooys:
The tests that were relying on the mocked date were added recently – none of the older ones were (and today is a new month :grimacing: ). So the date mocking was probably not working for a while there.

There's another test that's being a flake unfortunately – I can't find an easy fix for that one so I've had to add a skip for it. I've added a note to #2863 to have a look at it as part of that (it may be related to how the input `value` is being updated at the moment).

m7kvqbe1 (approved):
👍

PR #2873, opened 2021-12-06
feat(RangeSliderE): Threshold and multiple handle changes

### LINKED ISSUE #2849: `RangeSliderE` threshold and multi-handle changes

m7kvqbe1:
## Related issue

Closes #2849

## Overview

Update multi-handle and threshold functionality in line with spec.

## Link to preview

https://61b1e2177c3a0a4600dcf9cf--rangeslidere.netlify.app

## Reason

>Implement new RangeSliderE component and reach feature parity with legacy.

## Work carried out

- [x] Update multiple handle behaviour
- [x] Style thresholds in line with spec
- [x] Fixed some typos in Storybook
- [x] Added ESLint override

## Screenshot

Screenshot 2021-12-09 at 08 36 31


packages/react-component-library/cypress/specs/RangeSliderE/index.spec.ts
```diff
@@ -1,6 +1,9 @@
-import { before, cy, describe, it } from 'local-cypress'
+/* eslint-disable jest/valid-expect */
```
jpveooys: Would be good to add this to the list here: https://github.com/defencedigital/mod-uk-design-system/blob/master/packages/react-component-library/.eslintrc.js#L23

jpveooys:
Generally looks good :+1: 

Is this intended?

range-slider-e-single

Also, have you seen this about the spacing between the value labels?

image

m7kvqbe1:
@jpveooys 

~~Hmm that first one is bug with the thresholds - investigating.~~

~~Second one i've missed - I will stick this back in WIP and see if it's feasible to implement.~~

packages/react-component-library/src/components/RangeSliderE/ThresholdTrackE.tsx
```diff
@@ -80,7 +80,7 @@ export const ThresholdTrackE: React.FC<ThresholdTrackEProps> = ({
       <ChunkE
         $thresholdColor={RANGE_SLIDER_TRACK_ABOVE_THRESHOLDS}
         $left={thresholds[thresholds.length - 1]}
-        $width={target.percent - thresholds[thresholds.length - 1]}
+        $width={Math.max(0, target.percent - thresholds[thresholds.length - 1])}
```
m7kvqbe1, line 77: Fixes issue where a single threshold would remain visible when not within active range:

https://user-images.githubusercontent.com/66470099/144870716-90cf4491-aedd-4321-b6e7-3fdd740d159c.gif

packages/react-component-library/src/components/RangeSliderE/TickE.tsx
```diff
@@ -14,10 +14,25 @@ export interface TickEProps {
...
+  values: ReadonlyArray<number>,
+  tickValue: number,
+  tracksLeft: boolean,
+  tracksRight: boolean
+): boolean {
+  if (!tracksLeft && !tracksRight) {
+    return tickValue >= values[0] && tickValue <= values[1]
+  }
+
+  if ((tracksLeft && !tracksRight) || (tracksRight && !tracksLeft)) {
```
jpveooys: It might be nice to do 

```
if (tracksLeft && tracksRight) {
  return true
}
```

first to make the current second case simpler.

jpveooys (approved):

thyhjwb6 (approved):
👍

k9dh3zij:
Yay, RangeSlider - thanks @m7kvqbe1 :) 

One piece of feedback and one request please:

1. Please note the spec when thresholds are present and the handle is to the left of a threshold:
i. threshold warning icon and 'notch' should not disappear
ii. track to the right of the threshold should maintain its colour

Spec:
image

Current behaviour:

https://user-images.githubusercontent.com/73531035/145051646-778136d4-2c13-4049-9de0-ce93bb4b9494.mov

2. Multiple handles now working fine, thank you. Now I've had a chance to look at how this works irl, I was wondering if it's possible to bring the label for the focussed handle in the foreground? So, in the example below, if I grab (or tab to) the LHS handle, is it possible to change the z-index of its label to show over the RHS handle label? (And vice-versa for the RHS handle, of course, if I then tab to it)
image

Thanks!! ☺️

m7kvqbe1:
@k9dh3zij fixups added for both of the above requests

k9dh3zij:
Thank you so much for making those changes @m7kvqbe1 I think the two handles work much better now!

One minor visual tweak:

Please note the track mark warning z-index - the filled track shouldn't cover the marks.
Also, if possible to make the mark edges round that would be great, but not an issue if not possible. Thanks :) 

Current:
image

image

Spec:
image
image

k9dh3zij:
👍 Beautiful, if I may say so myself, thank you! 😁

PR #2896, opened 2021-12-08
fix(Timeline): Update scale index condition

### LINKED ISSUE #2765: Full-width timeline zooms out on viewport width increase
### Steps to reproduce

1. Go to https://storybook.design-system.digital.mod.uk/?path=/story/compound-timeline--full-width (canvas tab)
2. Decrease the viewport width
3. Zoom in the timeline fully
4. Increase the viewport width

The timeline resets to the month zoom level.

### Demo

timeline-zoom

thyhjwb6:
## Related issue
Closes #2765 

## Overview
There is a bug when zoomed all the way in and resizing the window.

It is possible for the scale index to be 0 which is falsy but we still want to satisfy the `true` side of the condition.

## Work carried out
- [x] Fix bug with test

### DIFF
--- a/packages/react-component-library/cypress/specs/timeline/index.spec.ts
+++ b/packages/react-component-library/cypress/specs/timeline/index.spec.ts
@@ -264,6 +264,18 @@ describe('Compound Timeline', () => {
         it('should fill the width', () => {
           cy.get(selectors.timeline.month).should('have.css', 'width', '1023px')
         })
+
+        describe('resizing the window', () => {
+          before(() => {
+            cy.viewport(1500, 768)
+          })
+
+          it('still renders 1 day', () => {
+            cy.get(selectors.timeline.day)
+              .should('have.length', 1)
+              .should('have.css', 'width', '1499px')
+          })
+        })
       })
     })
 
--- a/packages/react-component-library/src/components/Timeline/context/reducer.ts
+++ b/packages/react-component-library/src/components/Timeline/context/reducer.ts
@@ -106,7 +106,7 @@ export function reducer(
     case TIMELINE_ACTIONS.INITIALISE: {
       const scaleOptions = buildScaleOptions(state.options, action.width)
       const currentScaleIndex =
-        state.currentScaleIndex ||
+        state.currentScaleIndex ??
         scaleOptions.findIndex(({ isDefault }) => isDefault)
       return {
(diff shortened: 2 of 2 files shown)


packages/react-component-library/src/components/Timeline/context/reducer.ts
```diff
@@ -105,9 +105,9 @@ export function reducer(
     case TIMELINE_ACTIONS.CHANGE_WIDTH:
     case TIMELINE_ACTIONS.INITIALISE: {
       const scaleOptions = buildScaleOptions(state.options, action.width)
-      const currentScaleIndex =
-        state.currentScaleIndex ||
-        scaleOptions.findIndex(({ isDefault }) => isDefault)
+      const currentScaleIndex = isNumber(state.currentScaleIndex)
+        ? state.currentScaleIndex
+        : scaleOptions.findIndex(({ isDefault }) => isDefault)
```
jpveooys: Could consider:

```suggestion
      const currentScaleIndex = state.currentScaleIndex ?? scaleOptions.findIndex(({ isDefault }) => isDefault)
```

packages/react-component-library/src/components/Timeline/context/reducer.ts
```diff
@@ -3,7 +3,7 @@ import {
   eachMonthOfInterval,
   eachWeekOfInterval,
 } from 'date-fns'
-import { padStart } from 'lodash'
+import { padStart, isNumber } from 'lodash'
```
jpveooys: Would be nice to sort `isNumber` and `padStart`:

```suggestion
import { isNumber, padStart } from 'lodash'
```

packages/react-component-library/cypress/specs/timeline/index.spec.ts
```diff
@@ -264,6 +264,18 @@ describe('Compound Timeline', () => {
         it('should fill the width', () => {
           cy.get(selectors.timeline.month).should('have.css', 'width', '1023px')
         })
+
+        describe('resize window', () => {
```
jpveooys: Would be good to make this consistent with the other `describe`s – perhaps something like:

```suggestion
        describe('resizing the window', () => {
```

jpveooys (approved):
Nice find :+1:

PR #2897, opened 2021-12-08
feat(DatePickerE): Preserve invalid typed dates

### LINKED ISSUE #2863: DatePickerE: Preserve invalid typed dates

jpveooys:
## Related issue

Resolves #2863

## Overview

This makes a few changes:

- makes the date picker calendar close when the input is focused
- removes the select all behaviour when focusing the input when the mouse
- makes the input preserve invalid dates when they are typed
- makes the input preserve disabled dates when they are typed
- notifies the hosting component when invalid dates are typed
- fixes a problem where pasting a date into the field didn't work
- makes it possible to clear the date by deleting it
- prioritises the styling the input with the blue focus border over the red error border when the input is focused

## Link to preview

[Storybook](https://5e25c277526d380020b5e418-knbibnafih.chromatic.com/?path=/docs/date-picker-experimental--custom-format)

## Reason

> - makes the date picker calendar close when the input is focused

This is to be in line with https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html

> - removes the select all behaviour when focusing the input when the mouse

This was making it difficult to select a specific part of the date to correct using the mouse.

> - makes the input preserve invalid dates when they are typed

This is so mistakes can be easily corrected.

> - makes the input preserve disabled dates when they are typed

This is so the app can display an appropriate error message in line with our guidance:

image

> - notifies the hosting component when invalid dates are typed

Previously, this was not done, but this could result in the hosting component having a date value different to what's being displayed, and hence the wrong value being submitted in the case of a form.

> - fixes a problem where pasting a date into the input didn't work

Previously a pasted date was ignored.

> - makes it possible to clear the date by deleting it

Previously it was restoring the last value.

> - prioritises the styling the input with the blue focus border over the red error border when the input is focused

This is for consistency with other v3 components.

## Work carried out

- [x] Close picker when input focused
- [x] Don't select all text on mouse focus
- [x] Rework code that handles typed dates
- [x] Remove incorrect awaits from tests

### Before

picker-e-type-before

### After

picker-e-type-after

## Developer notes

The value used for an invalid date is `new Date(NaN)`. This was picked as it's what `date-fns` and the `Date` constructor use for invalid inputs. (The `isValid` function from `date-fns` can be used to check if a date object is valid.)

I've also added additional properties to the `onChange` callback to flag this and also when a disabled date has been typed.

The flakey skipped test mentioend in #2863 is still a problem unfortunately – I've split it out to #2898.


k9dh3zij:
Works like a treat, thank you!

Small visual niggle:

- Button border styling appears somewhat broken:
 image

jpveooys:
Just making some changes to the `onChange` callback based on feedback from @k9dh3zij.

jpveooys:
@k9dh3zij Thanks! I've had a look and this is affecting `NumberInputE` too. I'll add a note to https://github.com/defencedigital/mod-uk-design-system/issues/2889 as it's related I believe.

jpveooys:
I've updated this to add two additional properties (`startDateValidity` and `endDateValidity`) to `onChange` to indicate if a disabled or invalid date has been typed.

packages/react-component-library/src/components/DatePickerE/useInput.ts
```diff
@@ -0,0 +1,78 @@
...
+
+      if (isValid(from)) {
+        setInputValue(formatDatesForInput(from, null, datePickerFormat))
+      }
+    },
+    [datePickerFormat, from, isRange, setInputValue]
+  )
+
+  const handleInputBlur = useCallback(() => {
+    if (isRange) {
```
thyhjwb6, line 46: If this condition is removed then tests still pass. Is it possible to cover?

thyhjwb6 (approved):
Looks great 👍

packages/react-component-library/src/components/DatePickerE/useInput.ts
jpveooys: Yep – I've added a few additional tests.

packages/react-component-library/src/components/DatePickerE/DatePickerE.tsx
```diff
@@ -120,6 +139,9 @@ export const DatePickerE: React.FC<DatePickerEProps> = ({
   initialMonth,
   placement = 'bottom-start',
   onBlur,
+  // Formik can pass value – drop it to stop it being forwarded to the input
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
m7kvqbe1: Is it worth disabling this lint rule? Or maybe there is a option to ignore `_`.  I've also used this elsewhere.

packages/react-component-library/src/components/DatePickerE/useInput.ts
```diff
@@ -0,0 +1,78 @@
...
+  const handleInputChange = useCallback(
+    (event) => {
+      if (isRange) {
+        return
+      }
+
+      setInputValue(event.currentTarget.value)
+      const date = parseDate(datePickerFormat, event.currentTarget.value)
+      handleDayClick(date)
+      if (isValid(date)) {
```
m7kvqbe1, line 67: Could we add a line break here?

m7kvqbe1 (commented):
LGTM just a couple of tiny comments

packages/react-component-library/src/components/DatePickerE/DatePickerE.tsx
jpveooys: Yep I agree on on the latter – it should support patterns, let me check if it can be set to ignore variables starting with `_`.

m7kvqbe1 (approved):
👍

PR #2899, opened 2021-12-09
fix(NumberInputE): Sanitize input before invoking `parseInt`

### LINKED ISSUE #2767: NumberInput: Digits deleted when typing an alpha character
Typing an alpha character is not accepted - that's great.

However, if the cursor is not at the end of the number when I type an alpha character, all digits to the right of the cursor are deleted. It shouldn't delete any digits.

https://user-images.githubusercontent.com/73531035/139659223-6fe1ce21-7d3a-4633-83ba-c38187c3ac8b.mov

m7kvqbe1:
## Related issue

Closes #2767

## Overview

Sanitize input before parsing. You can now only type in numeric characters.

## Reason

>Typing an alpha character would cause everything after that character to be purged by `parseInt`.

## Work carried out

- [x] Supplement with additional tests
- [x] Sanitize input before parsing

### DIFF
--- a/packages/react-component-library/src/components/NumberInputE/NumberInputE.test.tsx
+++ b/packages/react-component-library/src/components/NumberInputE/NumberInputE.test.tsx
@@ -173,12 +173,12 @@ describe('NumberInputE', () => {
     })
 
     describe('and the user types values', () => {
-      beforeEach(async () => {
+      beforeEach(() => {
         const input = wrapper.getByTestId('number-input-input')
 
-        await userEvent.type(input, '1')
-        await userEvent.type(input, '2')
-        await userEvent.type(input, '3')
+        userEvent.type(input, '1')
+        userEvent.type(input, '2')
+        userEvent.type(input, '3')
       })
 
       it('calls the `onChange` callback once with `1`', () => {
@@ -193,7 +193,7 @@ describe('NumberInputE', () => {
       assertOnChangeCall(123, 3)
     })
 
-    describe('and the user types a value with an invalid character', () => {
+    describe('and the user types a value with invalid characters', () => {
       beforeEach(() => {
         const input = wrapper.getByTestId('number-input-input')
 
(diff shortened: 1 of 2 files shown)


packages/react-component-library/src/components/NumberInputE/NumberInputE.test.tsx
```diff
@@ -193,6 +193,31 @@ describe('NumberInputE', () => {
       assertOnChangeCall(123, 3)
     })
 
+    describe('and the user types a value with alpha characters', () => {
+      beforeEach(async () => {
+        const input = wrapper.getByTestId('number-input-input')
+
+        await userEvent.type(input, '1')
```
jpveooys: FYI `userEvent.type()` is typed to only return a promise if a delay is passed.

https://github.com/testing-library/user-event/blob/57a6ea4d4808bb5c07817089c437929ba9cdb070/src/type/index.ts#L4-L13

Is VS code wanting an `await` there?

packages/react-component-library/src/components/NumberInputE/NumberInputE.test.tsx
```diff
@@ -343,7 +368,7 @@ describe('NumberInputE', () => {
           })
         })
 
-        assertInputValue('1')
```
jpveooys, line 346: Might be nice to update the describe() name from `and the user types an invalid character` as the entire value is being replaced with `a` rather than an `a` being typed (confused me for a bit :smile: )

packages/react-component-library/src/components/NumberInputE/NumberInputE.test.tsx
m7kvqbe1: I was just following the pattern used in other tests and assumed it always needed to be awaited - I've updated this for all relevant tests now.

jpveooys (approved):
:+1:

jpveooys:
Just spotted that it still deletes characters if `.` is typed:

number

m7kvqbe1:
@jpveooys 

Ooops my regex was a little bit lenient - i'll disallow that character for now there is another ticket to support floats.

jpveooys (approved):

thyhjwb6 (approved):
👍

PR #2901, opened 2021-12-09
fix(Fieldset): Improve styling in IE11

### LINKED ISSUE #2847: Fix CheckboxE and RadioE fieldset styling issues in IE11
https://github.com/defencedigital/mod-uk-design-system/pull/2839

See comments re: IE11 in this PR

jpveooys:
## Related issue

Resolves #2847

## Overview

This fixes the main styling problems in IE11 when `CheckboxE` and `RadioE` are used in a `Fieldset` or a `FormikGroupE`.

## Link to preview

[Storybook](https://5e25c277526d380020b5e418-gdbyovpcdf.chromatic.com/?path=/docs/checkbox-experimental--with-formik-group)

## Reason

The styling was broken in IE11.

## Work carried out

- [x] Replace uses of `unset` with explicit values

#### Before

cbe-before

#### After

checkbox-e-after

#### Before

radio-e-before

#### After

radio-e-after

## Developer notes

These were happening because IE doesn't support the `unset` keyword (nor `initial` or `all`).

It does support `inherit` but that doesn't make sense for many properties.

I've changed replaced these usages of `unset` with what the initial values are according to MDN.

Note that we have other uses of `unset` etc. around the code base which may need to be reviewed separately.

### DIFF
--- a/packages/react-component-library/src/components/Fieldset/Fieldset.tsx
+++ b/packages/react-component-library/src/components/Fieldset/Fieldset.tsx
@@ -21,7 +21,7 @@ const { color, spacing } = selectors
 const StyledFieldset = styled.fieldset<StyledFieldsetProps>`
   display: inline-block;
   position: relative;
-  padding: unset;
+  padding: 0;
   border: none;
   border-radius: 15px;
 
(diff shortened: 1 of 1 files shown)


m7kvqbe1 (approved):
Awesome! I thought this was going to be horrible!

PR #2902, opened 2021-12-10
Make experimental component icons consistent

### LINKED ISSUE #2883: Make button sizes consistent
Required across all v3 components with an inline button eg `SelectE`, `NumberInputE`, the button size needs to be consistent with the design spec.

Also a good idea to refactor code while there.

image

thyhjwb6:
## Related issue
Fixes #2883 

## Overview
Inline buttons are being used across several components so the CSS can be reused.
Icons need to be consistently sized.

## Link to preview
https://inlinebutton-icon.netlify.app/

## Work carried out
- [x] Refactor code
- [x] Update icons
- [x] Fix separator

## Affected components
- `DatePickerE`
- `NumberInputE`
- `SelectE`
- `TextInputE`


packages/react-component-library/src/components/SelectE/partials/StyledInlineButtons.tsx
```diff
@@ -1,15 +1,7 @@
...
-  display: flex;
-  justify-content: center;
-  text-align: center;
-  align-items: center;
-
-  ${StyledInlineButton} + ${StyledInlineButton} {
-    margin-left: 0;
-  }
+export const StyledInlineButtons = styled(StyledInlineButtonsBase)`
+  border-left: unset;
```
jpveooys: `unset` doesn't work with IE unfortunately – so may have to give this an explicit value.

m7kvqbe1 (commented):
~~This also needs updating for `SwitchE` I think.~~

Ah sorry - I'm thinking about the `COMPONENT_SIZE` thing.

jpveooys (commented):
Code changes look good overall!

Couple of other comments:

- are some of these too big now (comparing to material icons 18dp)? Will leave to @k9dh3zij to confirm :sweat_smile: 
- there's the dead space around the icon that you can't click on and without the hover effect (this was mostly OK for date picker before so you could check what it was doing)

thyhjwb6:
Thank you for the feedback.

https://inlinebutton-icon.netlify.app/?path=/story/select-experimental--default

packages/react-component-library/src/components/DatePickerE/DatePickerE.tsx
```diff
@@ -24,6 +23,7 @@ import { useRangeHoverOrFocusDate } from './useRangeHoverOrFocusDate'
 import { useSelection } from './useSelection'
 import { useStatefulRef } from '../../hooks/useStatefulRef'
 import { WEEKDAY_TITLES } from './constants'
+import { InlineButton } from '../InlineButtons/InlineButton'
```
jpveooys: I'm never sure what exact rules we're using to sort import :smile: This one does look out of place though.

jpveooys (commented):
As discussed, I think the icons are a tad bit too small now (sorry :grimacing: )

Would also check the hover background size (I had it at 28x28px).

thyhjwb6:
Updates deployed https://inlinebutton-icon.netlify.app/?path=/story/select-experimental--default

jpveooys (approved):
:+1:

m7kvqbe1 (approved):

k9dh3zij:
👍 Perfect, thanks!

PR #2914, opened 2021-12-14
Make Formik a peer dependency and fix CRA template

### LINKED ISSUE #2769: Review how Formik should be consumed by downstream apps
Currently Formik is required by the `FormikGroup` component:

https://github.com/defencedigital/mod-uk-design-system/blob/6ac4e5f17cdfc79bad90ebc2bf64808c118653ef/packages/react-component-library/src/components/FormikGroup/FormikGroup.tsx#L4

This means imports of the React component library fail without Formik being installed. This includes the current CRA template. However, `formik` isn't a peer dependency and its requirement also isn't documented.

Formik could be made either a required or [an optional peer dependency](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#peerdependenciesmeta). The latter may make more sense, but would mean that Formik-related bits would need to be moved either to a separate package or a separate export (not the main index.ts in the component library).

The CRA template may also need updating: https://github.com/defencedigital/mod-uk-design-system/blob/master/packages/cra-template-defencedigital/template.json

jpveooys:
## Related issue

Resolves #2769 

## Overview

This:

- makes Formik a peer dependency and documents this
- adds `formik` to the CRA template
- fixes problems with the lint command in the CRA template
- fixes problems with the smoke test that hid these problems
- removes the dependency on `@defencedigital/css-framework ` from the CRA template

## Reason

To:

- make sure the documentation is accurate
- make sure the CRA template works and is current
- make sure the smoke test catches problems

## Work carried out

- [x] Move the test app in the smoke test to a different directory
- [x] Add formik as a peer dependency and document this
- [x] Fix the lint command in the CRA template and make sure this is tested
- [x] Remove the `@defencedigital/css-framework` dependency from the CRA template

## Developer notes

The main problem with the smoke test was that it was creating the test app within the monorepo, and normal Node.js import resolution means that the monorepo's `node_modules` directory was also being used (and hence formik was imported from there).

The fix used was to create the test app somewhere else (in the temp directory). (I was also going to put a `lerna clean` in, but didn't as it would probably interfere with the `node_modules` caching.)

I've created #2915 as a separate issue for making Formik optional. (I haven't labelled it as next major but it can be added if we think it should be.)

### DIFF
--- a/.github/workflows/npmsmoketest.yml
+++ b/.github/workflows/npmsmoketest.yml
@@ -44,19 +44,18 @@ jobs:
           )
           for i in "${PACKAGES[@]}";
            do
-             contents="$(jq '.package.dependencies."@defencedigital/'$i'" = "file:../packages/'$i'"' $TEMPLATE)" \
+             contents="$(jq '.package.dependencies."@defencedigital/'$i'" = "file:'$GITHUB_WORKSPACE'/packages/'$i'"' $TEMPLATE)" \
              && echo "${contents}" > $TEMPLATE
            done
           cat $TEMPLATE
 
       - name: Create boilerplate app
         run: |
-          if [ -d "my-app" ]; then rm -Rf my-app ; fi
-          npx create-react-app my-app --template file:../${{ github.event.repository.name }}/packages/cra-template-defencedigital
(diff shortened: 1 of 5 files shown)


m7kvqbe1 (approved):
LGTM

PR #2919, opened 2021-12-15
Add ability to type and jump to item to `SelectE`

### LINKED ISSUE #2885: Add ability to type and jump to item on `SelectE`
As highlighted in [review](https://github.com/defencedigital/mod-uk-design-system/pull/2838#issuecomment-982557303) the `SelectE` needs to behave like a `select` so that when a key is pressed, the item that begins with that item is selected.

thyhjwb6:
## Related issue
Closes #2885 

## Overview
Refactors `SelectE` to use `useSelect` and now it is possible to type and jump to an option.

## Link to preview
https://selecte-typing.netlify.app/?path=/story/select-experimental--default

## Reason
This is normal behaviour for a `select`.

## Work carried out
- [x] Refactor stories
- [x] Refactor `SelectE`

### DIFF
--- a/packages/react-component-library/src/components/SelectE/SelectE.tsx
+++ b/packages/react-component-library/src/components/SelectE/SelectE.tsx
@@ -1,5 +1,5 @@
 import React, { useState } from 'react'
-import Downshift from 'downshift'
+import { useSelect } from 'downshift'
 
 import { ArrowButton } from './ArrowButton'
 import { ClearButton } from './ClearButton'
@@ -29,7 +29,7 @@ export interface SelectEProps extends ComponentWithClass {
   /**
    * Collection of options to display within the Select.
    */
-  children: SelectChildType | SelectChildType[]
+  children: SelectChildrenType
   /**
    * Optional HTML `id` attribute to apply to the component.
    */
@@ -56,21 +56,6 @@ export interface SelectEProps extends ComponentWithClass {
   value?: string
 }
 
-function getChildBy(
-  children: SelectChildrenType,
-  propName: string,
-  value: string
-): React.ReactNode {
-  return React.Children.toArray(children).find((child: React.ReactNode) => {
-    return React.isValidElement(child) && child.props[propName] === value
-  })
-}
-
-function getSelectedValue(children: SelectChildrenType, value: string) {
-  const child = getChildBy(children, 'value', value)
-  return (child as React.ReactElement)?.props?.children
-}
-
 export const SelectE: React.FC<SelectEProps> = ({
   children,
   id = getId('select'),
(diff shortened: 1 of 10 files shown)


k9dh3zij:
Thanks @thyhjwb6, it behaves very well - just a couple of observations:

Comparing to the vanilla select [here](https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_select):

1. The timeout between keystrokes for deciding whether the user is still typing a single word or whether they've started typing a new word seems to be on the small side - not sure if there's a universally accepted 'default' and whether we should increase ours a bit?

2. If I press 'T' [wait] then press 'T' again it goes through all options starting with 'T', which is inconsistent with the vanilla Select behaviour (which sticks to the first option in the list). I think we should aim for consistency, if possible please.

thyhjwb6:
@k9dh3zij response as follows:

1. I can look at the timing although this is out-of-the-box from a widely used library [Downshift](https://www.npmjs.com/package/downshift)
2. This behaviour is consistent with how [an example on W3](https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html) behaves. 

If you use the following [example on Mozilla,](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select) you get stranger behaviour:

```
<select name="pets" id="pet-select">
    <option value="">--Please choose an option--</option>
    <option value="dog">Dog</option>
    <option value="cat">Cat</option>
    <option value="hamster">Hamster</option>
    <option value="parrot">Parrot</option>
    <option value="parrot">Parrot 2</option>
    <option value="spider">Spider</option>
    <option value="goldfish">Goldfish</option>
</select>
```

Also try using this example on your W3 schools link:

```
  <select name="cars" id="cars">
    <option value="volvo">Volvo</option>
    <option value="saab">Saab</option>
    <option value="sbab">Sbab</option>
    <option value="opel">Opel</option>
    <option value="audi">Audi</option>
  </select>
```

k9dh3zij:
@thyhjwb6 Ha! That example on Mozilla broken, it randomly went through to Spider for some reason...

So as it seems this behaviour is consistent with the W3 recommended behaviour (and so it's the vanilla Select that's inconsistent), let's keep as is. The timing was just an observation and I don't have a way of reliably measuring it, so let's not worry unless we get feedback that it's confusing.

👍 from me, thank you! :)

packages/react-component-library/src/components/SelectE/SelectEOption.tsx
```diff
@@ -14,30 +14,29 @@ export interface SelectEOptionProps extends ComponentWithClass {
...
-      >
-        {badge}
-      </StyledOptionBadge>
-    )}
-  </StyledOption>
-)
+export const SelectEOption = React.forwardRef<
+  HTMLLIElement,
+  SelectEOptionProps
+>(({ badge, icon, children, isHighlighted, ...rest }, ref) => {
```
jpveooys, line 20: Hmm `ref` isn't used here. Is it meant to be forwarded, or is the forwardRef unnecessary?

jpveooys (approved):
Looks good to me! I noticed it still triggers the Storybook shortcuts when typing – is there anything that can be done about that?

packages/react-component-library/src/components/SelectE/SelectE.tsx
```diff
@@ -83,125 +68,114 @@ export const SelectE: React.FC<SelectEProps> = ({
...
+    getToggleButtonProps,
+    highlightedIndex,
+    isOpen,
+    openMenu,
+    selectItem,
+    selectedItem,
+    toggleMenu,
+  } = useSelect({
+    itemToString,
+    initialSelectedItem: React.Children.toArray(children).find((child) => {
```
m7kvqbe1, line 87: Is it worth breaking some of this stuff out into functions - this code feels a little 'cramped'.

m7kvqbe1 (changes requested):
Noticed same behaviour as Anthony but code looks good - just a small comment.

jpveooys:
@m7kvqbe1 That looks like design tokens needs to be rebuilt to me.

m7kvqbe1 (approved):

PR #2924, opened 2021-12-20
feat(NumberInputE): Add float step support

### LINKED ISSUE #1128: Add support for floating point numbers to `NumberInputE`

m7kvqbe1:
## Related issue

Closes #1128

## Overview

Add support for stepped floats to `NumberInputE`.

## Link to preview

https://61c1ab613913be2c341df66a--numberinputefloats.netlify.app

## Reason

Component would not work correctly with float steps.

## Work carried out

- [x] Fix issue where validation was not invoked on step
- [x] Handle floating point steps
- [x] Add ability to type in a floating point number

## Developer notes

If we let users enter typed input and there is a step, should we be validating their typed input? 

Not doing so allows them to circumvent the step.

- If component is stepped:
    - Valid typed user input is:
        - Numbers that are a result of X increments/decrements of step (from initial value) / modulus 0
        - e.g. Typed Input: 2.25, Initial Value: 1, Step: 0.25, Increments: 5 / (2.25 % 0.25 === 0)

### DIFF
--- a/packages/react-component-library/src/components/NumberInputE/Input.tsx
+++ b/packages/react-component-library/src/components/NumberInputE/Input.tsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import { isFinite, isNil } from 'lodash'
+import { isNil } from 'lodash'
 
 import { ComponentSizeType } from '../Forms'
 import { InputValidationProps } from '../../common/InputValidationProps'
@@ -18,7 +18,7 @@ export interface InputProps extends InputValidationProps {
   onFocus: (event: React.FormEvent<HTMLInputElement>) => void
   placeholder?: string
   size: ComponentSizeType
-  value?: number
+  value?: string
 }
 
 export const Input: React.FC<InputProps> = ({
@@ -26,7 +26,6 @@ export const Input: React.FC<InputProps> = ({
   isDisabled,
   id,
   label,
-  placeholder = '',
   size,
   value,
   ...rest
@@ -52,7 +51,7 @@ export const Input: React.FC<InputProps> = ({
         data-testid="number-input-input"
         disabled={isDisabled}
         id={id}
-        value={isFinite(value) ? value : ''}
+        value={value || ''}
         {...rest}
       />
     </StyledInputWrapper>
--- a/packages/react-component-library/src/components/NumberInputE/Buttons.tsx
+++ b/packages/react-component-library/src/components/NumberInputE/Buttons.tsx
@@ -7,6 +7,7 @@ import { InlineButton } from '../InlineButtons/InlineButton'
 import { NUMBER_INPUT_BUTTON_TYPE } from './constants'
 import { StyledDivider } from './partials/StyledDivider'
 import { StyledInlineButtons } from '../InlineButtons/partials/StyledInlineButtons'
+import { countDecimals } from '../../helpers'
 
 export interface ButtonsProps {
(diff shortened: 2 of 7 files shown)


jpveooys:
Hmm I'm still unable to type a number with a decimal point though (or do we have a separate issue for that?)

m7kvqbe1:
> Hmm I'm still unable to type a number with a decimal point though (or do we have a separate issue for that?)

Ah ok interesting it lets you add the decimal point in to an existing number but won't accept that character if it's the last character (e.g. `1.`). I am investigating this now.

packages/react-component-library/src/components/NumberInputE/Input.tsx
```diff
@@ -52,7 +52,7 @@ export const Input: React.FC<InputProps> = ({
         data-testid="number-input-input"
         disabled={isDisabled}
         id={id}
-        value={isFinite(value) ? value : ''}
+        value={value || placeholder || ''}
```
jpveooys: Hmm it seems a bit odd to pass a `placeholder` to `value` (also for accessibility reasons). Also, the placeholder can't be deleted using the backspace key.

What about using the `placeholder` attribute with custom styling if needed?

jpveooys (commented):
There seems to be a regression when the input is cleared with the backspace key – it's calling onChange with `NaN` (previously it was `null`):

image

packages/react-component-library/src/components/NumberInputE/Input.tsx
m7kvqbe1: Ooh yes much better - I forgot about that native attribute.

jpveooys (approved):
:+1:

k9dh3zij:
1. If there's already a decimal point, can we stop the user from typing a second one? (as we're already restricting them from typing any 'invalid' characters)

2. There seems to be some unexpected rounding going on:

Example:

Expected behaviour (with a step of 0.25):
- Type 0.9
- Step up; it's now 1.15 - it added 0.25 and the result is 1.15 - great!

Unexpected behaviour:
- Type 0.009
- Step up; it's now 0.26 - it added 0.25 and rounded up to 0.26
- I expected the result to be 0.0259

3.
> If we let users enter typed input and there is a step, should we be validating their typed input?
> Not doing so allows them to circumvent the step.

Good question. I think this should be an option. 
Looking at [this](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/step) : "_the value must be a positive number - integer or float -- or the special value any, which means no stepping is implied, and any value is allowed_". 

I tried 'any' and it defaulted the step to 1 while accepting floating values. 
However, I don't see why someone shouldn't be able to define a step of (say) 100, but accept typed integers which are not multiples of 100. This feels like bad UX (eg. if valid inputs range from 10,000 to 50,000 we don't want a step of 1, but may want to accept all integers between 10,000 and 50,000 as valid inputs).

m7kvqbe1:
The following tickets have been created off the back of @k9dh3zij feedback:

https://github.com/defencedigital/mod-uk-design-system/issues/2948
https://github.com/defencedigital/mod-uk-design-system/issues/2949
https://github.com/defencedigital/mod-uk-design-system/issues/2950

PR #2923, opened 2021-12-17
Resolve security alerts and update SVGR

jpveooys:
## Related issue

Resolves #2816

## Overview

This:

- removes unneeded package resolutions
- updates dependencies and adds new resolutions to resolve security alerts
- updates SVGR to version 6.1.2

## Link to preview

[Storybook](https://5e25c277526d380020b5e418-dodgigxxqi.chromatic.com/)

## Reason

Make sure old and/or unneeded resolutions are removed, and to address GitHub Dependabot security alerts.

## Work carried out

- [x] Remove old resolutions
- [x] Add new resolutions
- [x] Update SVGR

## Developer notes

There are some minor differences in the optimised SVGs, but these shouldn't be anything significant.

@m7kvqbe1 I remember you had some problems previously running tests when things in the lock file were updated – do you want to check again on this branch?

PR #2930, opened 2021-12-21
Resolve security alerts, update SVGR and improve icon library build time

### LINKED ISSUE #2816: Review package resolutions
### Description

Review whether any of the resolutions in `package.json` can now be removed. These are candidates from a quick look:

- jszip
- url-parse
- ws
- normalize-url
- css-what
- postcss
- tar

Note that in some cases, patches have been released for older versions of the library therefore they may be update-able to a good version without using a resolution.

jpveooys:
## Related issue

Resolves #2816 and #2932

## Overview

This:

- removes unneeded package resolutions
- updates dependencies and adds new resolutions to resolve security alerts
- updates SVGR to version 6.1.2
- removes @svgr/webpack
- optimises the icon library build to reduce the build time
- merges the browserslist configurations

This includes the changes that were in #2923 but reverted while the icon library build performance problems were investigated (which turned out to be unrelated).

## Link to preview

https://5e25c277526d380020b5e418-twjvgnuyxr.chromatic.com/

## Reason

- make sure old and/or unneeded resolutions are removed
- to address GitHub Dependabot security alerts
- as there are no longer any SVG files being imported in the component library
- the icon library build was taking way too long, and the bundle size had also increased significantly
- there were multiple inconsistent browserslist configurations

## Work carried out

- [x] Remove old resolutions
- [x] Add new resolutions
- [x] Update SVGR
- [x] Remove `@svgr/webpack`
- [x] Optimise icon library build
- [x] Unify browserslist configuration

## Developer notes

There are some minor differences in the optimised SVGs, but these shouldn't be anything significant.

It looks like `@svgr/webpack` was added in #34 for #39, but that is long gone.

The icon library build time was improved by:

- making use of @babel/plugin-transform-runtime to each individual icon component being bloated with Babel helpers (see [this](https://babeljs.io/docs/en/babel-plugin-transform-runtime) and [this](https://webpack.js.org/loaders/babel-loader/#babel-is-injecting-helpers-into-each-file-and-bloating-my-code) for more detail)
- moving the width, height and fill attribute manipulation to SVGR itself (only a small benefit, but this avoids them running twice for the ES and CJS builds)
- turning off a Terser option that was mega slow and had negligible benefit

These changes also reduce the size of both the ES and CJS icon library builds.

Note that there is probably scope to apply similar optimisations to the other packages, but that hasn't been tackled for now.

#### Before (master)

```
$ BABEL_ENV=es babel src --out-dir "dist/es" --extensions ".ts,.tsx"
Successfully compiled 1616 files with Babel (12092ms).
$ BABEL_ENV=cjs webpack --mode production --config=webpack/prod.js
asset index.js 1.99 MiB [emitted] [minimized] [big] (name: main) 1 related asset
orphan modules 3.47 MiB [orphan] 1616 modules
runtime modules 937 bytes 4 modules
./index.ts + 1616 modules 3.57 MiB [not cacheable] [built] [code generated]
webpack 5.65.0 compiled successfully in 100393 ms
```

#### After

```
$ BABEL_ENV=es babel src --out-dir "dist/es" --extensions ".ts,.tsx"
Successfully compiled 1616 files with Babel (9046ms).
$ BABEL_ENV=cjs webpack --mode production --config=webpack/prod.js
asset index.js 1.1 MiB [emitted] [minimized] [big] (name: main) 1 related asset
orphan modules 1.97 MiB [orphan] 1616 modules
runtime modules 937 bytes 4 modules
built modules 2.08 MiB [built]
  ./index.ts + 1616 modules 2.08 MiB [not cacheable] [built] [code generated]
  ../../../node_modules/@babel/runtime/helpers/extends.js 582 bytes [built] [code generated]
  ../../../node_modules/@babel/runtime/helpers/objectWithoutPropertiesLoose.js 462 bytes [built] [code generated]
webpack 5.65.0 compiled successfully in 16524 ms
```


packages/icon-library/.svgrrc.js
```diff
@@ -1,23 +1,29 @@
...
-      import React from 'react'
-      import { SVGUniqueID } from 'react-svg-unique-id';
-      import { SVGIconProps } from '../types'
+const template = (variables, { tpl }) => {
+  return tpl`
+import React from 'react'
+import { SVGUniqueID } from 'react-svg-unique-id'
+import { SVGIconProps } from '../types'
+
+${variables.interfaces}
```
m7kvqbe1: Is it worth destructuring `variables`?

```typescript
template({ interfaces, componentName, jsx }, ...)

// or

const { interfaces, componentName, jsx } = variables
```

m7kvqbe1 (approved):
Nice!

thyhjwb6 (approved):
👍

PR #2952, opened 2022-01-05
test(DatePickerE): Fix intermittently failing test

### LINKED ISSUE #2898: DatePickerE: Fix intermittently failing test
The test under the `describe()` block ```when `startDate` and `endDate` are updated externally``` is not reliably passing and is currently skipped.

It should be fixed.

(Split from #2863.)

jpveooys:
## Related issue

Fixes #2898

## Overview

This rewrites the skipped DatePickerE test for prop updates that was previously failing in some conditions.

## Reason

The test was only disabled temporarily as it wasn't reliably passing

## Work carried out

- [x] Investigate reason for failure
- [x] Rewrite test in a more reliable fashion

## Developer notes

The failures were somehow linked to using Jest fake timers, but the failures continued even if real timers were reactivated just for this test. There was no clear reason for this (the component state was updating correctly, but the expected rerender wasn't happening). The same example also behaves correctly in Storybook.

The test has been amended to use `rerender()` to test updating props in a similar fashion to this example: https://testing-library.com/docs/example-update-props

This is behaving as expected. It can also be verified that the test is updating props and not remounting the component by commenting out this line and observing that the test fails:

https://github.com/defencedigital/mod-uk-design-system/blob/bad6d69b216f80c4c6ef91d20604bf2628d3cea6/packages/react-component-library/src/components/DatePickerE/useSelection.ts#L76

### DIFF
--- a/packages/react-component-library/src/components/DatePickerE/DatePickerE.test.tsx
+++ b/packages/react-component-library/src/components/DatePickerE/DatePickerE.test.tsx
@@ -1,5 +1,5 @@
 import { isValid } from 'date-fns'
-import React, { useState } from 'react'
+import React from 'react'
 import '@testing-library/jest-dom/extend-expect'
 import { ColorDanger800 } from '@defencedigital/design-tokens'
 import {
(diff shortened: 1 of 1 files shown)


m7kvqbe1 (approved):
👍

PR #2953, opened 2022-01-06
test(Forms): Add vanilla example coverage

### LINKED ISSUE #2893: Expand on form Storybook e2e examples: vanilla

m7kvqbe1:
## Related issue

Closes #2893

## Overview

Consume form components without relying on any companion libraries (Formik, react-hook-form etc).

## Reason

>We need to be sure that the form components work in a variety of use cases.

## Work carried out

- [x] Create new vanilla forms example
- [x] Tweak some types and code style

### DIFF
--- a/packages/react-component-library/cypress/specs/forms/index.spec.ts
+++ b/packages/react-component-library/cypress/specs/forms/index.spec.ts
@@ -23,7 +23,7 @@ const expectedResult = {
     exampleNumberInput: 1,
     exampleRangeSlider: [28],
   },
-  Vanilla: {
+  Native: {
     email: 'hello@world.com',
     password: 'password',
     description: 'Hello, World!',
@@ -46,15 +46,15 @@ describe('Form Examples', () => {
       const examples = [
         {
           name: 'Formik',
-          uri: '/iframe.html?id=forms-formik--default&viewMode=story',
+          uri: '/iframe.html?id=forms-formik--example&viewMode=story',
         },
         {
           name: 'react-hook-form',
-          uri: '/iframe.html?id=forms-react-hook-form--default&viewMode=story',
+          uri: '/iframe.html?id=forms-react-hook-form--example&viewMode=story',
         },
         {
-          name: 'Vanilla',
-          uri: '/iframe.html?id=forms-vanilla--default&viewMode=story',
+          name: 'Native',
+          uri: '/iframe.html?id=forms-native--example&viewMode=story',
         },
       ]
 
--- a/packages/react-component-library/cypress/specs/forms/index.spec.ts
+++ b/packages/react-component-library/cypress/specs/forms/index.spec.ts
@@ -23,6 +23,16 @@ const expectedResult = {
     exampleNumberInput: 1,
     exampleRangeSlider: [28],
   },
+  Vanilla: {
+    email: 'hello@world.com',
+    password: 'password',
+    description: 'Hello, World!',
+    exampleCheckbox: [],
+    exampleRadio: ['Option 1'],
+    exampleSwitch: '1',
+    exampleNumberInput: 1,
+    exampleRangeSlider: [28],
+  },
 }
 
 describe('Form Examples', () => {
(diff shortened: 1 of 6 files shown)


packages/react-component-library/src/forms/vanilla/useVanillaForm.ts
```diff
@@ -0,0 +1,99 @@
...
+  initialErrors: FormErrors,
+  validation: Record<string, (value: string) => boolean>
+) => {
+  const [formErrors, setFormErrors] = useState<FormErrors>(initialErrors)
+  const [formPayload, setFormPayload] = useState<T | undefined>()
+  const [formState, setFormState] = useState<T>(initialState)
+  const [isSubmitting, setIsSubmitting] = useState<boolean>(false)
+
+  const handleChange = (e: React.FormEvent | SyntheticFormEvent): void => {
+    const { name, value } = e.target as HTMLInputElement
```
jpveooys: Are we able to use `React.FormEvent<HTMLInputElement>` and `e.currentTarget` here if `SyntheticFormEvent` is changed to match?

packages/react-component-library/src/forms/vanilla/useVanillaForm.ts
```diff
@@ -0,0 +1,99 @@
+import React, { useState } from 'react'
```
jpveooys: Would be nice to exclude this file from SonarQube coverage if we're able to.

packages/react-component-library/src/forms/vanilla/vanilla.stories.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+      <pre data-testid="form-example-values">
+        {JSON.stringify(formPayload, null, 2)}
+      </pre>
+    </main>
+  )
+}
+
+export default {
+  title: 'Forms/Vanilla',
```
jpveooys: I wonder if 'Native' would be more descriptive than 'Vanilla'...

packages/react-component-library/src/forms/vanilla/vanilla.stories.tsx
```diff
@@ -0,0 +1,194 @@
...
+  component: ExampleVanillaForm,
+} as Meta
```
jpveooys: Btw they've changed the types a bit in Storybook 6.4:

image

But we need to update it everywhere...

jpveooys (commented):
Nice!

Few other points...

1. Noticed something a bit weird – in some cases the JSON data updates without the form being submitted:

    vanilla

    Any ideas?

2. There seems to be two identical entries in the Storybook sidebar:

    image

    Any way to get rid of one?
    
3. Would be good to keep the initial error state of the email field consistent with the other examples if possible.

packages/react-component-library/src/forms/vanilla/vanilla.stories.tsx
m7kvqbe1: Updated for forms and created a [new ticket](https://github.com/defencedigital/mod-uk-design-system/issues/2961) to handle this across the rest of the codebase.

sonar-project.properties
```diff
@@ -7,8 +7,7 @@ sonar.projectName=mod-uk-design-system
 # Path to sources
 sonar.sources=packages/react-component-library/src
 sonar.exclusions=**/*.test.ts,**/*.test.tsx*,**/*.stories.tsx*
-sonar.coverage.exclusions=**/*.test.ts,**/*.test.tsx*,**/*.stories.tsx*
-
+sonar.coverage.exclusions=**/*.test.ts,**/*.test.tsx*,**/*.stories.tsx*,forms/**/*
```
jpveooys: Might need to be the full path:

```suggestion
sonar.coverage.exclusions=**/*.test.ts,**/*.test.tsx*,**/*.stories.tsx*,packages/react-component-library/src/forms/**/*
```

packages/react-component-library/cypress/specs/forms/index.spec.ts
```diff
@@ -42,6 +52,10 @@ describe('Form Examples', () => {
           name: 'react-hook-form',
           uri: '/iframe.html?id=forms-react-hook-form--default&viewMode=story',
```
jpveooys: These URLs need updating now I think...

jpveooys (approved):
:+1:

PR #2954, opened 2022-01-06
fix(NumberInputE): Improve handling of invalid input

### LINKED ISSUE #2948: Only allow a single decimal point character to be typed in `NumberInputE`
Block pasting of invalid input

Make the mask actually work as a mask instead of sanitizing invalid input.

jpveooys:
## Related issue

Fixes #2948

## Overview

This makes a few changes to NumberInputE. It:

- stops invalid numbers from being pasted. Previously, invalid characters were stripped from pasted values, but we decided that that was undesirable and blocking invalid pastes would be an improvement
- stops a second decimal point being entered (whether typed or pasted before or after the first decimal point)
- where possible, stops the cursor jumping about when invalid characters are typed in the middle of the field
- removes some `onBlur` logic that appeared to be redundant

## Link to preview

- [Storybook](https://5e25c277526d380020b5e418-gswqgmkssr.chromatic.com/?path=/docs/number-input-experimental--default/)
- [Non-Storybook](https://red-playground.netlify.app/) (works in IE 11)

## Reason

For more consistent behaviour, and to avoid user input being silently altered.

## Work carried out

- [x] Tidy up tests slightly
- [x] Rework number validation logic

## Demos (Chrome)

<details>
<summary>Typing multiple decimal points</summary>

### Before

type-point-before

### After

type-point-after

</details>

<details>
<summary>Pasting text</summary>

### Before

pasting-before

### After

pasting-after

</details>

<details>
<summary>Typing letters earlier in the middle of the field</summary>

### Before

letters-before

### After

letters-after

</details>

## Developer notes

Relevant comments are in the code.

I've tested this in Firefox, Chrome and IE 11 and it seems to behave well.

(I wanted to add Cypress tests for pasting as well, but it doesn't have that functionality built in, and simulating the event doesn't work for this.)

### DIFF
--- a/packages/react-component-library/cypress/specs/NumberInputE/index.spec.ts
+++ b/packages/react-component-library/cypress/specs/NumberInputE/index.spec.ts
@@ -3,7 +3,7 @@ import { before, cy, describe, it } from 'local-cypress'
 import selectors from '../../selectors'
 
 describe('NumberInputE', () => {
-  before(() => {
+  beforeEach(() => {
     cy.visit(
       '/iframe.html?id=number-input-experimental--default&viewMode=story'
     )
@@ -14,7 +14,7 @@ describe('NumberInputE', () => {
   })
 
   describe('when the increase button is clicked', () => {
-    before(() => {
+    beforeEach(() => {
       cy.get(selectors.numberInputE.increase).click()
     })
 
@@ -24,12 +24,30 @@ describe('NumberInputE', () => {
   })
 
   describe('when the decrease button is clicked', () => {
-    before(() => {
+    beforeEach(() => {
       cy.get(selectors.numberInputE.decrease).click()
     })
 
     it('should decrement the value', () => {
-      cy.get(selectors.numberInputE.input).should('have.value', '0')
+      cy.get(selectors.numberInputE.input).should('have.value', '-1')
+    })
+  })
+
+  const TYPED_VALUE_CASES = [
+    ['100.123.456', '100.123456'],
+    ['100.123', '100.123'],
+    ['100.invalid456', '100.456'],
+  ]
+
+  TYPED_VALUE_CASES.forEach(([textToType, expectedValue]) => {
+    describe(`when typing "${textToType}"`, () => {
+      beforeEach(() => {
+        cy.get(selectors.numberInputE.input).type(textToType)
+      })
+
+      it(`the value is "${expectedValue}"`, () => {
+        cy.get(selectors.numberInputE.input).should('have.value', expectedValue)
+      })
     })
   })
 })
--- a/packages/react-component-library/src/components/NumberInputE/NumberInputE.tsx
+++ b/packages/react-component-library/src/components/NumberInputE/NumberInputE.tsx
@@ -136,15 +136,16 @@ function formatValue(
   return displayValue
 }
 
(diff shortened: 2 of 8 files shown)


packages/react-component-library/cypress/specs/NumberInputE/index.spec.ts
```diff
@@ -3,7 +3,7 @@ import { before, cy, describe, it } from 'local-cypress'
 import selectors from '../../selectors'
 
 describe('NumberInputE', () => {
-  before(() => {
+  beforeEach(() => {
```
jpveooys, line 6: The component state was leaking between tests so I had to change this.

packages/react-component-library/src/components/NumberInputE/NumberInputE.tsx
```diff
@@ -190,6 +191,60 @@ export const NumberInputE: React.FC<NumberInputEProps> = ({
     )
   }
 
+  /**
```
m7kvqbe1: Nice trick this is new to me.

m7kvqbe1 (approved):
👍

packages/react-component-library/src/components/NumberInputE/NumberInputE.tsx
```diff
@@ -190,6 +191,60 @@ export const NumberInputE: React.FC<NumberInputEProps> = ({
...
+   * Block invalid input before the DOM is updated where possible (to
+   * avoid the cursor jumping).
+   *
+   * Note: React simulates this event using other native events.
+   */
+  const handleBeforeInput = (
```
thyhjwb6: I wonder if we start to think about moving some of this stuff out of here.

packages/react-component-library/src/components/NumberInputE/NumberInputE.tsx
```diff
@@ -190,6 +191,60 @@ export const NumberInputE: React.FC<NumberInputEProps> = ({
...
+
+  const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    if (!isValueValid(event.currentTarget.value)) {
+      return
+    }
+
+    const newValue = getNewValue(event)
+
+    if (!newValue || canCommit(newValue)) {
+      setCommittedValue(newValue)
```
thyhjwb6: Maybe because this interacts with `setCommittedValue` it is an indication that it could belong in a hook.

thyhjwb6 (approved):
Looks good - just minor observations about how the code could evolve. 👍

jpveooys:
@thyhjwb6 I've moved various bits from `NumberInputE.tsx` into separate files. Let me know if that's what you had in mind.

thyhjwb6 (approved):
That looks good 👍

k9dh3zij:
Thanks @jpveooys - love the demos! :) 
Everything looks great, except it won't let me type in or paste a negative number in. However using the buttons, it does go to negatives, so I'm assuming we are disallowing '-', which I think we shouldn't.

Thank you!

jpveooys:
@k9dh3zij Good spot! That was the current behaviour for NumberInputE, unfortunately: https://storybook.design-system.digital.mod.uk/?path=/docs/number-input-experimental--default.

It's a bit trickier as `-` by itself isn't a valid number. I think we should create a new issue for it.

PR #2974, opened 2022-01-12
docs(Storybook): Resolve deprecation warnings

jpveooys:
## Overview

This fixes a couple of deprecation warnings I noticed.

## Work carried out

- [x] Fix deprecation warnings

## Developer notes

See:

- https://github.com/storybookjs/storybook/blob/master/MIGRATION.md#deprecated-showroots-config-option
- https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#deprecated-package-composition-disabled-parameter

### DIFF
--- a/packages/react-component-library/src/components/ThemedExample/ThemedExample.stories.tsx
+++ b/packages/react-component-library/src/components/ThemedExample/ThemedExample.stories.tsx
@@ -12,7 +12,6 @@ export default {
   title: 'Custom Theming',
   parameters: {
     a11y: {
-      disabled: true,
       config: {
         rules: [
           {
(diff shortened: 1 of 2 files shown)


packages/react-component-library/src/components/ThemedExample/ThemedExample.stories.tsx
```diff
@@ -12,7 +12,6 @@ export default {
   title: 'Custom Theming',
   parameters: {
     a11y: {
-      disabled: true,
```
jpveooys, line 15: `disabled` is deprecated (`disable` is the replacement), but on closer inspection this was disabling the a11y addon and disabling one of its rules. Both shouldn't be necessary, so I've left it at disabling the failing rule only.

thyhjwb6 (approved):
👍

PR #2977, opened 2022-01-13
fix(NumberInputE): Don't blur buttons on click

### LINKED ISSUE #2759: NumberInputE: Clicking on +/- buttons shouldn't blur button
From https://github.com/defencedigital/mod-uk-design-system/pull/2734#issuecomment-952889944:

> Tab with keyboard to plus or minus button > press Enter > focus moves elsewhere, which shouldn't happen.

jpveooys:
## Related issue

Resolves #2759

## Overview

This stops the `NumberInputE` increase and decrease buttons from blurring themselves on click. (This affects all ways of clicking e.g. left mouse button, Enter, Space).

## Link to preview

[Storybook](https://5e25c277526d380020b5e418-nujclwrjnc.chromatic.com/?path=/docs/number-input-experimental--default)

[Non-Storybook](https://red-playground.netlify.app/)

## Reason

The focus shouldn't be manipulated in this way, and it prevents e.g. pressing Enter multiple times on the button.

## Work carried out

- [x] Remove blurring logic

### Before

Screencast_13-01-22_11:17:04

### After

Screencast_13-01-22_11:17:52

### DIFF
--- a/packages/react-component-library/src/components/NumberInputE/Buttons.tsx
+++ b/packages/react-component-library/src/components/NumberInputE/Buttons.tsx
@@ -39,11 +39,7 @@ export const Buttons: React.FC<ButtonsProps> = ({
 
   function onButtonClick(getNewValue: () => string) {
     return (event: React.MouseEvent<HTMLButtonElement>) => {
-      const target = event.currentTarget
-      target.blur()
-
       const newValue = getNewValue()
-
       onClick(event, newValue)
     }
   }
--- a/packages/react-component-library/src/components/NumberInputE/NumberInputE.test.tsx
+++ b/packages/react-component-library/src/components/NumberInputE/NumberInputE.test.tsx
@@ -175,6 +175,28 @@ describe('NumberInputE', () => {
       })
     })
 
+    describe.each([
+      { buttonType: 'increase', expectedValue: '1' },
+      { buttonType: 'decrease', expectedValue: '-1' },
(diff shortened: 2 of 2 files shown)


k9dh3zij:
Thanks @jpveooys ! 

One small thing: pressing Enter doesn't trigger the 'button pressed' style (background colour) change. Seems to work fine with Space and for mouse clicks.

Other than that 👍 from me :)

jpveooys:
> One small thing: pressing Enter doesn't trigger the 'button pressed' style (background colour) change. Seems to work fine with Space and for mouse clicks.

Hmm the behaviour seems consistent with a standard button for me:

- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#example; or
- https://design-system.service.gov.uk/components/button/

Are you seeing something different on those?

k9dh3zij:
@jpveooys Good point. Pressing Enter triggers an immediate full click (mouse button down followed by mouse button up/release) before Enter is even released, which is probably why there's no visual feedback. 

Pressing Space triggers the equivalent events on key down and then on key up/release, which is why the button visual state (correctly) stays pressed while the key is pressed. 

It's unfortunate that this results in no visual feedback when Enter is pressed, but we can't change the Internet, can we? 🙃

👍 from me, thanks! 🙂

thyhjwb6 (approved):
👍

PR #2979, opened 2022-01-13
feat(AutocompleteE): Embolden filter value

### LINKED ISSUE #2959: Add `Autocomplete` - show bold text when filtering
Split out from #2890

thyhjwb6:
## Related issue
Closes #2959 

## Overview
Show options with the text used to filter in bold.

## Link to preview
https://autocomplete-bold.netlify.app/?path=/story/autocomplete-experimental--default

## Reason
User needs to clearly see why option was filtered.

## Work carried out
- [x] Add bold text

### DIFF
--- a/packages/react-component-library/src/components/AutocompleteE/AutocompleteEOption.tsx
+++ b/packages/react-component-library/src/components/AutocompleteE/AutocompleteEOption.tsx
@@ -1,13 +1,25 @@
 import React from 'react'
+import reactStringReplace from 'react-string-replace'
 
-import { SelectBaseOption, SelectBaseOptionProps } from '../SelectBase'
+import { SelectBaseOption, SelectBaseOptionAsStringProps } from '../SelectBase'
 
 // eslint-disable-next-line @typescript-eslint/no-empty-interface
-export interface AutocompleteEOptionProps extends SelectBaseOptionProps {}
+export interface AutocompleteEOptionProps
+  extends SelectBaseOptionAsStringProps {
+  inputValue?: string
+}
 
 export const AutocompleteEOption = React.forwardRef<
   HTMLLIElement,
   AutocompleteEOptionProps
->((props, ref) => <SelectBaseOption {...props} ref={ref} />)
+>(({ children, inputValue, ...rest }, ref) => {
+  return (
+    <SelectBaseOption ref={ref} {...rest}>
+      {reactStringReplace(children, inputValue, (match, i) => (
+        <strong key={i}>{match}</strong>
+      ))}
+    </SelectBaseOption>
+  )
+})
 
 AutocompleteEOption.displayName = 'AutocompleteEOption'
--- a/packages/react-component-library/src/components/SelectBase/SelectBaseOption.tsx
+++ b/packages/react-component-library/src/components/SelectBase/SelectBaseOption.tsx
@@ -6,13 +6,26 @@ import { StyledOptionBadge } from './partials/StyledOptionBadge'
 import { StyledOptionText } from './partials/StyledOptionText'
 import { ComponentWithClass } from '../../common/ComponentWithClass'
 
-export interface SelectBaseOptionProps extends ComponentWithClass {
+export interface SelectBaseOptionBaseProps extends ComponentWithClass {
   badge?: string | number
   icon?: React.ReactNode
-  children: string
   isHighlighted?: boolean
   value: string
 }
+
+export interface SelectBaseOptionAsArrayProps
+  extends SelectBaseOptionBaseProps {
+  children: React.ReactNodeArray
+}
+
+export interface SelectBaseOptionAsStringProps
+  extends SelectBaseOptionBaseProps {
+  children: string
+}
+
+export type SelectBaseOptionProps =
+  | SelectBaseOptionAsArrayProps
+  | SelectBaseOptionAsStringProps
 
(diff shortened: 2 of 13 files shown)


packages/react-component-library/src/components/AutocompleteE/AutocompleteEOption.tsx
```diff
@@ -3,11 +3,28 @@ import React from 'react'
...
 
 // eslint-disable-next-line @typescript-eslint/no-empty-interface
-export interface AutocompleteEOptionProps extends SelectBaseOptionProps {}
+export interface AutocompleteEOptionProps extends SelectBaseOptionProps {
+  inputValue?: string
+}
+
+function emboldenInputValue(children: string, inputValue: string) {
+  if (inputValue) {
+    const regex = new RegExp(`(${inputValue})`, 'gi')
```
jpveooys: Will need to first RegEx escape `inputValue` (for example typing `*` in the component breaks things at the moment) before putting it through the RegExp (and also do some HTML escaping :confused: ).

So something like (not tested...):

```tsx
import {escape as escapeHTML, escapeRegExp} from 'lodash'

const escapedInputValue = escapeRegExp(escapeHTML(inputValue))

// use escapedInputValue in the RegExp

return escapeHTML(children).replace(regex, '<strong>$1</strong>')
```

packages/react-component-library/src/components/AutocompleteE/AutocompleteEOption.tsx
```diff
@@ -3,11 +3,28 @@ import React from 'react'
...
+    return children.replace(regex, '<strong>$1</strong>')
+  }
+
+  return children
```
jpveooys: Need to HTML escape this one as well:

```suggestion
  return escapeHTML(children)
```

(Can test by making an option e.g. `&lt;b&gt;thirty`).

packages/react-component-library/src/components/AutocompleteE/AutocompleteEOption.tsx
jpveooys: The RegExp is searching anywhere in the string while the actual search is only searching from the beginning. That leads to e.g.:

image

(the second 't' in thirty is bold).

Would be good to make the two consistent.

packages/react-component-library/src/components/AutocompleteE/AutocompleteEOption.tsx
```diff
@@ -3,11 +3,28 @@ import React from 'react'
...
+}
 
 export const AutocompleteEOption = React.forwardRef<
   HTMLLIElement,
   AutocompleteEOptionProps
->((props, ref) => <SelectBaseOption {...props} ref={ref} />)
+>(({ children, inputValue, ...rest }, ref) => {
+  return (
+    <SelectBaseOption ref={ref} {...rest}>
+      {emboldenInputValue(children, inputValue)}
```
jpveooys: Less important than the other comments, but if it's HTML I think it would be better to pass it as e.g. an `html` prop as you wouldn't normally expect `children` to contain raw HTML. (Alternatively could try to rejig things to use `dangerouslySetInnerHTML` here instead of in `SelectBaseOption`.)

packages/react-component-library/src/components/AutocompleteE/AutocompleteEOption.tsx
thyhjwb6: Actually there is a bug with the actual filter because it should search anywhere to reach feature parity with the existing `Select` so it's a good time to fix that bug.

packages/react-component-library/src/components/AutocompleteE/AutocompleteEOption.tsx
thyhjwb6: This is a good idea as I felt uncomfortable just applying it to `children` and I will give your final comment some thought.

jpveooys:
Another idea to avoid `dangerouslySetInnerHTML` and some of the escaping: maybe it's possible to split the string and wrap the fragments in `<strong>` as needed?

k9dh3zij:
Thanks @thyhjwb6 - all looking good except if I type a `+`, `*`, `\` or `[` (and possibly other chars) it tells me off 😔

image

packages/react-component-library/src/components/SelectBase/SelectBaseOption.tsx
```diff
@@ -6,14 +6,27 @@ import { StyledOptionBadge } from './partials/StyledOptionBadge'
...
   badge?: string | number
   icon?: React.ReactNode
-  children: string
   isHighlighted?: boolean
   value: string
 }
 
+export interface SelectBaseOptionAsArrayProps
+  extends SelectBaseOptionBaseProps {
+  children: React.ReactNodeArray
```
jpveooys, line 18: Hmm this is marked as deprecated with the comment:

> Use either ReactNode[] if you need an array or Iterable<ReactNode> if its passed to a host component.

jpveooys:
@k9dh3zij Think Netlify is out-of-date – try this one: https://5e25c277526d380020b5e418-otrfxmwylm.chromatic.com/?path=/docs/autocomplete-experimental--default

jpveooys (commented):
Looks excellent now!

jpveooys (approved):

k9dh3zij:
👍 Looks perfect, thank you!

PR #2980, opened 2022-01-14
feat(DatePickerE): Add focus trap

### LINKED ISSUE #2870: DatePickerE: Add focus trap
From @k9dh3zij (in #2841):

> Can we make it so that the Sheet contains its own tab sequence. (Tab and Shift + Tab do not move focus outside the dialog)

This is based on https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html

There is a library that can do this: https://github.com/focus-trap/focus-trap-react

However, from testing, there seems to be a bug in the underlying library (https://github.com/focus-trap/focus-trap) – when a date after the first of the month is focused, the tab key stops functioning (while Shift+Tab continues to work).

This needs further investigation and a bug to be reported.

jpveooys:
## Related issue

Resolves #2870

## Overview

This adds a focus trap to DatePickerE using the react-focus-trap library. This means that the focus stays within the picker when it is open and the tab key is pressed.

Note: General keyboard navigation within the picker is handled by react-day-picker and is unaffected by this change. (Note that it has been significantly improved for react-day-picker v8, so we'll get those fixes when we are able to update.)

## Link to preview

[Storybook](https://5e25c277526d380020b5e418-pzlguvhnwj.chromatic.com/?path=/docs/date-picker-experimental--default)

[Non-Storybook](https://red-playground.netlify.app/)

## Reason

This is to be consistent with https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html.

## Work carried out

- [x] Integrate react-focus-trap library with DatePickerE

### Before

Screencast_14-01-22_13:52:37

### After

Screencast_14-01-22_13:53:34

## Developer notes

This also replaces the existing document click and escape key logic for this component as these are both handled by react-focus-trap (and focus-trap).

The `isOpen` prop was also changed to affect the first render only and renamed to `initialIsOpen` (matching `SidebarE`). This is because it doesn't make much sense to have it affect subsequent renders – you never know whether the prop or the internal state is correct at any point in time.

Finally, the `usehooks-ts` library was added to gain access to some utility hooks it provides (`useBoolean` in this case).


k9dh3zij:
Thanks very much @jpveooys ! As discussed:

1. There is a browser highlight around buttons and the sheet itself now, which shouldn't be there
image

2. The initial focus is currently on the sheet itself. If possible we should set focus either 
- on the currently selected/typed date; or
- on today's date, if no other date has been selected. 

If that's not possible, then setting focus on the left button would be preferable. 

3. Currently the keyboard arrows change:
- the month/year when focus is on the sheet or buttons  
- the selected date when focus is on a calendar day

Even though it's great to have a way to quickly change month/year, this behaviour can lead to mistakes, because the same keys work in different ways in a very similar context. 

Our future plans include a month/year selector within the visual date picker itself, so for now I'd suggest disabling this additional functionality if/when possible.

Thanks!

jpveooys:
Thanks for the review @k9dh3zij !

> 1. There is a browser highlight around buttons and the sheet itself now, which shouldn't be there

I've removed these (though nothing has changed on our side – could be something has changed in Chrome if it has changed...)

> 2. The initial focus is currently on the sheet itself. If possible we should set focus either ...

As discussed, in react-day-picker v7 it's not possible to focus a date in that manner, but it will be possible in v8. I've changed the initial focus to the previous month button for now though.

> 3. Currently the keyboard arrows change: ...

This also isn't customisable in react-day-picker v7 (only way to turn it off also turns off the next and previous month buttons...) But it's gone in v8 :smile: So it's there for now, but I'll pick up the v8 issue again separately and see if we can work around the blocker that was preventing us updating :slightly_smiling_face:

packages/react-component-library/package.json
```diff
@@ -181,7 +182,8 @@
     "react-toast-notifications": "^2.4.0",
     "react-transition-group": "^4.4.1",
     "styled-normalize": "^8.0.7",
-    "styled-theming": "^2.2.0"
+    "styled-theming": "^2.2.0",
+    "usehooks-ts": "^2.2.1"
```
m7kvqbe1, line 187: Nice - this has some cool stuff.

m7kvqbe1 (approved):
LGTM behaves nicely.

k9dh3zij:
Thanks for making those changes @jpveooys. Looks great and I think the initial focus on the button rather than the sheet itself works better.

PR #2986, opened 2022-01-14
chore(Storybook): Use Storybook 6.3 component types where possible

### LINKED ISSUE #2961: Update Storybook story format to use new typings (as of 6.4)
image

jpveooys:
## Related issue

Resolves #2961

## Overview

This updates stories to use the new `ComponentStory` and `ComponentMeta` types added in Storybook 6.3 (and used in their code samples) where possible.

## Link to preview

https://5e25c277526d380020b5e418-mgkgifavsh.chromatic.com/

## Reason

To stay up-to-date and be consistent with official Storybook code samples.

## Work carried out

- [x] Update to `ComponentStory` and `ComponentMeta` types where possible

## Developer notes

There is some information about the new types here: https://github.com/storybookjs/storybook/blob/86a34fac51307b5d4d3add7d6fac0abb147d062b/app/react/src/client/preview/types-6-3.ts

Note that there were a a few cases where it wasn't possible to update to the new types. These are mainly where the type of the component's props is a union, and it didn't know which member of the union was meant. We could decide to stick with `Story` and `Meta` everywhere if we wanted as they are still supported (but we should be passing the props to `Meta` as well if we did that).

### DIFF
--- a/packages/react-component-library/src/components/Alert/Alert.stories.tsx
+++ b/packages/react-component-library/src/components/Alert/Alert.stories.tsx
@@ -1,11 +1,17 @@
 import React from 'react'
-import { Story, Meta } from '@storybook/react'
+import { ComponentStory, ComponentMeta } from '@storybook/react'
 
-import { Alert, AlertProps, ALERT_VARIANT } from './index'
+import { Alert, ALERT_VARIANT } from './index'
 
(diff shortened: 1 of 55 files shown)


m7kvqbe1 (approved):
👍

PR #2995, opened 2022-01-18
feat(GlobalStyleProvider): Remove root font-size breakpoints

### LINKED ISSUE #2761: Review the need for font breakpoints and remove if not needed

m7kvqbe1:
## Related issue

Closes #2761

## Overview

Remove redundant font-size breakpoints in `GlobalStyleProvider`.

## Reason

>These are redundant and cause weird visual issues in the majority of applications. Some downstream teams are already writing overrides to turn it off.

## Work carried out

- [x] Remove redundant breakpoints

### DIFF
--- a/packages/design-tokens/README.md
+++ b/packages/design-tokens/README.md
@@ -21,7 +21,7 @@ The reccomended way to access tokens is using the supplied selector functions.
 Simply import the selector object and then destructure the selectors you want to use out of this.
 
 ```javascript
-import { selectors } from "@defencedigital/design-tokens";
+import { selectors } from '@defencedigital/design-tokens'
 
 const { color, spacing, mediaQuery } = selectors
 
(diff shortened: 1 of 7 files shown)


jpveooys (commented):
Nice :+1: 

Do we want to remove them from `packages/design-tokens` as well?

m7kvqbe1:
> Nice 👍
> 
> Do we want to remove them from `packages/design-tokens` as well?

I see what you mean - yes i'll remove the `baseFontSize` properties.

jpveooys:
> I see what you mean - yes i'll remove the `baseFontSize` properties.

Yep, sorry, that's what I meant. Still a few more references in these files I think though:

- `packages/design-tokens/src/tokens/dark/breakpoints.json`
- `packages/design-tokens/src/types.ts`
- `packages/design-tokens/README.md`
- `packages/design-tokens/src/getters.ts`
- `packages/design-tokens/src/__tests__/selectors.test.tsx`

jpveooys (approved):
:+1:

Issue #2872, opened 2021-12-06
DatePickerE: Tabbing away from day in calendar should clear range highlight


jpveooys:
When the date picker is in range mode and a second date is being selected (focused) using the keyboard arrow keys, the days between the first and focused date are shaded.

However, the shading isn't cleared when tabbing to a different element. This is due to react-day-picker v7 not exposing a `onDayBlur` prop (which would allow us to clear the highlight).

However, the prop has been added for react-day-picker v8 which is currently in beta, so this should be fixable once a stable version of it is released.

jpveooys:
Note: v8 is due for release before the end of the year: https://github.com/gpbl/react-day-picker/discussions/1316

We could start using it for DatePickerE only using an alias.

gpbl:
👋 hey thanks for looking at react-day-picker! 

Please remember v8 is [a major update](https://github.com/gpbl/react-day-picker/discussions/1088) mainly incompatible with the previous version. If you are evaluating the beta or considering to upgrade, consider to contact the mantainers at the [DayPicker Slack](https://github.com/gpbl/react-day-picker/discussions/1318). We can help for real use-cases for DayPicker and see how it is implemented around.

jpveooys:
Hi @gpbl :wave: 

Thanks for visiting!  :smile: 

Our date picker component (`DatePicker`) is built on react-day-picker and we're currently working on a new version of it (`DatePickerE`) as you've probably gathered :slightly_smiling_face: 

We're planning on having a look at migrating `DatePickerE` to the react-day-picker v8 beta as there's lots of good stuff in there we've been looking for :smile: But realistically, I think this will be in January now (with Christmas time off etc.)

Thanks again!

jpveooys:
Working branch: https://github.com/defencedigital/mod-uk-design-system/tree/feature/date-picker-e-v8

Issue #2871, opened 2021-12-06
DatePickerE: Focus date when picker opened


jpveooys:
From @k9dh3zij (in #2841):

> When opening the Sheet, can we move focus to the currently selected date? (as in https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html)

The linked example focuses the selected date if one has been previously selected, or otherwise the current date, when the picker is opened.

This is related to the following issue: https://github.com/gpbl/react-day-picker/issues/1278

PR #3001, opened 2022-01-18
feat(DatePicker): Migrate to react-day-picker v8

jpveooys:
## Related issue

#2871
#2872 

## Overview

Work in progress PR to migrate DatePicker to react-day-picker v8.

## Link to preview

todo

## Reason

To resolve various accessibility and keyboard navigation problems.

## Work carried out

- [x] Migrate DatePicker to react-day-picker v8
- [x] Fix bug with shading of keyboard range selection when second date is blurred
- [x] Focus selected date or current date on open

## Demo

todo

## Developer notes

todo

### DIFF
--- a/packages/react-component-library/cypress/e2e/DatePicker/index.cy.ts
+++ b/packages/react-component-library/cypress/e2e/DatePicker/index.cy.ts
@@ -2,10 +2,10 @@ import { describe, cy, it, before } from 'local-cypress'
 import { addDays, startOfMonth, format } from 'date-fns'
 import { ColorNeutral200 } from '@defencedigital/design-tokens'
 
+import { formatDatesForInput } from '../../../src/components/DatePicker/utils'
 import { DATE_FORMAT } from '../../../src/constants'
 import { hexToRgb } from '../../helpers'
 import selectors from '../../selectors'
-import { formatDatesForInput } from '../../../src/components/DatePicker/utils/formatDatesForInput'
 
 describe('DatePicker', () => {
   describe('when a day is selected', () => {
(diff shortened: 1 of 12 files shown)


c9qsm62x (commented):
https://github.com/defencedigital/mod-uk-design-system/blob/9eb0eb2597b4cbcba88d0e1c2a158ec0e3a9c3ba/packages/react-component-library/src/components/DatePicker/DatePicker.test.tsx#L962-L972

We have failing test here but was wondering if this is needed because I cant thing of a scenario when this will happen as its readOnly

jpveooys:
@c9qsm62x It was added for coverage IIRC as those code paths still exist in the `useInput` hook.

I think the test has stopped doing what it was meant to since user-event was updated to v14 – `fireEvent.keyDown(input, { key: 'Enter' })` probably makes more sense.

c9qsm62x:
Test
DatePicker › when a single date picker is rendered and the picker is opened › when Shift-Tab is pressed once › traps the focus within the picker

it looks like this behaviour has changed in V8  
  https://react-day-picker.js.org/basics/keyboard#example
  Having a look at this example it looks like when a date is in focus and you press shift+tab it will focus on the next month button.

jpveooys:
Yep that test looks wrong now as the initial focus is now set to the selected date.

Should be something like this I think:

```tsx
    describe('when Tab is pressed once', () => {
      beforeEach(async () => {
        await user.tab()
      })

      it('traps the focus within the picker', () => {
        expect(
          wrapper.getByLabelText(PREVIOUS_MONTH_BUTTON_LABEL)
        ).toHaveFocus()
      })

      describe('and Shift-Tab is then pressed once', () => {
        beforeEach(() => {
          return user.tab({ shift: true })
        })

        it('still traps the focus within the picker', () => {
          expect(
            wrapper.getByRole('button', { name: '5th December (Thursday)' })
          ).toHaveFocus()
        })
      })
    })
```

m7kvqbe1:
Closing for now because it's very stale - hoping to come back to this soon.

Issue #2971, opened 2022-01-11
DatePickerE: Component misbehaves if value from onChange passed back to component


jpveooys:
If the component is used in a fully-controlled manner and value(s) received from the `onChange` callback are passed back to the `startDate` and `endDate` props, the component misbehaves when a date is being typed.

However, this should work, as it's the typical way a controlled component is used in a form. 

For example, this should work when typing a date in the date picker input:

```tsx
function Example() {
  const [value, setValue] = React.useState<Date | null>(null)

  return (
    <div>
      <DatePickerE startDate={value} onChange={({ startDate }) => setValue(startDate)} />
      <button onClick={() => setValue(new Date())}>Today</button>
    </div>
  );
}
```

(The reason that it doesn't work is that the component is constantly trying to react to changes to `startDate`, and it's an invalid date while one is being typed.)

After comparing the behaviour to `<input>` and other libraries, my initial suggestion would be that the component should be comparing the `startDate` and `endDate` props to the current value in the internal state, and reacting if they're different to that, rather than reacting whenever the prop value changes.

For example, compare with a standard `<input>` component:

https://codesandbox.io/s/mystifying-cartwright-8kpzk?file=/src/App.tsx

Or a similar component in another library:

https://codesandbox.io/s/basicdatepicker-material-demo-forked-6gbve?file=/demo.js

Note that this would mean passing startDate a fixed value (i.e. `startDate={new Date(...)}`) wouldn't work – the parent component's state would need to be initialised with that value, and it kept updated there (typically by a form library). (It could be made to work with a separate prop analogous to `defaultValue` in `<input>` e.g. `initialStartDate` and `initialEndDate`.)

PR #3004, opened 2022-01-20
fix(DatePickerE): Allow state to be held externally

jpveooys:
## Related issue

Resovles #2971

## Overview

The makes DatePickerE compatible with use cases where the state is held in the parent component (e.g. a form library) and passed back to the component.

## Link to preview

https://5e25c277526d380020b5e418-vgnydmrmpy.chromatic.com/?path=/docs/date-picker-experimental--default

## Reason

This scenario should work as it's a common pattern when using forms.

## Work carried out

- [x] Split `startDate` and `endDate` props into `intialStartDate`, `intialEndDate`, `startDate` and `endDate`
- [x] Refactor the component to use the `useReducer` hook
- [x] Add stories demonstrating state being kept externally

## Developer notes

`initialStartDate` and `initialEndDate` are analogous to `defaultValue` in `<input>`, while `startDate` and `endDate` are analogous to `value` in `<input>`.

If `startDate` and `endDate` are used it is expected that the parent component keeps them updated using the onChange callback (same as `value` for `<input>`).

See the updated stories for examples.

Some further tidying up has been split into a separate PR:

- #3005

### DIFF
--- a/packages/react-component-library/src/components/DatePickerE/DatePickerE.stories.tsx
+++ b/packages/react-component-library/src/components/DatePickerE/DatePickerE.stories.tsx
@@ -2,12 +2,12 @@ import { action } from '@storybook/addon-actions'
 import { ComponentMeta, ComponentStory } from '@storybook/react'
 import { parseISO } from 'date-fns'
 import { Field, Form, Formik } from 'formik'
-import React from 'react'
+import React, { useState } from 'react'
 import * as yup from 'yup'
 
 import { withFormik } from '../../enhancers'
 import { ButtonE } from '../ButtonE'
-import { DatePickerE } from '.'
+import { DatePickerE, DatePickerEOnChangeData } from '.'
 
 export default {
   component: DatePickerE,
@@ -33,17 +33,17 @@ const Template: ComponentStory<typeof DatePickerE> = (args) => (
 
 export const Default = Template.bind({})
 
-export const WithExistingValue = Template.bind({})
-WithExistingValue.storyName = 'With existing value'
-WithExistingValue.args = {
-  startDate: parseISO('2021-12-15'),
+export const WithInitialValue = Template.bind({})
+WithInitialValue.storyName = 'With initial value'
+WithInitialValue.args = {
+  initialStartDate: parseISO('2021-12-15'),
 }
 
 export const CustomFormat = Template.bind({})
 CustomFormat.storyName = 'Custom format'
 CustomFormat.args = {
   format: 'yyyy/MM/dd',
-  startDate: parseISO('2021-01-11'),
+  initialStartDate: parseISO('2021-01-11'),
 }
 
 export const CustomInitialMonth = Template.bind({})
@@ -66,7 +66,7 @@ Disabled.args = {
 export const DisabledDays = Template.bind({})
 DisabledDays.storyName = 'Disabled days'
 DisabledDays.args = {
-  startDate: parseISO('2021-04-01'),
+  initialStartDate: parseISO('2021-04-01'),
   disabledDays: [
     parseISO('2021-04-12'),
     parseISO('2021-04-02'),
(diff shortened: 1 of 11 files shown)


packages/react-component-library/src/components/DatePickerE/useInput.ts
```diff
@@ -35,40 +34,42 @@ export function useInput(
...
 
   const handleInputChange = useCallback(
     (event) => {
       if (isRange) {
         return
       }
 
-      setInputValue(event.currentTarget.value)
+      dispatch({
+        type: 'update',
```
m7kvqbe1: Is it worth using constants for these action types? See Timeline for an example.

packages/react-component-library/src/components/DatePickerE/useDatePickerEReducer.ts
```diff
@@ -0,0 +1,103 @@
...
+      return {
+        ...state,
+        inputValue: formatDatesForInput(
+          state.from,
+          state.to,
+          state.datePickerFormat
+        ),
+      }
+    default:
+      throw new Error('Unknown reducer action type')
```
m7kvqbe1, line 47: I wonder if we change the error message elsewhere to be consistent across the code base:

https://github.com/defencedigital/mod-uk-design-system/blob/12e4eda364b81fd657257be1417334f0c068038d/packages/react-component-library/src/components/Timeline/context/reducer.ts#L145

packages/react-component-library/src/components/DatePickerE/useDatePickerEReducer.ts
```diff
@@ -0,0 +1,103 @@
...
+}: Omit<DatePickerEState, 'inputValue'>) {
+  return {
+    from,
+    to,
+    datePickerFormat,
+    inputValue: formatDatesForInput(from, to, datePickerFormat),
+  }
+}
+
+function reducer(state: DatePickerEState, action: DatePickerEAction) {
```
m7kvqbe1: Is is worth adding return types to some of these functions?

packages/react-component-library/src/components/DatePickerE/DatePickerE.tsx
```diff
@@ -57,7 +57,10 @@ export interface DatePickerEProps
   extends ComponentWithClass,
     InputValidationProps {
   /**
-   * End date of the picker (only relevant in range mode).
+   * The end of the selected date range. (Only relevant if isRange is set.)
```
m7kvqbe1, line 61: Should `isRange` be in backticks?

packages/react-component-library/src/components/DatePickerE/DatePickerE.stories.tsx
```diff
@@ -33,17 +33,37 @@ const Template: ComponentStory<typeof DatePickerE> = (args) => (
...
-WithExistingValue.storyName = 'With existing value'
-WithExistingValue.args = {
-  startDate: parseISO('2021-12-15'),
+export const WithInitialValue = Template.bind({})
+WithInitialValue.storyName = 'With initial value'
+WithInitialValue.args = {
+  initialStartDate: parseISO('2021-12-15'),
 }
 
+export const WithExternalState: ComponentStory<typeof DatePickerE> = () => {
```
m7kvqbe1: I wonder if the Storybook story is superfluous? Albeit necessary to develop the feature.

m7kvqbe1 (approved):
LGTM nice change! Just a few small comments.

packages/react-component-library/src/components/DatePickerE/DatePickerE.stories.tsx
jpveooys: I think they're useful for showing how to do this type of stuff. I'll remove them though as they aren't meant to be 'use buttons in this way' either...

packages/react-component-library/src/components/DatePickerE/useInput.ts
jpveooys: I'm not 100% convinced it's an improvement. But have added for consistency :smile:

thyhjwb6 (approved):
👍

PR #3005, opened 2022-01-20
refactor(DatePickerE): Tidy up reducer and related logic

### LINKED ISSUE #2971: DatePickerE: Component misbehaves if value from onChange passed back to component

jpveooys:
## Related issue

Resolves #2971

## Overview

This tidies up various bits of DatePickerE

## Link to preview

https://5e25c277526d380020b5e418-afotyleznc.chromatic.com/?path=/docs/date-picker-experimental--default

## Reason

Split from #3004 to reduce PR size and noise.

## Work carried out

- [x] Rename useSelection hook
- [x] Rename `from` and `to` state properties to `startDate` and `endDate ` for consistency
- [x] Move `hasError` into the reducer
- [x] Remove redundant `click()` function from tests

## Developer notes

n/a

### DIFF
--- a/packages/react-component-library/src/components/DatePickerE/useInput.ts
+++ b/packages/react-component-library/src/components/DatePickerE/useInput.ts
@@ -2,11 +2,7 @@ import { addHours, isValid, parse } from 'date-fns'
 import React, { useCallback } from 'react'
 
 import { RETURN } from '../../utils/keyCodes'
-import {
-  DATEPICKER_E_ACTION,
-  DatePickerEAction,
-  DatePickerEState,
-} from './types'
+import { DATEPICKER_E_ACTION, DatePickerEAction } from './types'
 
 function parseDate(datePickerFormat: string, value: string) {
   if (!value) {
@@ -26,12 +22,8 @@ export function useInput(
   datePickerFormat: string,
   isRange: boolean,
   handleDayClick: (date: Date) => void,
-  state: DatePickerEState,
-  setHasError: React.Dispatch<React.SetStateAction<boolean>>,
   dispatch: React.Dispatch<DatePickerEAction>
 ) {
-  const { startDate } = state
-
   const handleKeyDown = useCallback(
     (event: React.KeyboardEvent<HTMLInputElement>) => {
       if (isRange || event.keyCode !== RETURN) {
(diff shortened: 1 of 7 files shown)


packages/react-component-library/src/components/DatePickerE/useInput.ts
```diff
@@ -64,12 +59,8 @@ export function useInput(
 
       const date = parseDate(datePickerFormat, event.currentTarget.value)
       handleDayClick(date)
-
-      if (isValid(date)) {
```
jpveooys, line 72: This was redundant as the focus border has higher priority than the error border (so this change had no effect as it wasn't visible).

m7kvqbe1 (approved):
👍

PR #3008, opened 2022-01-21
Handle `AutocompleteE` focus/blur

### LINKED ISSUE #2957: Add `Autocomplete` - `onFocus` behaviour with already selected item
Split out from #2890

- presents unfiltered items after and item has been selected
- sets item when tabbing out
- does not set the item when clicking out
- if anything is bad then error the field (do not attempt to fix)

In a separate ticket:
- Cursor starts exactly where the user clicks on (or at the start or end if click was not exactly on the input string) - this is consistent with how Text Input works.

Reference:
- https://react-select.com/creatable

thyhjwb6:
## Related issue
Closes #2957 

## Overview
Correctly handles focus/blur behaviour for `AutocompleteE`.

## Link to preview
https://autocomplete-focus.netlify.app/?path=/docs/autocomplete-experimental--default

## Reason
These are the last bits for the foundation of `AutocompleteE` functionality.

## Work carried out
- [x] Refactor code to support changes
- [x] Make changes for focus/blur

## Developer notes
- [A separate ticket](https://github.com/defencedigital/mod-uk-design-system/issues/3010) has been raised for cursor position when clicking.


packages/react-component-library/src/components/AutocompleteE/AutocompleteE.tsx
```diff
@@ -11,8 +11,7 @@ import {
 } from '../SelectBase'
 import { NoResults } from './NoResults'
 import { useHighlightedIndex } from './hooks/useHighlightedIndex'
-import { useInput } from './hooks/useInput'
-import { useItems } from './hooks/useItems'
+import { useAutocomplete } from './hooks/useAutocomplete'
 
 // eslint-disable-next-line @typescript-eslint/no-empty-interface
 export interface AutocompleteEProps extends SelectBaseProps {}
```
m7kvqbe1, line 18: If we are going to keep the empty interface maybe we add an exclusion for this rule? 

I think i've also done this elsewhere.

packages/react-component-library/src/components/AutocompleteE/AutocompleteE.tsx
```diff
@@ -16,15 +16,53 @@ import { useAutocomplete } from './hooks/useAutocomplete'
...
+
+function onFocus(isOpen: boolean, openMenu: () => void) {
+  return () => {
+    if (!isOpen) {
+      openMenu()
+    }
+  }
+}
+
+function onKeyDown(
```
m7kvqbe1: Is it worth adding return types for these?

I try to be explicit with them but I get that the compiler can infer return types in some instances so maybe it's noise?

m7kvqbe1 (approved):
LGTM

I noticed a slight visual Firefox clipping bug that maybe warrants a seperate ticket.

packages/react-component-library/src/components/AutocompleteE/AutocompleteE.tsx
```diff
@@ -11,23 +11,58 @@ import {
...
 import { NoResults } from './NoResults'
 import { useHighlightedIndex } from './hooks/useHighlightedIndex'
-import { useInput } from './hooks/useInput'
-import { useItems } from './hooks/useItems'
+import { useAutocomplete } from './hooks/useAutocomplete'
 
 // eslint-disable-next-line @typescript-eslint/no-empty-interface
 export interface AutocompleteEProps extends SelectBaseProps {}
 
+function onBlur(resetHighlightedIndex: () => void) {
```
jpveooys: I find the name of these a little bit confusing, as they create event handlers rather than handle the event directly.

So would suggest something like e.g. `createBlurHandler` (or, alternatively, turn them into hooks in which case `useCallback` could be used as well).

jpveooys (commented):
A few other questions...

1. What is the expected behaviour if a partial value is entered and the drop-down is closed and reopened?

    Currently it shows all items with the first one highlighted; maybe that's OK but just wanted to check:
    
    Screencast_24-01-22_09:45:45

2. If you enter an invalid value, click somewhere else (so that the red border is displayed), and then clear the input using the clear button, the red border persists. But if you open and close the drop-down it then goes away:

    Screencast_24-01-22_11:41:11

    Should the border go after clicking the clear button?

3. One for a separate issue I think but if you select a value using the Enter key, nothing is focused afterwards and it's a bit tricky to change your answer using the keyboard.

packages/react-component-library/src/components/AutocompleteE/AutocompleteE.tsx
thyhjwb6: It doesn't hurt 👍

packages/react-component-library/src/components/AutocompleteE/AutocompleteE.tsx
thyhjwb6: As discussed we will leave this as it is for now.

thyhjwb6:
@jpveooys

1. Looks like you are onto something here, I will get @k9dh3zij to confirm correct behaviour
2. ✅ 
3. [Issue raised](https://github.com/defencedigital/mod-uk-design-system/issues/3015)

k9dh3zij:
Thanks @thyhjwb6  - regarding the items raised by @jpveooys 

1. When the Sheet reappears, is it possible to filter the list so it shows only the items that match whatever has already been typed in (or "No results found for xxxx")?

2. This one still persists for me, but assume it's just the code has not been updated.

Nothing more from me, thanks!

thyhjwb6:
@jpveooys @k9dh3zij - fix has been applied and redeployed

https://autocomplete-focus.netlify.app/?path=/story/autocomplete-experimental--default

jpveooys (approved):

PR #3014, opened 2022-01-24
fix(Fonts): Resolve issues with fonts and Storybook

### LINKED ISSUE #3006: Load correct fonts once from canonical source and use GlobalStyleProvider in Storybook
- The font package and fonts.googleapis.com are both being used to include fonts within Storybook.

- It also looks as though the `GlobalStyleProvider` isn't being utilised.

- There seems to be a flashing of an italic font in the new `Autocomplete` implementation. Investigate the use of `font-display: swap`.

- The `Breadcrumbs` component also seems to have alignment issues with a `700` weight font on some browsers and operating systems.

m7kvqbe1:
## Related issue

Closes #3006

## Overview

Resolve issues with font source files and update Storybook config.

## Link to preview

https://61eeb48087f59f0a53ccd317--happy-fermi-3f7b8f.netlify.app

## Reason

>Bad font source files were causing alignment issues on some browsers and operating systems (notably Breadcrumbs).

## Work carried out

- [x] Stop loading fonts from Google CDN
- [x] Add Storybook global decorator for `GlobalStyleProvider`
- [x] Remove `css-framework` include
- [x] Remove deprecated components reliant on `css-framework` package
- [x] Update broken font source files (https://github.com/defencedigital/mod-uk-design-system/pull/3003)
- [x] Add missing default anchor styles (previously existed in `css-framework` package)
- [x] Fix base font-size in `GlobalStyleProvider`


packages/react-component-library/src/components/Link/Link.tsx
```diff
@@ -1,21 +1,19 @@
 import React from 'react'
+import styled from 'styled-components'
 
 import { AnchorType } from '../../common/Link'
 
+const StyledLink = styled.a``
```
jpveooys: Is the empty styled component intended here?

packages/react-component-library/.storybook/preview.js
```diff
@@ -1,6 +1,7 @@
-import '@defencedigital/css-framework/index.scss'
```
jpveooys, line 1: Are we able to get rid of `@storybook/preset-scss` and maybe `@storybook/addon-postcss` with this gone?

jpveooys (commented):
Looks good :+1:  (FYI the preview link is out of date as as it's missing the font size fix in the last commit).

Couple of questions ~~above~~ below :smile:.

packages/react-component-library/src/components/Link/Link.tsx
m7kvqbe1: Ah that's left over from figuring out what was wrong with link styling - i'll remove.

jpveooys (approved):
Looks good :+1:  I think `packages/react-component-library/postcss.config.js` can go now.

k9dh3zij:
👍

PR #3026, opened 2022-01-25
Update `Avatar` pattern

### LINKED ISSUE #1927: Make `Avatar` pattern consistent
`Avatar` is not consistent with other components. We want to compose it as:

```
      <Avatar>AB</Avatar>
```

Which implies `initials` becomes `children`.

thyhjwb6:
## Related issue
Closes #1927 

## Overview
Update `Avatar` pattern.

## Reason
Consistency with other components.

## Work carried out
- [x] Update pattern

### DIFF
--- a/packages/react-component-library/src/components/TopLevelNavigation/SidebarE/SidebarUserE.tsx
+++ b/packages/react-component-library/src/components/TopLevelNavigation/SidebarE/SidebarUserE.tsx
@@ -47,7 +47,9 @@ const SidebarAvatarWithItems: React.FC<SidebarAvatarWithItemsProps> = ({
         aria-label="Show user options"
         data-testid="user-button"
         icon={
-          <StyledUserAvatar initials={initials} data-testid="sidebar-avatar" />
+          <StyledUserAvatar data-testid="sidebar-avatar">
+            {initials}
+          </StyledUserAvatar>
         }
       />
     }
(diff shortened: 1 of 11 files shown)


packages/react-component-library/src/components/TopLevelNavigation/SidebarE/SidebarUserE.tsx
```diff
@@ -47,7 +47,9 @@ const SidebarAvatarWithItems: React.FC<SidebarAvatarWithItemsProps> = ({
         aria-label="Show user options"
         data-testid="user-button"
         icon={
-          <StyledUserAvatar initials={initials} data-testid="sidebar-avatar" />
+          <StyledUserAvatar data-testid="sidebar-avatar">
+            {initials}
+          </StyledUserAvatar>
         }
       />
     }
```
thyhjwb6, line 55: I wasn't sure what to do with `SidebarAvatarWithItems` on line 78 as the pattern does not feel quite right but also it does not seem right to use `children` for the `initials`. Open to ideas

jpveooys: Hmm maybe how that component is composed could be improved, but I'm not sure it's worth it as it's not exported as far as I can see (I think `initials` makes more sense than `children` there in any case).

jpveooys (approved):
:+1:

m7kvqbe1 (approved):

PR #3027, opened 2022-01-26
refactor(DataList): Rename to `DescriptionList`

### LINKED ISSUE #2423: Rename `Data List` to `Description List` to better match HTML spec
We currently have a `Data List` component that does not match the [Data List spec](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist). Instead, the component we have uses `<dl>`, `<dt>` and `<dd>` tags, which are specific to [Description Lists](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl).

- Recommendation is to change the name of the `Data List` component to `Description List`, to be better match the functionality and behaviour of said component.

m7kvqbe1:
## Related issue

Closes #2423

## Overview

Rename component to better match W3C spec.


thyhjwb6 (approved):
Looks good 👍 

I have been taking the opportunity to add a Cypress test in too....

packages/react-component-library/src/components/DescriptionList/DescriptionList.stories.tsx
```diff
@@ -0,0 +1,64 @@
+import React from 'react'
+import { ComponentStory, ComponentMeta } from '@storybook/react'
+
+import { DescriptionList, DescriptionListItem } from './index'
+
+export default {
+  component: DescriptionList,
+  subcomponents: { DescriptionListItem },
+  title: 'Data List',
```
jpveooys: Think this needs updating:

```suggestion
  title: 'Desscription List',
```

packages/react-component-library/src/components/DescriptionList/DescriptionList.test.tsx
```diff
@@ -2,21 +2,21 @@ import React from 'react'
...
-          </DataListItem>
-          <DataListItem title="Two">2</DataListItem>
-          <DataListItem title="Three">3</DataListItem>
-        </DataList>
+          </DescriptionListItem>
+          <DescriptionListItem title="Two">2</DescriptionListItem>
+          <DescriptionListItem title="Three">3</DescriptionListItem>
+        </DescriptionList>
       )
     })
```
jpveooys, line 22: Do the `data list`s in the `it()`s below need updating?

packages/react-component-library/src/components/DescriptionList/DescriptionList.tsx
```diff
@@ -54,7 +54,7 @@ export const DataList: React.FC<DataListProps> = ({
   const sheetId = ariaAttributes && ariaAttributes['aria-owns']
 
   return (
-    <StyledDataList
+    <StyledDescriptionList
       $isCollapsible={isCollapsible}
       $isOpen={open}
       data-testid="data-list"
```
jpveooys: Do the test IDs need updating? (This one and the two below.)

packages/react-component-library/src/components/DescriptionList/DescriptionListItem.tsx
```diff
@@ -20,7 +20,7 @@ export interface DataListItemProps {
   isCollapsible?: boolean
 }
 
-export const DataListItem: React.FC<DataListItemProps> = ({
+export const DescriptionListItem: React.FC<DescriptionListItemProps> = ({
   children,
   title,
   isCollapsible,
```
jpveooys, line 26: Does the test ID below need updating?

jpveooys (commented):
Just a few small bits that look like they need updating too.

jpveooys (approved):

PR #3028, opened 2022-01-26
fix(NumberInputE): Allow negative values to be entered

### LINKED ISSUE #2967: Allow typing and pasting negative numbers into NumberInputE
Currently NumberInputE won't let me type in or paste a negative number. However using the buttons, it does go to negatives, so I'm assuming we are disallowing '-', which we shouldn't.

jpveooys:
## Related issue

Resolves #2967

## Overview

Adjust the NumberInputE input masking logic to allow the entry of negative numbers when `min` is unset or less than 0.

## Link to preview

[Storybook](https://5e25c277526d380020b5e418-arnusdcqvy.chromatic.com/?path=/docs/number-input-experimental--default)

[Non-Storybook (works in IE11)](https://red-playground.netlify.app/)

## Reason

It should be possible to type or paste negative numbers.

## Work carried out

- [x] Rework masking and validation logic to allow negative numbers
- [x] Adjust button logic to handle case of single `-` having being typed
- [x] Add error border if single `-` entered

## Demo

Screencast_26-01-22_15:01:10

## Developer notes

I looked into the possibility of using a library first. Unfortunately, the first one I found (`react-input-mask`) appears to be abandoned. The second one (`react-number-format`) had some undesirable behaviours that made it much less appealing. Therefore I went with updating our existing masking implementation.

Note that there is now the case of the input containing only a `-`. This is treated as invalid and the buttons do nothing in this case. The value sent to `onChange` is NaN in this case (I thought about other options: `null` would imply the field is empty which isn't true, and skipping the call to `onChange` would mean the parent component would still have the previous value and wouldn't know it's invalid).

Finally, I noticed the component has an `isValid` prop that doesn't seem to do anything so I created this issue:

- #3029


packages/react-component-library/src/components/NumberInputE/NumberInputE.test.tsx
```diff
@@ -262,44 +272,74 @@ describe('NumberInputE', () => {
...
-        assertInputValue('100.125')
-        assertOnChangeCall(100.125, 2)
+        assertInputValue('-')
+
+        it('does not call the onChange callback', () => {
+          expect(onChangeSpy).not.toHaveBeenCalled()
+        })
       })
 
-      describe('and a second decimal point is typed before the first', () => {
```
jpveooys, line 283: This was removed and replaced with an equivalent Cypress test as user-event's `type`/`keyboard` functions aren't realistic enough to trigger `onBeforeInput`.

packages/react-component-library/src/components/NumberInputE/useEarlyValidation.ts
```diff
@@ -1,15 +1,40 @@
...
   ) => void
+  handlePaste: (event: React.ClipboardEvent<HTMLInputElement>) => void
 } {
+  /**
+   * Block invalid pastes before the DOM is updated (to
+   * avoid the cursor jumping).
+   *
+   * Only needed for Firefox and IE.
+   */
+  const handlePaste = useCallback(
```
jpveooys, line 29: Added this to stop cursor jumping on invalid paste in Firefox and IE.

packages/react-component-library/src/components/NumberInputE/useEarlyValidation.ts
```diff
@@ -33,32 +58,18 @@ export function useEarlyValidation(): {
...
-      }
-    },
-    []
-  )
-
-  /**
-   * Block a second decimal point from being typed (before the DOM is
-   * updated, to avoid the cursor jumping).
-   */
-  const handleKeyDown = useCallback(
```
jpveooys, line 48: No longer needed as `handleBeforeInput` does this now.

packages/react-component-library/src/components/NumberInputE/useEarlyValidation.ts
```diff
@@ -1,15 +1,44 @@
...
-import { areCharactersValid } from './validation'
+import { isValueValid } from './validation'
 
-export function useEarlyValidation(): {
-  handleKeyDown: (event: React.KeyboardEvent<HTMLInputElement>) => void
+/**
+ * Given some text that is being entered in an input, work out what the
+ * new value will be based on the cursor position ond selection.
+ */
+function getNewInputValue(input: HTMLInputElement, newText: string): string {
```
jpveooys, line 9: `handleBeforeInput` uses this now so it can check the entire string instead of just the new characters (this enables it to check that the minus sign is at the beginning of the string).

packages/react-component-library/cypress/specs/NumberInputE/index.spec.ts
```diff
@@ -37,6 +37,12 @@ describe('NumberInputE', () => {
...
     ['100.invalid456', '100.456'],
+    ['100.25{home}.25', '25100.25'],
+    ['-100.123', '-100.123'],
+    ['-100.1-23', '-100.123'],
+    ['100.1-23', '100.123'],
+    ['12.34{selectall}56.78', '56.78'],
+    ['-12.34{selectall}-56.78', '-56.78'],
   ]
 
   TYPED_VALUE_CASES.forEach(([textToType, expectedValue]) => {
```
jpveooys, line 48: Unfortunately, I haven't found a way to test pasting in Cypress (typing Ctrl-V doesn't work, and you can simulate the event but that is only the event and not actually pasting).

k9dh3zij:
👍 Awesome, thanks @jpveooys

packages/react-component-library/cypress/specs/NumberInputE/index.spec.ts
m7kvqbe1: Is [this](https://egghead.io/blog/handling-copy-and-paste-in-cypress) close enough?

```javascript
cy.get(input-el).invoke('val', textToPaste)
```

packages/react-component-library/src/components/NumberInputE/NumberInputE.test.tsx
```diff
@@ -21,6 +21,16 @@ const defaultProps = {
   ) => true,
 }
 
+function paste(element: HTMLElement, text: string) {
+  // userEvent 13.x doesn't set clipboardData automatically, though
```
m7kvqbe1, line 26: 😓

m7kvqbe1 (approved):
LGTM

packages/react-component-library/cypress/specs/NumberInputE/index.spec.ts
jpveooys: Sadly not, that just manipulates that value attribute using jQuery unfortunately :disappointed:

PR #3034, opened 2022-01-28
fix(CheckboxE): Fix inability to use space to toggle state in Firefox

### LINKED ISSUE #3012: CheckboxE: Cannot toggle state using space key in Firefox
### Steps to reproduce

1. Go to https://storybook.design-system.digital.mod.uk/?path=/story/checkbox-experimental--default in Firefox
2. Focus the checkbox
3. Press the space key

The state appears to not change. The Actions tab suggests it may be being toggled twice.

This works fine in Chrome.

### Demo

Screencast_21-01-22_15:48:12

jpveooys:
## Related issue

Resolves #3012

## Overview

This fixes a bug with CheckboxE where the space key would double-toggle the input.

## Link to preview

CheckboxE:

https://5e25c277526d380020b5e418-jyjeyhqiwf.chromatic.com/?path=/docs/checkbox-experimental--default

RadioE (behaviour should be unchanged):

https://5e25c277526d380020b5e418-jyjeyhqiwf.chromatic.com/?path=/docs/radio-experimental--default

## Reason

To make sure the component works correctly using the keyboard.

## Work carried out

- [x] Add Cypress tests
- [x] Fix bug
- [x] Tidy up tests slightly
- [x] Copy fix to RadioE

### Before

Screencast_21-01-22_15:48:12

### After

Screencast_28-01-22_10:51:47

## Developer notes

RadioE shouldn't have the same problem in practice as the space key only checks the input (rather than toggles it like it does for a checkbox), and so an extra click has no effect. However, I've changed the code there as well for consistency.

### DIFF
--- a/packages/react-component-library/cypress/selectors/CheckboxE.ts
+++ b/packages/react-component-library/cypress/selectors/CheckboxE.ts
@@ -1,0 +1,4 @@
+export default {
+  input: '[data-testid="checkbox-input"]',
+  wrapper: '[data-testid="checkbox"]',
+}
--- a/packages/react-component-library/cypress/selectors/index.ts
+++ b/packages/react-component-library/cypress/selectors/index.ts
@@ -1,4 +1,5 @@
 import autocompleteE from './AutocompleteE'
+import checkboxE from './CheckboxE'
 import contextMenu from './ContextMenu'
 import datePicker from './DatePicker'
 import form from './form'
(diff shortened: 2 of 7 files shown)


m7kvqbe1 (approved):
LGTM

k9dh3zij:
Thanks @jpveooys , looks good here

PR #3053, opened 2022-02-01
Update component callbacks

### LINKED ISSUE #645: The method used to raise events on components is inconsistent
An example of this can be seen in `NumberInput`:

```
  onChange: (event: any) => void
```

A preferred approach might be:

```
  onChange: (event: React.ChangeEvent<{}>, value: number) => void
```

It is recommended that all components are audited to ensure a consistent approach. This will be a breaking change and needs to wait for `v3`.

thyhjwb6:
## Related issue
Closes #645 

## Overview
Updates component callbacks so that the event is the first parameter.

## Reason
Consistency with other callbacks and standard practice.

## Work carried out
- [x] Update callbacks

### DIFF
--- a/packages/react-component-library/src/components/Pagination/usePageChange.tsx
+++ b/packages/react-component-library/src/components/Pagination/usePageChange.tsx
@@ -1,24 +1,46 @@
 /* eslint-disable no-shadow */
-import { useState } from 'react'
+import React, { useCallback, useState } from 'react'
 
+import { OnChangeEventType } from './types'
 import { RETURN } from '../../utils/keyCodes'
 
 export const usePageChange = (
   initialPage: number,
   totalPages: number,
-  onChange?: (currentPage: number, totalPages: number) => void
-) => {
+  onChange?: (
+    event: OnChangeEventType,
+    currentPage: number,
+    totalPages: number
+  ) => void
+): {
+  onKeyDown: (event: React.KeyboardEvent) => void
+  hasError: boolean
+  onPaginationButtonClickHandler: (
+    pageNumber: number
+  ) => (event: React.MouseEvent) => void
+  currentPage: number
+} => {
   const [currentPage, setCurrentPage] = useState(initialPage)
   const [hasError, setHasError] = useState<boolean>(false)
 
-  function changePage(pageNumber: number): void {
-    setHasError(false)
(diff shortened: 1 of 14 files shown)


packages/react-component-library/src/components/CheckboxEnhanced/CheckboxEnhanced.tsx
```diff
@@ -25,7 +25,6 @@ export const CheckboxEnhanced: React.FC<CheckboxEnhancedProps> = ({
   name,
   tabIndex = 0,
   title,
-  isInvalid,
```
jpveooys: Hmm seems like `isInvalid` was/is broken here... but I guess CheckboxEnhanced doesn't matter for v3?

packages/react-component-library/src/components/Pagination/usePageChange.tsx
```diff
@@ -1,24 +1,55 @@
...
+  )
+
+  const onPaginationButtonClickHandler = useCallback(
+    (pageNumber: number) => (event: React.MouseEvent) => {
+      changePage(event, pageNumber)
+    },
+    [changePage]
+  )
+
+  // const onPaginationButtonClickHandler = useCallback(
```
jpveooys: Did you mean to leave this in?

packages/react-component-library/src/components/TabSet/TabSet.tsx
```diff
@@ -163,12 +170,7 @@ export const TabSet: React.FC<TabSetProps | ScrollableTabSetProps> = ({
         {Children.map(
           children,
           (child: React.ReactElement<TabProps>, index: number) => {
-            const {
-              children: tabChildren,
-              title,
-              isActive,
-              ...tabRest
-            } = child.props
+            const { children: tabChildren, title, ...tabRest } = child.props
```
jpveooys: Think we need to destructure `isActive` here to drop it from `tabRest` (as its set below). (Think this is the conflict as well...)

thyhjwb6: I rebased so this commit should be updated.

jpveooys (approved):

m7kvqbe1 (approved):
Nice change!

Issue #3056, opened 2022-02-02
Wrong background colour for RadioE selected state


k9dh3zij:
https://storybook.design-system.digital.mod.uk/?path=/docs/radio-experimental--default

Selected background colour should be Action 000

![image](https://user-images.githubusercontent.com/73531035/152193359-4411a74f-eebc-4113-892c-cbd55c16a730.png)

jpveooys:
This would be a good opportunity to look at breaking up some of the styles (particularly for `StyledRadio`).

Issue #3007, opened 2022-01-21
RadioE and CheckboxE in Firefox have strange sheered rendering


m7kvqbe1:
![Screenshot 2022-01-21 at 11 35 56](https://user-images.githubusercontent.com/48086589/150520572-2d36e7ca-493b-4ffc-96e4-599e34106e4c.png)
![Screenshot 2022-01-21 at 11 36 48](https://user-images.githubusercontent.com/48086589/150520631-2b6b3cba-dfcd-4522-8832-67306b714467.png)

PR #3057, opened 2022-02-02
fix(RadioE): Fix disabled styling and Firefox border glitch

### LINKED ISSUE #3052: RadioE disabled styling is incorrect
Currently there is only an unchecked disabled state which looks like this:

image

The checked disabled state is currently missing.

The disabled states should instead look like this:

image

jpveooys:
## Related issue

#3007
Resolves #3052 

## Overview

This:

- fixes a glitch with the hover and checked border in Firefox
- adjusts the size of that border to 2px
- corrects the disabled styles for both checked and unchecked states

## Link to preview

- [Storybook](https://5e25c277526d380020b5e418-jrrxzijbql.chromatic.com/?path=/docs/radio-experimental--default)
- [Non-Storybook](https://red-playground.netlify.app/)

## Reason

Make the component better and make the component line up with the design spec.

## Work carried out

- [x] Rework hover/checked border
- [x] Correct various disabled styles
- [x] Update stories

### Checked

<details>
<summary>Before and after animation</summary>

checked

</details>

<details>
<summary>Firefox, before</summary>

checked-firefox-before

</details>

<details>
<summary>Firefox, after</summary>

checked-firefox-after

</details>

#### Before

disabled-before

#### After

Screenshot 2022-02-02 at 16-50-53 Radio (Experimental) - Disabled ⋅ Storybook

#### Before

disabled-checked-before

#### After

disabled-checked-after

## Developer notes

Note: I'd recommend using the browser zoom instead of the Storybook zoom in Firefox to check this – Storybook uses a `transform` to zoom and it seems this causes glitches with a round box-shadow in Firefox.

The original Firefox glitch in the issue was due to half the border being a border and the other half being a box-shadow.

One follow-up issue has been created:

- #3056

The same Firefox glitch for CheckboxE will be resolved in a separate PR.

### DIFF
--- a/packages/react-component-library/src/components/RadioE/partials/StyledRadio.tsx
+++ b/packages/react-component-library/src/components/RadioE/partials/StyledRadio.tsx
@@ -10,7 +10,7 @@ interface StyledRadioProps {
   $isChecked?: boolean
 }
 
-const RADIO_FOCUS_WIDTH = '0.1rem'
+const RADIO_ACTIVE_BORDER_WIDTH = '2px'
 
 const { spacing, fontSize, color, animation } = selectors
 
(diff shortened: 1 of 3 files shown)


packages/react-component-library/src/components/RadioE/partials/StyledRadio.tsx
```diff
@@ -10,17 +10,33 @@ interface StyledRadioProps {
   $isChecked?: boolean
 }
 
-const RADIO_FOCUS_WIDTH = '0.1rem'
+const RADIO_ACTIVE_BORDER_WIDTH = '2px'
 
 const { spacing, fontSize, color, animation } = selectors
 
+const BackgroundColour = css<StyledRadioProps>`
```
jpveooys: Was getting a bit repetitive and confusing following the `$isDisabled` overrides, so I've used a slightly different approach here.

packages/react-component-library/src/components/RadioE/partials/StyledRadio.tsx
```diff
@@ -10,17 +10,33 @@ interface StyledRadioProps {
...
+  ${({ $isDisabled }) =>
+    $isDisabled ? color('neutral', '200') : color('action', '500')}
+`
+
+const CheckmarkCheckedFillColour = css<StyledRadioProps>`
+  ${({ $isDisabled }) =>
+    $isDisabled ? color('neutral', '200') : color('action', '500')}
+`
+
 export const StyledRadio = styled.div<StyledRadioProps>`
```
thyhjwb6, line 32: I know that this was already quite long but I do think it's a tricky maintenance point as it's effectively a function with nearly 100 lines and there are comments which sets alarm bells that it could be broken up.

thyhjwb6 (approved):
The end result looks good. I wonder if the styled component could be refactored but it's not a blocker.

👍

packages/react-component-library/src/components/RadioE/partials/StyledRadio.tsx
jpveooys: Yeah, definitely. It will need refactoring to fix #3056 so I'll add a comment to that issue

k9dh3zij:
👍  Thanks @jpveooys  looks good on Chrome! Thanks for spotting the background colour issue too! :)

Issue #2755, opened 2021-10-27
Remove `"strictNullChecks": false"` from TypeScript configuration


jpveooys:
### Description

Currently, the TypeScript configuation in `packages/react-component-library/tsconfig.json` contains:

```
    "strictNullChecks": false,
```

This means that types implicitly allow null and undefined. This is undesirable as it means:

- types are misleading
- there may be hidden bugs (either due to values being null or undefined unintentionally, or due to null or undefined not being handled)
- there may be missed test cases

PR #3069, opened 2022-02-07
Fix various strict null type errors

jpveooys:
## Related issue

#2755

## Overview

## Link to preview

https://5e25c277526d380020b5e418-ddpqvyfkig.chromatic.com/

## Reason

> 
> ```
> ```
> 
> 
> - types are misleading

## Work carried out

- [x] Resolve a number of strict null errors

## Developer notes

These were a number of straightforward errors that should be safe to fix on master (as opposed to `release/3.0.0`). Just banking these ones to reduce the risk of conflicts – the rest are to follow.


m7kvqbe1 (approved):
👍

Issue #3048, opened 2022-02-01
Investigate grid system that can compliment forms (Spike)


m7kvqbe1:
Form layouts and responsive form layouts should be made possible by using some kind of grid system. Investigate existing options and experiment.

End result should be:

- Well maintained and with good adoption
- Not re-inventing the wheel (if own implementation)
- Meet the layout criteria defined in the v3 Forms spec (where possible and sensible to do so)

thyhjwb6:
Note this is particularly evident in the `DatePickerE` example:

![image](https://user-images.githubusercontent.com/56078793/153379192-417d73b2-718f-4fbc-b97e-91a5d62b90fb.png)

PR #3070, opened 2022-02-07
refactor(ReactComponentLibrary): Remove margins

### LINKED ISSUE #748: Check external margins for components
# Overview
Components inside the Design System should not have any external margins. This is because we cannot necessarily predict where a component will be used, potentially requiring toolkit users to unset these predefined margins.

If components require separation from one another, then either utility classes or custom CSS should be added to the _instance_ of a component. This allows a toolkit users to lay out their interfaces without the toolkit maintainers making assumptions about said layouts.

### Fix
This ticket is completed when:
- [ ] All current components have had their external margins removed.
- [ ] ~Additional classes (e.g. `.form`) have been added that have situational margins for components.~

###### Note
This is a breaking change so should be included in `v3`.

thyhjwb6:
## Related issue
Closes #748 

## Overview
Removes margins from components.

## Reason
Appearance can be unpredictable for consumers.

## Work carried out
- [x] `DatePickerE`
- [x] `DescriptionList`
- [x] `NumberInputE`
- [x] `TextInputE`

## Developer notes
`CheckboxEnhanced` and `RadioEnhanced` will be removed so these margins have been left.

### DIFF
--- a/packages/react-component-library/src/components/CheckboxEnhanced/partials/StyledCheckboxEnhanced.tsx
+++ b/packages/react-component-library/src/components/CheckboxEnhanced/partials/StyledCheckboxEnhanced.tsx
@@ -14,7 +14,6 @@ export const StyledCheckboxEnhanced = styled.div`
   flex-direction: column;
   width: 22rem;
   padding: ${spacing('10')};
-  margin: ${spacing('4')};
   border-radius: 4px;
   border: 1px solid ${color('neutral', '100')};
   outline: none;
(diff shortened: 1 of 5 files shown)


packages/react-component-library/src/components/CheckboxEnhanced/partials/StyledCheckboxEnhanced.tsx
```diff
@@ -14,7 +14,6 @@ export const StyledCheckboxEnhanced = styled.div`
   flex-direction: column;
   width: 22rem;
   padding: ${spacing('10')};
-  margin: ${spacing('4')};
```
jpveooys, line 17: Were you going to leave this one (as per your comment)?

jpveooys (approved):
Changes look good :+1:  Might need to do something to the last DatePickerE story, but maybe #3048 will help with that:

image

m7kvqbe1 (approved):

PR #3077, opened 2022-02-08
Remove `isValid`

### LINKED ISSUE #3029: Remove `isValid` prop from v3 components
Various v3 components have an `isValid` prop. This is due to their prop interfaces inheriting from `InputValidationProps`.

The prop does not have any function with v3 components and so should be removed.

thyhjwb6:
## Related issue
Closes #3029 

## Overview
Removes the use of `isValid`.

## Reason
This is no longer a presentational state for our components.

## Work carried out
- [x] Remove `isValid`


m7kvqbe1 (approved):
👍

jpveooys (approved):
:+1:

PR #3078, opened 2022-02-08
Fix various strict null errors (part 2)

jpveooys:
## Related issue

#2755

## Overview

## Link to preview

https://5e25c277526d380020b5e418-gdusxcvbhe.chromatic.com/

## Reason

> 
> ```
> ```
> 
> 
> - types are misleading

## Work carried out

- [x] Resolve strict null errors for `SelectE`, `AutocompleteE`, `FloatingBox`, `DatePickerE`, `Tooltip`, `NumberInputE` and `RadioE`

## Developer notes

These should be safe to fix on master (as opposed to `release/3.0.0`).

There was a small bit of reorganising of some of the DatePickerE utility functions.

### DIFF
--- a/packages/react-component-library/src/components/AutocompleteE/hooks/useAutocomplete.ts
+++ b/packages/react-component-library/src/components/AutocompleteE/hooks/useAutocomplete.ts
@@ -19,7 +19,7 @@ export function useAutocomplete(
   ) => void
 } {
   const [hasError, setHasError] = useState<boolean>(isInvalid)
-  const inputRef = useRef<HTMLInputElement>()
+  const inputRef = useRef<HTMLInputElement>(null)
   const [items, setItems] = useState<SelectChildWithStringType[]>(children)
 
   function onInputValueChange({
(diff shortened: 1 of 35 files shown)


packages/react-component-library/src/components/AutocompleteE/hooks/useAutocomplete.ts
```diff
@@ -32,7 +32,7 @@ export function useAutocomplete(
           if (!React.isValidElement(item)) {
             return false
           }
-          const filter = inputValue.toLowerCase()
+          const filter = (inputValue as string).toLowerCase()
```
jpveooys, line 35: This looks like a mistake in the types based on the examples in the Downshift docs, so I went for a type assertion here.

packages/react-component-library/src/components/DatePickerE/utils/isDateValid.ts
```diff
@@ -0,0 +1,10 @@
+import { isValid } from 'date-fns'
+
+type DateAny = Parameters<typeof isValid>[0]
+
+/**
+ * Version of isValid with a type predicate.
+ */
+export function isDateValid(value: DateAny): value is Date | number {
```
jpveooys, line 8: Small wrapper for date-fns' `isValid` with a type predicate added (apparently the library doesn't have them because the types are generated from JSDoc).

m7kvqbe1 (approved):
LGTM

Relying on confidence in the test coverage - haven't manually tested every component behaviour, but I have had a good click about on what has been touched.

thyhjwb6 (approved):
👍

PR #3079, opened 2022-02-09
Fix various strict null errors (part 3)

jpveooys:
## Related issue

#2755

## Overview

## Link to preview

https://5e25c277526d380020b5e418-lylsvjcbge.chromatic.com/

## Reason

> 
> ```
> ```
> 
> 
> - types are misleading

## Work carried out

- [x] Resolve strict null errors for various components

## Developer notes

These should be safe to fix on master (as opposed to `release/3.0.0`).

### DIFF
--- a/packages/react-component-library/src/components/RangeSliderE/RangeSliderE.test.tsx
+++ b/packages/react-component-library/src/components/RangeSliderE/RangeSliderE.test.tsx
@@ -23,8 +23,8 @@ describe('RangeSliderE', () => {
     values: ReadonlyArray<number>,
     data: { activeHandleID: string }
   ) => void
-  let domain: number[] // lower and upper bounds
-  let values: readonly [number, number?] // initial handle values
+  let domain: readonly [number, number] // lower and upper bounds
+  let values: readonly [number] | readonly [number, number] // initial handle values
   let step: number
   let tickCount: number
 
--- a/packages/react-component-library/src/components/RangeSliderE/SliderE.tsx
+++ b/packages/react-component-library/src/components/RangeSliderE/SliderE.tsx
@@ -42,13 +42,13 @@ export interface RangeSliderEProps extends Omit<SliderProps, SliderOmitType> {
    * Two element array of numbers providing the min and max values for the slider [min, max] e.g. [0, 100].
    * It does not matter if the slider is reversed on the screen, domain is always [min, max] with min < max.
    */
-  domain?: ReadonlyArray<number>
+  domain: readonly [number, number]
   /**
    * An array of numbers. You can supply one for a value slider, two for a range slider.
    * The values should correspond to valid step values in the domain.
    * The numbers will be forced into the domain if they are two small or large.
    */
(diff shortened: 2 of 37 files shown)


packages/react-component-library/src/components/RangeSliderE/RangeSliderE.test.tsx
```diff
@@ -23,8 +23,8 @@ describe('RangeSliderE', () => {
     values: ReadonlyArray<number>,
     data: { activeHandleID: string }
   ) => void
-  let domain: number[] // lower and upper bounds
-  let values: readonly [number, number?] // initial handle values
+  let domain: readonly [number, number] // lower and upper bounds
```
jpveooys, line 26: `readonly [number, number?]` allowed `[number, undefined]`, which was incompatible with `readonly number[]`.

packages/react-component-library/src/components/RangeSliderE/SliderE.tsx
```diff
@@ -42,13 +42,13 @@ export interface RangeSliderEProps extends Omit<SliderProps, SliderOmitType> {
    * Two element array of numbers providing the min and max values for the slider [min, max] e.g. [0, 100].
    * It does not matter if the slider is reversed on the screen, domain is always [min, max] with min < max.
    */
-  domain?: ReadonlyArray<number>
+  domain: readonly [number, number]
```
jpveooys, line 45: The component breaks without `domain` so I made this required.

packages/react-component-library/src/components/TopLevelNavigation/Masthead/useMastheadSearch.ts
```diff
@@ -8,13 +8,17 @@ export function useMastheadSearch() {
...
+
     event.currentTarget.blur()
     const newShowSearch = !showSearch
 
     // if opening the searchbar then get the container width and set that
     // as the width of the searchbar so that it does not spill
     // over to other parts of the page, such as the sidebar.
-    if (newShowSearch === true) {
+    if (newShowSearch) {
       setContainerWidth(mastheadContainerRef.current.offsetWidth)
```
jpveooys, line 22: I noticed this doesn't react to container/viewport width changes – but out of scope for this issue.

packages/react-component-library/src/components/RangeSliderE/RangeSliderE.test.tsx
```diff
@@ -23,8 +23,8 @@ describe('RangeSliderE', () => {
...
+  let values: readonly [number] | readonly [number, number] // initial handle values
```
jpveooys, line 27: `readonly [number, number?]` allowed `[number, undefined]`, which was incompatible with `readonly number[]`.

m7kvqbe1: tupletastic

m7kvqbe1 (approved):

thyhjwb6 (approved):
👍

PR #3080, opened 2022-02-10
fix(Modal): Stop regenerating IDs on every render

### LINKED ISSUE #3059: Modal: New Aria ID are generated on each render
The Aria key don't seem to be stable between re-renders. 

https://github.com/defencedigital/mod-uk-design-system/blob/99218f56c00d1fe783ada7ec3dd8051634ec9795/packages/react-component-library/src/components/Modal/Modal.tsx#L85

I've attached a video and a code sandbox 

https://user-images.githubusercontent.com/75253719/152205419-cb31db61-c179-45cd-a188-16337d8bf03d.mp4

https://codesandbox.io/s/silly-platform-bt0mo?file=/src/App.js:0-467

Update: Ive used the titleId and the descriptionId to stop this but, I can see this has been marked as private props.

jpveooys:
## Related issue

Resolves #3059
#2755 

## Overview

This updates `Modal` to use the `useExternalId` hook to avoid regenerating title and body IDs on every render.

## Link to preview

[Example app](https://red-playground.netlify.app)

(Note: I have used a standard `<input>` to isolate Modal from other components.)

[Storybook](https://5e25c277526d380020b5e418-tyunscyfnm.chromatic.com/?path=/docs/modal--default)

## Reason

IDs and aria keys should be stable between renders.

## Work carried out

- [x] Memoise Modal IDs
- [x] Fix Modal strict null errors

### Before

Screencast_10-02-22_13:16:40

### After

Screencast_10-02-22_13:18:45

## Developer notes

A number of other components have the same problem. I've created #3081 as a follow-up for other components and have marked `getId()` as deprecated to stop the same problem being introduced in new components. (I've updated existing uses that were already memoised as those should be safe to change without further testing.)

Strict null errors in the `Modal` component were also fixed.

### DIFF
--- a/packages/react-component-library/src/helpers.ts
+++ b/packages/react-component-library/src/helpers.ts
@@ -14,6 +14,9 @@ function getKey(prefix: string, suffix: string | number): string {
   return `${prefix}-${suffix}`.replace(/\s/g, '')
 }
 
+/**
+ * @deprecated Use the useExternalId() hook instead as it's memoised.
+ */
 function getId(prefix: string): string {
   return getKey(prefix, uuidv4())
 }
--- a/packages/react-component-library/src/components/DatePickerE/DatePickerE.tsx
+++ b/packages/react-component-library/src/components/DatePickerE/DatePickerE.tsx
@@ -182,8 +182,8 @@ export const DatePickerE: React.FC<DatePickerEProps> = ({
   ...rest
 }) => {
   const id = useExternalId(externalId)
-  const titleId = `datepicker-title-${useExternalId()}`
-  const contentId = `datepicker-contentId-${useExternalId()}`
+  const titleId = useExternalId('datepicker-title')
+  const contentId = useExternalId('datepicker-contentId')
   const buttonRef = useRef<HTMLButtonElement>(null)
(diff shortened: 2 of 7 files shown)


packages/react-component-library/src/helpers.ts
```diff
@@ -14,6 +14,9 @@ function getKey(prefix: string, suffix: string | number): string {
   return `${prefix}-${suffix}`.replace(/\s/g, '')
 }
 
+/**
+ * @deprecated Use the useExternalId() hook instead as it's memoised.
+ */
 function getId(prefix: string): string {
```
thyhjwb6, line 20: Is it easy enough to get rid of this whilst we are at it?

thyhjwb6 (approved):
Looks good 👍

c9qsm62x:
Looks great, thank you!

packages/react-component-library/src/helpers.ts
jpveooys: Created #3081 as a follow-up – shouldn't be too hard but it's used in quite a few places (would want to test form components a bit more carefully).

PR #3087, opened 2022-02-14
Fix various strict null errors (part 4)

jpveooys:
## Related issue

#2755

## Overview

This fixes a subset of type errors on `release/3.0.0` that were being suppressed by `"strictNullChecks": false`, and enables strict null checks for the icon library and design tokens packages.

## Link to preview

https://5e25c277526d380020b5e418-xtetyxjycj.chromatic.com/

## Reason

> 
> ```
> ```
> 
> 
> - types are misleading

## Work carried out

- [x] Resolve strict null errors for `DescriptionList`, `DismissibleBanner` icon library and design tokens
- [x] Turn on strict null checks for the icon library and design tokens packages
- [x] Improve error handling in the design tokens package

## Developer notes

n/a

### DIFF
--- a/packages/icon-library/tsconfig.json
+++ b/packages/icon-library/tsconfig.json
@@ -14,7 +14,6 @@
     "isolatedModules": true,
     "noEmit": true,
     "jsx": "preserve",
-    "strictNullChecks": false,
     "suppressImplicitAnyIndexErrors": true
   },
   "include": ["src"]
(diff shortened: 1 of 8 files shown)


packages/icon-library/tsconfig.json
```diff
@@ -14,7 +14,6 @@
     "isolatedModules": true,
     "noEmit": true,
     "jsx": "preserve",
-    "strictNullChecks": false,
```
m7kvqbe1, line 17: 🎉

m7kvqbe1 (approved):
LGTM

PR #3088, opened 2022-02-14
Fix various strict null errors (part 5)

jpveooys:
## Related issue

#2755

## Overview

This resolves all strict null errors in the Timeline component.

`startDate` was made required as the component breaks without it. Also, `today` now defaults to the current date to match the docs.

## Link to preview

https://5e25c277526d380020b5e418-lkcmkjloqn.chromatic.com/?path=/story/compound-timeline--default

## Reason

> 
> ```
> ```
> 
> 
> - types are misleading

## Work carried out

- [x] Resolve strict null errors for `Timeline`
- [x] Default `today` to current date

## Developer notes

The docs for `today` say:

> Today's current date - default is the system date time.

However, if today wasn't specified, it was previously showing the marker at the start of every month:

image

I've fixed this so the behaviour matches the docs.

Open to any ideas for any other additional tests that could be written.

### DIFF
--- a/packages/react-component-library/src/components/Timeline/Timeline.tsx
+++ b/packages/react-component-library/src/components/Timeline/Timeline.tsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import { isAfter } from 'date-fns'
+import { isAfter, setHours, startOfToday } from 'date-fns'
 
 import { ComponentWithClass } from '../../common/ComponentWithClass'
 import { TimelineProvider } from './context'
@@ -40,7 +40,7 @@ export interface TimelineProps extends ComponentWithClass {
   /**
    * A month will display either side of this start date.
    */
-  startDate?: Date
+  startDate: Date
   /**
    * Bound the timeline by the specified start and end dates.
    */
@@ -74,7 +74,7 @@ function getHoursBlockSize(
   )
 }
 
-function getEndDate(startDate: Date, endDate: Date) {
+function getEndDate(startDate: Date, endDate: Date | null) {
   if (startDate && endDate && isAfter(startDate, endDate)) {
     logger.error('`startDate` is after `endDate`')
     return null
@@ -92,9 +92,9 @@ export const Timeline: React.FC<TimelineProps> = ({
   hideToolbar = false,
   isFullWidth = false,
   startDate,
-  endDate,
-  today,
-  range,
+  endDate = null,
+  today = setHours(startOfToday(), 12),
+  range = 1,
   unitWidth,
   ...rest
 }) => {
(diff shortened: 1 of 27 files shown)


packages/react-component-library/src/components/Timeline/Timeline.tsx
```diff
@@ -40,7 +40,7 @@ export interface TimelineProps extends ComponentWithClass {
   /**
    * A month will display either side of this start date.
    */
-  startDate?: Date
+  startDate: Date
```
jpveooys, line 43: This was made required as the component breaks without it.

packages/react-component-library/src/components/Timeline/hooks/useTimelineZoom.ts
```diff
@@ -14,19 +14,20 @@ export function useTimelineZoom(): {
...
-  const canZoomIn = isNil(currentScaleIndex) || currentScaleIndex > 0
-  const canZoomOut = currentScaleIndex < scaleOptions.length - 1
+  const canZoomIn = !isNil(currentScaleIndex) && currentScaleIndex > 0
+  const canZoomOut =
+    !isNil(currentScaleIndex) && currentScaleIndex < scaleOptions.length - 1
 
   function zoomIn() {
     dispatch({
-      scaleIndex: currentScaleIndex - 1,
+      getScaleIndex: (scaleIndex: number) => scaleIndex - 1,
```
jpveooys, line 23: Changed this to simplify dealing with a null `currentScaleIndex` – but it also makes more sense for it to be either an offset or function as doing the calculation here would do the wrong thing if two zooms were dispatched at once.

(I went for a function as it's more flexible, and the context is public so someone could be using it in theory.)

packages/react-component-library/src/components/Timeline/Timeline.tsx
```diff
@@ -92,9 +92,9 @@ export const Timeline: React.FC<TimelineProps> = ({
   hideToolbar = false,
   isFullWidth = false,
   startDate,
-  endDate,
-  today,
-  range,
+  endDate = null,
+  today = startOfToday(),
```
jpveooys: Waiting for confirmation from @k9dh3zij if this should be the start of the day or midday (affects where the marker appears).

k9dh3zij: Let's go for midday for now, please. I've raised #3095 to look into this further, as I think there are better ways of indicating the current day

m7kvqbe1 (approved):
👍

packages/react-component-library/src/components/Timeline/hooks/useTimelineZoom.ts
m7kvqbe1: I think the microsite docs will probably also need updating:

https://design-system.digital.mod.uk/timeline

packages/react-component-library/src/components/Timeline/Timeline.tsx
m7kvqbe1: I think the microsite docs will probably also need updating:

https://design-system.digital.mod.uk/timeline

jpveooys: Oh yes! This one was already correct on there but some of the other stuff will need updating. Also think the render prop stuff will need updating on there. Will create a new issue...

PR #3098, opened 2022-02-16
feat(TabSet): Improve scrolling behaviour

### LINKED ISSUE #3096: Improve scrollable TabSet behaviour
Currently the scrollable TabSet scrolling behaviour is different for left and right scrolling, and right scrolling is slow:

Screencast_15-02-22_14:36:23

The inconsistent logic is also preventing fixing #2755.

Fast right scrolling should be allowed as it is for left scrolling.

jpveooys:
## Related issue

Resolves #3096
#2755 

## Overview

This:
- improves TabSet scrolling behaviour so that repeated clicks are quick for both scrolling left and right
- resolves problems with scrolling getting stuck after resizing the viewport, or sometimes in Firefox when scrolling left
- fixes strict null errors in the TabSet component

The scrolling tests were rewritten in Cypress as layout isn't supported by jsdom.

## Link to preview

https://tab-set-scrollable.netlify.app/?path=/docs/tab-set--scrollable

## Reason

Fix strict null errors while improving the scrolling behaviour at the same time.

## Work carried out

- [x] Rewrite scrolling logic
- [x] Fix strict null errors
- [x] Add new scrolling tests in Cypress

#### Slow right scrolling

slow-right

#### Left scrolling stuck in Firefox

left-stuck

#### Stuck after resizing

resize-stuck

### After

after-2

## Developer notes

My main aim was to fix the strict null errors, but trying to do this in the old logic exacerbated the problems mentioned above, so I wrote new logic to avoid making things worse (while otherwise behaving similarly).

The scrolling tests were rewritten in Cypress as layout isn't supported by jsdom (the old tests were mocking various scrolling and layout DOM stuff, which broke after changing the scrolling logic).

`use-resize-observer` was used to fix the resize behaviour – a `window.ResizeObserver` polyfill is needed for old browsers and Jest.

### DIFF
--- a/packages/cra-template-defencedigital/template.json
+++ b/packages/cra-template-defencedigital/template.json
@@ -16,6 +16,7 @@
       "@graphql-codegen/typescript": "^2.4.1",
       "@graphql-codegen/typescript-operations": "^2.2.1",
       "@graphql-codegen/typescript-react-apollo": "^3.2.2",
+      "@juggle/resize-observer": "^3.3.1",
       "@testing-library/jest-dom": "^5.5.0",
       "@testing-library/react": "^12.1.2",
       "@testing-library/user-event": "^13.5.0",
--- a/packages/cra-template-defencedigital/template/src/setupTests.ts
+++ b/packages/cra-template-defencedigital/template/src/setupTests.ts
@@ -2,5 +2,8 @@
 // allows you to do things like:
 // expect(element).toHaveTextContent(/react/i)
 // learn more: https://github.com/testing-library/jest-dom
-import '@testing-library/jest-dom';
+import '@testing-library/jest-dom'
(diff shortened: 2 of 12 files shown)


k9dh3zij:
👍  LGTM, right scrolling much smoother now! Couldn't break it. Thanks

m7kvqbe1 (approved):
LGTM!

Does it makes sense for us to bundle the polyfill and do the global assignment in the component module if we support IE11?

jpveooys:
> Does it makes sense for us to bundle the polyfill and do the global assignment in the component module if we support IE11?

I think we could explore providing a separate bundle of polyfills if it would be useful (rather than trying to do it in our main bundle). For now the CRA template has been updated and I've added a note to the v3 migration docs.

PR #3106, opened 2022-02-17
feat(ReactComponentLibrary): Turn on strict null checks

### LINKED ISSUE #2755: Remove `"strictNullChecks": false"` from TypeScript configuration

jpveooys:
## Related issue

Resolves #2755

Depends on #3098

## Overview

This change turns on strict null checks for the component library by removing `"strictNullChecks": false` from `packages/react-component-library/tsconfig.json`.

## Link to preview

https://5e25c277526d380020b5e418-xixshjtsou.chromatic.com/

## Reason

Having strict null checks off means that types implicitly allow null and undefined. This is undesirable as it means:

- types are misleading for users of the component library
- there may be hidden bugs (either due to values being null or undefined unintentionally, or due to null or undefined not being handled)
- there may be missed test cases

## Work carried out

- [x] Resolve remaining strict null errors
- [x] Ignore errors in components pending removal
- [x] Remove ``"strictNullChecks": false``

## Developer notes

There should be no more references to `strictNullChecks` in the repo.

Errors in `Button`, `CheckboxEnhanced`, `DatePicker`, `NumberInput`, `RadioEnhanced`, `RangeSlider`, `Select` and `Switch` have been ignored as these are deprecated components for which there is a separate piece of work in progress to remove them.

### DIFF
--- a/packages/react-component-library/src/components/Field/Field.tsx
+++ b/packages/react-component-library/src/components/Field/Field.tsx
@@ -7,7 +7,7 @@ import { StyledHintText } from './partials/StyledHintText'
 import { getId } from '../../helpers'
 
 export type ErrorType = {
-  error: boolean | string
+  error: boolean | string | null | undefined
 }
 
 export interface FieldProps extends ComponentWithClass {
--- a/packages/react-component-library/src/components/TopLevelNavigation/Masthead/Masthead.stories.tsx
+++ b/packages/react-component-library/src/components/TopLevelNavigation/Masthead/Masthead.stories.tsx
@@ -117,7 +117,7 @@ export const CustomLogo: ComponentStory<typeof Masthead> = (props) => (
 
 CustomLogo.storyName = 'Custom logo'
 CustomLogo.args = {
-  onSearch: null as undefined,
+  onSearch: null,
 }
 
 export const WithoutLogo: ComponentStory<typeof Masthead> = (props) => (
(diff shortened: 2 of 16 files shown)


packages/react-component-library/src/components/Field/Field.tsx
```diff
@@ -7,7 +7,7 @@ import { StyledHintText } from './partials/StyledHintText'
 import { getId } from '../../helpers'
 
 export type ErrorType = {
-  error: boolean | string
+  error: boolean | string | null | undefined
```
jpveooys, line 10: `undefined` was coming from `meta.error` in the Formik example – I think it's reasonable to allow `null` and `undefined` as they could be being used if errors are being held as strings.

packages/react-component-library/src/components/TopLevelNavigation/Masthead/Masthead.stories.tsx
```diff
@@ -118,7 +118,7 @@ export const CustomLogo: ComponentStory<typeof Masthead> = (props) => (
 
 CustomLogo.storyName = 'Custom logo'
 CustomLogo.args = {
-  onSearch: null as undefined,
+  onSearch: null,
```
jpveooys, line 120: I may have commented on this on a previous PR – the problem is that if `onSearch` is set, the component shows the search button, and Storybook is injecting a value for `onSearch` if it's undefined (so these stories incorrectly end up with the search button).

Probably it'd be better if the component had a separate `hasSearch` prop but I've just allowed `null` for `onSearch` for now (`null as undefined` was still a type error).

m7kvqbe1 (approved):
🎉

PR #3119, opened 2022-02-22
feat(ContextMenu): Intelligently position context menu

### LINKED ISSUE #2465: Refactor `ContextMenu` to intelligently position based on screen real estate
Recent PR #2412 means that we can now position the `ContextMenu` based on available space so that it is always visible on the screen.

m7kvqbe1:
## Related issue

Closes #2465

## Overview

Intelligently position the `ContextMenu` based on the available screen real-estate.

## Link to preview

https://6214cbcefa42c7b48aad1817--optimistic-williams-bf676e.netlify.app

## Reason

>Remove the need for the consumer to be explicit about the positioning of the menu.

## Work carried out

- [x] ~~Rename `useHideShow` hook `isOpen` param to `initialIsOpen`~~
- [x] Improve the behaviour of the `useClickMenu` hook and leverage `useFloatingElement`


packages/react-component-library/src/hooks/useClickMenu.ts
```diff
@@ -1,118 +1,91 @@
...
+      return {
+        top: clientY,
+        right: clientX,
+        bottom: clientY,
+        left: clientX,
+        width: 0,
+        height: 0,
+      }
+    },
+  } as VirtualElement
```
jpveooys: I think doing the type assertion on DOMRect would make it a bit clearer why it's there:

```suggestion
      } as DOMRect
    },
  }
```

jpveooys (commented):
Works well from testing :+1: 

There's a build error in the smoke test fyi.

Also, I'm not sure about `initialIsOpen` in all of these cases – some of them try to react to the prop changing as well I think (there's a `useEffect` in `useOpenClose`).

jpveooys (approved):
I still have this question:

> Also, I'm not sure about initialIsOpen in all of these cases – some of them try to react to the prop changing as well I think (there's a useEffect in useOpenClose).

Otherwise looks good :+1:

m7kvqbe1:
Dropped the `initialIsOpen` change as the `useEffect` hook means it's in most cases reacting to the prop change.

jpveooys (approved):

PR #3085, opened 2022-02-11
feat(Forms): Add `Field`, `Fieldset` and `SectionDivider`

m7kvqbe1:
## Related issue

Closes #3047

## Overview

Implement `Field`, `Fieldset` and `SectionDivider` components that apply error messages, hint text and opinionated layout related styles.

## Link to preview

https://620b9e01a230900a5c420aae--hopeful-shirley-e606db.netlify.app

## Reason

>We need a way of decorating components used in forms with extras so that they can be consumed downstream in a variety of different ways (formik, react-hook-form, native, etc).

## Work carried out

- [x] Create and expose new `Field` component
- [x] Update forms examples
- [x] Remove deprecated `withFormik` HOC and `FormikGroup`
- [x] Style `Fieldset` and add legend
- [x] Create `SectionDivider` component

## Screenshot

Screenshot 2022-02-14 at 11 47 50

## Developer notes

- Happy to split this out into smaller PRs if you would prefer. 

- There is additional ticketed work that iterates upon what has been implemented here. 

- This is intended to get us to a point where we can ship the next major.

PR #3132, opened 2022-03-02
Remove outdated Formik references

jpveooys:
## Related issue

n/a

## Overview

This:
- removes Formik references from the main README and the component library README
- removes Formik as a dependency in the CRA template

## Reason

The dependency on Formik was removed in #3085 (and hence it's no longer required).

## Work carried out

- [x] Update READMEs
- [x] Update CRA template

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,13 +29,13 @@ To install and save to your project's package.json dependencies, run:
 
 ```
 # with npm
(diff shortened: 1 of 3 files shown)


thyhjwb6 (approved):
👍

PR #3136, opened 2022-03-04
fix(Radio): Resolve incorrect active container background color

### LINKED ISSUE #3056: Wrong background colour for RadioE selected state

### LINKED ISSUE #3135: Abstract and consolidate Checkbox / Radio styling
These components are visually very similar but they each have their own sets of styles.

The code for these styles appears to have diverged quite radically.

It would be good to abstract and consolidate these styles.

m7kvqbe1:
## Related issue

Closes #3056 
Closes #3135

## Overview

Refactor and resolve incorrect `Radio` component container variant active state styling.

## Link to preview

https://62275b4f39eb3d0cb221031d--admiring-allen-740480.netlify.app

## Reason

>Needs to be styled in line with spec and to match `Checkbox`.

## Work carried out

- [x] Supplement automated tests
- [x] Refactor Checkbox and Radio implementations
- [x] Resolve styling
- [x] Handle implicit deselection of radio within collection


jpveooys (commented):
Seems to be a problem with the radio groups in the example stories (my screen recorder isn't working at the moment :disappointed: ):

After selecting the first then second radio:

image

And then repeating:

image

jpveooys:
@m7kvqbe1 FYI there's a type error in the smoke test...

packages/react-component-library/src/components/CheckboxRadioBase/CheckboxRadioBaseProps.ts
```diff
@@ -0,0 +1,56 @@
...
+   */
+  variant?: CheckboxRadioVariantType
+  /**
+   * HTML `type` attribute associated with the input (Checkbox or Radio).
+   */
+  type: 'checkbox' | 'radio'
+  /**
+   * Custom implementations of presentational styled-components.
+   */
+  partials: Record<'root' | 'checkmark', React.ReactElement>
```
jpveooys: Are we able to pass the components themselves through (as `React.ComponentType`) so we don't need to use `React.cloneElement`?

packages/react-component-library/src/components/CheckboxRadioBase/CheckboxRadioBase.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+    const handleKeyUp = (_: React.KeyboardEvent) => {
+      localRef.current?.focus()
+    }
+
+    const handleOnChange = (e: React.FormEvent<HTMLInputElement>) => {
+      setIsChecked(!isChecked)
+
+      if (type === 'radio') {
+        const event = new CustomEvent('_radio:selected', {
```
jpveooys: Would be good to include some vendor namespace I think – e.g. `_mod-uk-ds:radio:selected`

packages/react-component-library/src/components/CheckboxRadioBase/CheckboxRadioBase.tsx
```diff
@@ -0,0 +1,144 @@
...
+        const isDifferent = emittedRef !== localRef
+        const isInCollection = emittedRef.current.attributes.name.value === name
+        const isRadio = type === 'radio'
+
+        if (isDifferent && isInCollection && isRadio) {
+          setIsChecked(false)
+        }
+      }
+
+      document.addEventListener('_radio:selected', handleSelected)
```
jpveooys: Just a note for reference really, I think `document.addEventListener('change', handleSelected, {capture: true})` could've been made to work too with some checks on the target (but the custom event works fine so let's stick with that).

jpveooys (commented):
Nice work! Couple of other comments above.

jpveooys (approved):

PR #3137, opened 2022-03-04
Ensure element and ARIA IDs are stable between renders

### LINKED ISSUE #3081: Element and ARIA IDs should be stable between renders
A number of components generate new IDs (used for the `id` and various ARIA attributes) on every render. These should be made stable by using memoisation (e.g. the `useExternalId` hook).

This was reported in #3059 for the Modal component but it applies to several other components too.

The `getId()` function should be removed at the same time.

jpveooys:
## Related issue

Resolves #3081

## Overview

This ensures all `id` attributes (and hence related ARIA attributes) are stable between renders of a component

## Link to preview

https://5e25c277526d380020b5e418-kaqafeyiho.chromatic.com/

## Reason

It's undesirable to have the element ID and ARIA IDs changing unnecessarily.

## Work carried out

- [x] Replace uses of `getId()` and `uuid.v4()` for `id` attributes with `useExternalId()`
- [x] Add additional tests
- [x] Remove `getId()`

## Developer notes

n/a

### DIFF
--- a/packages/react-component-library/src/components/Alert/Alert.test.tsx
+++ b/packages/react-component-library/src/components/Alert/Alert.test.tsx
@@ -155,4 +155,35 @@ describe('Alert', () => {
       })
     })
   })
+
+  describe('when the Alert content changes', () => {
+    let initialContentId: string
+    let initialTitleId: string | null
+
+    const ExampleAlert = ({ content }: { content: string }) => (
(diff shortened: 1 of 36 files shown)


thyhjwb6 (approved):
👍

m7kvqbe1 (approved):

PR #3156, opened 2022-03-14
feat(Select): Add ability to hide clear button

### LINKED ISSUE #3151: Optional clearable button
Hi
In our app we have a select component that should not be clearable in v2 of the app we used react-select isClearable boolean. Could we add this to v3 please. 

https://github.com/defencedigital/mod-uk-design-system/blob/6a830d9ab465bf60e9f158a0b06e9a2a3d2227fd/packages/react-component-library/src/components/SelectBase/SelectLayout.tsx#L117-L120

jpveooys:
## Related issue

Resolves #3151

## Overview

This adds a new prop `hideClearButton` to `Select` and `Autocomplete`. If set, this the clear button is hidden.

## Link to preview

https://red-playground.netlify.app/?path=/docs/autocomplete--no-clear-button

## Reason

Required by a downstream app to replicate functionality that was in v2.

## Work carried out

- [x] Add `hideClearButton` prop to `Select` and `Autocomplete`

## Screenshot

image

## Developer notes

~~The existing default of being clearable is kept, and hence the prop is named `isUnclearable` (to avoid having to do `isClearable={false}`).~~

~~Open to any other suggestions for what to name the prop etc.~~

### DIFF
--- a/packages/react-component-library/src/components/Autocomplete/Autocomplete.stories.tsx
+++ b/packages/react-component-library/src/components/Autocomplete/Autocomplete.stories.tsx
@@ -31,7 +31,7 @@ const Template: ComponentStory<typeof Autocomplete> = (args) => (
   </div>
 )
 
-const TemplateWIthIconsAndBadges: ComponentStory<typeof Autocomplete> = (
+const TemplateWithIconsAndBadges: ComponentStory<typeof Autocomplete> = (
   args
 ) => (
   <div style={{ height: args.isDisabled ? 'initial' : '18rem' }}>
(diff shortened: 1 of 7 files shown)


thyhjwb6 (approved):
👍

k9dh3zij:
Thanks @jpveooys lgtm in terms of functionality.

I wonder if calling it `isUnclearable` is somewhat misleading, given that the user can manually clear the entry using backspace. How about `hideClearButton` or something similar? (and maybe worth noting that "The user can still clear using the Backspace key")

jpveooys:
Thanks @k9dh3zij , I've renamed it to `hideClearButton` and updated the docs for `Autocomplete` (but not `Select` as the text can't be deleted there).

Issue #2716, opened 2021-10-19
(Spike) Investigate switching from yarn v1 to npm v8 or yarn v3


jpveooys:
### Description

Spike to investigate migrating from yarn v1 to npm v8 or yarn v3

Yarn v1 is now only receiving security fixes and so isn't a good long-term option for us.

npm now has support for workspaces so may be able to migrate to it.

jpveooys:
I've added notes from the spike here: https://gist.github.com/jpveooys/c6a1ba37395744d92a92df2b888f9532

jpveooys:
Closing this – findings outlined in the linked gist. We'll look to migrate to npm once some outstanding problems are resolved.

PR #3168, opened 2022-03-17
Fix incorrect dependencies and remove CSS loaders

### LINKED ISSUE #3166: Errors in dependency tree
#2716 identified that there are various errors in the dependency tree. For example, `lodash` is missing as a dependency of `packages/react-component-library` while `packages/design-tokens` is missing `react`. 

These only worked because the packages were being installed by something else and picked up from the root `node_modules`.

Additionally, there are various errors relating to peer dependencies.

These errors should all be resolved (particularly before updating to a newer package manager).

jpveooys:
## Related issue

Resolves #3166

## Overview

This:

- adds various missing dependencies
- removes some unused dependencies
- removes CSS-related loaders from `packages/react-component-library`
- moves the `lint-staged` dependency to the workspace root and stops running it in parallel
- moves a couple of dependencies to devDependencies
- remove out-of-date dependabot ignore rules

## Link to preview

- [Storybook](https://5e25c277526d380020b5e418-wjawwotdwg.chromatic.com/)

## Reason

To:

- ensure dependencies are correctly specified
- smooth an upgrade to a newer package manager
- remove dependencies that are no longer needed
- ensure lint-staged works when run from the root (with newer package managers) – see also https://github.com/okonet/lint-staged#how-to-use-lint-staged-in-a-multi-package-monorepo 
- fix lint-staged Git errors due to locking (see https://github.com/okonet/lint-staged/issues/1105#issuecomment-1039140085)

(These were discovered during #2716.)

## Work carried out

- [x] Fix incorrect dependencies in packages/design-tokens
- [x] Fix incorrect dependencies in packages/react-component-library
- [x] Fix incorrect dependencies in packages/icon-library
- [x] Hoist lint-staged dependency
- [x] Remove CSS-related loaders

## Developer notes

n/a


packages/design-tokens/package.json
```diff
@@ -37,24 +37,29 @@
...
+    "@types/jest": "^27.4.1",
     "@types/lodash": "^4.14.161",
+    "@types/react": "^17.0.38",
+    "@types/styled-components": "^5.1.24",
     "babel-loader": "^8.1.0",
     "clean-webpack-plugin": "^4.0.0",
     "eslint": "^8.2.0",
     "jest": "^27.3.1",
     "jest-styled-components": "^7.0.3",
-    "lodash": "^4.17.20",
```
jpveooys, line 47: This is already in `dependencies` below.

thyhjwb6 (approved):
👍

PR #3187, opened 2022-03-22
chore(Security): Update minimist and downgrade usehooks-ts

jpveooys:
## Overview

This updates minimist to version 1.2.6 to resolve a security alert.

It also downgrades usehooks-ts [as dependabot updated it yesterday](https://github.com/defencedigital/mod-uk-design-system/pull/3180) but newer versions aren't compatible with IE11.

## Developer notes

See https://github.com/substack/minimist#security for confirmation of the security fix.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,3 +13,7 @@ updates:
   commit-message:
     prefix: chore(Security)
   rebase-strategy: auto
+  ignore:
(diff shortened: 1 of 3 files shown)


jpveooys:
~~Hmm, looks like this breaks IE11.~~

m7kvqbe1 (approved):

thyhjwb6 (approved):
👍

PR #3189, opened 2022-03-22
fix(NumberInputE): Deprecate `min` and `max`

### LINKED ISSUE #3031: Setting min/max to NumberInput may result in user unable to provide valid input
As discussed with @jpveooys :

Set min=1000 and user can't type anything

This has other UX implications as well:

Set min=0. Input won't allow user to type in a '-'. However the reason may not be obvious to the user if their app interface doesn't make it clear that there is a minimum of 0.

Set max = 500. If user types 999, it will stop at 99. (which the user may not notice)

This increases the change for mistakes happening with numbers eg if a user starts typing "-500" but ends up with "500" instead, because of a min=0. This could be very bad indeed. 

My preference in this case would be to accept the (valid number) input and then error it with a message like "Please enter a number between x and y" or "over z" etc.

m7kvqbe1:
## Related issue

Closes #3031

## Overview

App-side validation should be used (ideally triggered `onBlur`) instead of this feature.

## Reason

>After some discussion we have come to the conclusion that responsibility for validating min and max input should be devolved to the consuming application (as with all other validation).

## Work carried out

- [x] Mark as deprecated

## Developer notes

A seperate ticket (https://github.com/defencedigital/mod-uk-design-system/issues/3188) has been created to look into potentially providing a validation library that allows consumers to apply common and consistent validation rules across applications.

### DIFF
--- a/packages/react-component-library/src/components/NumberInput/NumberInput.tsx
+++ b/packages/react-component-library/src/components/NumberInput/NumberInput.tsx
@@ -45,6 +45,10 @@ interface NumberInputBaseProps
   label?: string
   /**
    * Maximum value selectable by the component (upper bound).
+   *
+   * Deprecated: Use app-side validation (ideally triggered `onBlur`).
+   *
+   * @deprecated
    */
   max?: number
   /**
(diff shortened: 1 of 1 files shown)


thyhjwb6 (approved):
👍 

does this make our first v4 ticket? 😂

jpveooys (approved):

PR #3197, opened 2022-03-25
fix(SelectBase): Resolve inconsistent popover in Chrome

### LINKED ISSUE #3169: Select popover for options that overflow container does not always appear
Can't see anything that may be causing it - looks random, but is reproducible. 
Happens inside the default Storybook container, as well as when opened in a new tab.
On latest Chrome on Mac (99.0.4844.74)

https://user-images.githubusercontent.com/73531035/158871604-658f1b15-97fd-43a3-a566-360bd67cd678.mov

m7kvqbe1:
## Related issue

Closes #3169

## Overview

Resolve `SelectBase` inconsistent popover behaviour.

## Reason

>In Chrome when slowly hovering over the truncated selected option the Popover does not display.

## Work carried out

- [x] Extend `FloatingBox` styles

## Screenshot

Before

After


jpveooys (approved):
:+1:

thyhjwb6 (approved):
👍

PR #3198, opened 2022-03-25
feat(Timeline): Change toolbar button variant to tertiary

### LINKED ISSUE #3134: Change timeline toolbar buttons from secondary to tertiary
After the v3 button style update, our secondary buttons have a stronger visual style, that is overpowering when used in the timeline toolbar. 

From an IA point of view it also makes sense that these buttons should be tertiary, not secondary. The issue in the past was that our tertiary button visual style looked like a link rather than a button.

**This ticket is to change all timeline toolbar buttons from secondary to tertiary.**

Current state, after v3 update to button styles:

image

m7kvqbe1:
## Related issue

Closes #3134

## Overview

Change `Timeline` toolbar button variant to tertiary.

## Reason

>After the v3 button style update, our secondary buttons have a stronger visual style, that is overpowering when used in the timeline toolbar.
>
>From an IA point of view it also makes sense that these buttons should be tertiary, not secondary. The issue in the past was that our tertiary button visual style looked like a link rather than a button.

## Work carried out

- [x] Change Timeline toolbar button variant
- [x] Change DatePicker sheet button variant
- [x] Rename DatePicker reducer action constant

## Screenshot

Screenshot 2022-03-25 at 11 21 32

## Developer notes

Should these buttons also be tertiary?

Screenshot 2022-03-25 at 11 30 16


thyhjwb6 (approved):
👍 

I would agree that the `DatePicker` also needs updating.

PR #3200, opened 2022-03-25
Update to axe v4 and resolve new errors

### LINKED ISSUE #2575: `axe-core` (deeply nested dep) has had a major bump somewhere upstream - we now have 18 new a11y errors that need attention
- Short term fixed tag (`3.x.x`) has been added to root `package.json` resolutions (102b48418986772c843df7c3164bafb4d9d5bea1).

- Remove resolution and fix resulting newly flagged a11y errors from major bump (`4.x.x`).

jpveooys:
## Related issue

Resolve #2575

## Overview

This unpins axe and updates it to the latest version, and resolves errors that have been highlighted by the new version.

**Note: Currently blocked by https://github.com/dequelabs/axe-core-npm/issues/484**

## Link to preview

https://5e25c277526d380020b5e418-agrbksufqr.chromatic.com/

## Reason

Ensure the current version of axe is being used and accessibility errors are resolved.

## Work carried out

- [ ] Resolve errors for `Checkbox` and `Radio`
- [x] Resolve errors for `FloatingBox`
- [x] Resolve errors for `List`
- [x] Resolve errors for `TabSet`
- [x] Resolve errors for `TabNav`
- [x] Resolve errors for `Modal`
- [x] Update `axe-core`, `@axe-core/puppeteer` and `puppeteer`

## Developer notes

Comments about specific changes have been left inline.

Useful references:

- https://www.w3.org/TR/wai-aria-1.2/
- https://www.w3.org/TR/html-aria/
- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles

### DIFF
--- a/packages/react-component-library/src/components/CheckboxRadioBase/CheckboxRadioBase.tsx
+++ b/packages/react-component-library/src/components/CheckboxRadioBase/CheckboxRadioBase.tsx
@@ -96,8 +96,6 @@ export const CheckboxRadioBase = React.forwardRef<
       <StyledWrapper>
         <Root
           className={className}
-          role={type}
-          aria-checked={isChecked}
           $isDisabled={isDisabled}
           $hasContainer={hasContainer}
           $isInvalid={isInvalid}
--- a/packages/react-component-library/src/components/List/List.tsx
+++ b/packages/react-component-library/src/components/List/List.tsx
@@ -34,7 +34,7 @@ export const List: React.FC<ListProps> = ({ children, ...rest }) => {
   )
 
   return (
-    <StyledList role="group" data-testid="list" {...rest}>
+    <StyledList data-testid="list" {...rest}>
       {mapped}
     </StyledList>
   )
--- a/packages/react-component-library/src/components/Modal/Modal.tsx
+++ b/packages/react-component-library/src/components/Modal/Modal.tsx
@@ -12,6 +12,10 @@ import { useOpenClose } from '../../hooks/useOpenClose'
 
 export interface ModalProps extends ComponentWithClass {
   /**
+   * Optional override for the accessible name of the Modal.
+   */
+  'aria-label'?: string
+  /**
    * Optional arbitrary JSX content to display in the main body of the component.
    */
   children?: React.ReactNode
@@ -42,6 +46,9 @@ export interface ModalProps extends ComponentWithClass {
   tertiaryButton?: ButtonProps
   /**
    * Optional text title to display at the top of the component.
+   *
+   * If no title is set, you should set aria-label instead to ensure
+   * the Modal is accessible.
    */
   title?: string
   /**
@@ -52,6 +59,7 @@ export interface ModalProps extends ComponentWithClass {
 }
 
 export const Modal: React.FC<ModalProps> = ({
+  'aria-label': ariaLabel,
   children,
   className,
   descriptionId: externalDescriptionId,
@@ -82,7 +90,10 @@ export const Modal: React.FC<ModalProps> = ({
       className={className}
       role="dialog"
       aria-modal
-      aria-labelledby={title || externalTitleId ? titleId : undefined}
+      aria-label={ariaLabel}
+      aria-labelledby={
+        (title || externalTitleId) && !ariaLabel ? titleId : undefined
+      }
       aria-describedby={descriptionId}
       data-testid="modal-wrapper"
       {...rest}
--- a/packages/react-component-library/src/components/TabNav/TabNavItem.tsx
+++ b/packages/react-component-library/src/components/TabNav/TabNavItem.tsx
@@ -2,12 +2,16 @@ import React from 'react'
 
 import { NavItem } from '../../common/Nav'
 import { StyledTabItem } from '../TabBase/partials/StyledTabItem'
-import { StyledTab } from '../TabBase/partials/StyledTab'
+import { StyledTabNavTab } from './partials/StyledTabNavTab'
 
 export const TabNavItem: React.FC<NavItem> = ({ isActive, link }) => (
   <StyledTabItem data-testid="tab-nav-tab">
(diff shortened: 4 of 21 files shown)


packages/react-component-library/src/components/CheckboxRadioBase/CheckboxRadioBase.tsx
```diff
@@ -96,8 +96,6 @@ export const CheckboxRadioBase = React.forwardRef<
       <StyledWrapper>
         <Root
           className={className}
-          role={type}
```
jpveooys, line 99: The error here was:

https://dequeuniversity.com/rules/axe/4.4/nested-interactive

This was occurring as we effectively had a checkbox within a checkbox (as we also have a visually-hidden `<input>`).

The simply removes the roles from this element.

A further improvement would be to make the outer element the `<label>` and turn the child elements that are `<div>`s into `<span>`s. However, this is a larger refactor so I've left it out.

packages/react-component-library/src/components/TabSet/TabItem.tsx
```diff
@@ -37,17 +37,18 @@ export const TabItem = forwardRef<HTMLLIElement, TabItemProps>(
...
-        tabIndex={!isActive ? -1 : 0}
+        role="presentation"
         data-testid="tab-set-tab"
-        aria-label={children.toString()}
-        onKeyDown={onKeyDown}
       >
         <StyledTab
           $isActive={isActive}
           $isScrollable={isScrollable}
+          id={`tab-button-${tabId}`}
```
jpveooys, line 46: The error here was:

https://dequeuniversity.com/rules/axe/4.4/nested-interactive

This was due to the `tab` role containing a `button` role. I simply rejigged things so that the `<button>` is now the `tab` (following the same pattern as the GOV.UK Design System).

Also removed `aria-label` as a button gets an accessible name automatically, and `children.toString()` was `[object Object]` in some cases.

packages/react-component-library/src/components/TabSet/TabContent.tsx
```diff
@@ -18,9 +18,9 @@ export const TabContent: React.FC<TabContentProps> = ({
   <StyledContent
     $isActive={isActive}
     role="tabpanel"
-    aria-labelledby={tabId}
+    aria-labelledby={`tab-button-${tabId}`}
```
jpveooys, line 21: I noticed the `aria-labelledby` here appeared to be incorrect as it was the same as the `id` for this element. 

I've changed it to be the `id` of the `tab` (matching the GOV.UK Design System).

packages/react-component-library/src/components/List/List.tsx
```diff
@@ -34,7 +34,7 @@ export const List: React.FC<ListProps> = ({ children, ...rest }) => {
   )
 
   return (
-    <StyledList role="group" data-testid="list" {...rest}>
```
jpveooys, line 37: The error for this component was:

https://dequeuniversity.com/rules/axe/4.4/aria-required-parent

This was due to `listtem` roles not being within a `list` role. (On further investigation, the `group` role can be used to group multiple `listitem`s within a `list`, but isn't a replacement for a `list`.)

I've removed the roles completely as `ol` and `li` are already being used (and these have the correct roles automatically).

packages/react-component-library/src/primitives/FloatingBox/FloatingBox.stories.tsx
```diff
@@ -13,29 +13,24 @@ export default {
...
     <div style={{ padding: '0 1rem' }}>
       <pre>Arbitrary JSX content</pre>
     </div>
   </FloatingBox>
 )
 
+export const Default = Template.bind({})
 Default.args = {
   scheme: FLOATING_BOX_SCHEME.LIGHT,
+  'aria-label': 'A floating box',
```
jpveooys, line 27: This error here was:

https://dequeuniversity.com/rules/axe/4.4/aria-dialog-name

This simply sets an accessible name for these stories (which can be done in multiple ways).

packages/react-component-library/src/components/TabNav/TabNavItem.tsx
```diff
@@ -2,12 +2,16 @@ import React from 'react'
...
-import { StyledTab } from '../TabBase/partials/StyledTab'
+import { StyledTabNavTab } from './partials/StyledTabNavTab'
 
 export const TabNavItem: React.FC<NavItem> = ({ isActive, link }) => (
   <StyledTabItem data-testid="tab-nav-tab">
-    <StyledTab $isActive={isActive} data-testid="tab-nav-tab-button">
-      <div>{link}</div>
-    </StyledTab>
+    <StyledTabNavTab
+      as="div"
```
jpveooys, line 10: The error here was:

https://dequeuniversity.com/rules/axe/4.4/nested-interactive

This was due to the `button` role containing a `link` role. This resolves the problem by changing the `<button>` element to be a `<div>`.

packages/react-component-library/src/components/Modal/Modal.tsx
```diff
@@ -11,6 +11,10 @@ import { useExternalId } from '../../hooks/useExternalId'
 import { useOpenClose } from '../../hooks/useOpenClose'
 
 export interface ModalProps extends ComponentWithClass {
+  /**
+   * Optional override for the accessible name of the Modal.
+   */
+  'aria-label'?: string
```
jpveooys, line 17: The error for Modal was:

https://dequeuniversity.com/rules/axe/4.4/aria-dialog-name

This was specific to Modals without a title and due to such Modals not having an accessible name.

This was tricky to resolve without any breaking changes, and also taking into account how Dialog is using Modal.

Personally I'm not sure why title is optional – but this simply makes it possible to explicitly set `aria-label` for a Modal to resolve this without any breaking changes.

jpveooys:
Closing this as I'm going to split it up into multiple PRs.

PR #3216, opened 2022-03-30
feat(CheckboxRadioBase): Convert to a controlled component

### LINKED ISSUE #3214: Allow Checkbox component to be a controlled input
Would this be possible if  the checkbox component was passed a checked prop  that updated the value when the props changed instead of a defaultProp.

We have a feature in our app that allows user to remove items from a different component. 
Demo use case
https://codesandbox.io/s/input-controlled-jznj9j

m7kvqbe1:
## Related issue

Closes #3214

## Overview

Support both uncontrolled and controlled usage and fix usage examples.

## Reason

>Would this be possible if the checkbox component was passed a checked prop that updated the value when the props changed instead of a defaultProp.

## Work carried out

- [x] Supplement with additional tests
- [x] Support both uncontrolled and controlled usage
- [x] Fix broken react-hook-form example usage
- [x] Fix broken native example usage

### DIFF
--- a/packages/react-component-library/src/components/CheckboxRadioBase/CheckboxRadioBase.tsx
+++ b/packages/react-component-library/src/components/CheckboxRadioBase/CheckboxRadioBase.tsx
@@ -18,6 +18,7 @@ export const CheckboxRadioBase = React.forwardRef<
     {
       className = '',
       id: externalId,
+      checked = false,
       defaultChecked,
       description,
       isDisabled = false,
@@ -36,7 +37,16 @@ export const CheckboxRadioBase = React.forwardRef<
   ) => {
     const id = useExternalId(`${type}-input`, externalId)
     const localRef = useRef<HTMLInputElement>(null)
-    const [isChecked, setIsChecked] = useState(defaultChecked)
+    const isControlled = defaultChecked === undefined
+    const hasContainer = variant !== CHECKBOX_RADIO_VARIANT.NO_CONTAINER
+    const [isChecked, setIsChecked] = useState<boolean>(
+      defaultChecked ?? checked
+    )
+
+    // Respect external changes to `defaultChecked` and `checked`
+    useEffect(() => {
+      setIsChecked(defaultChecked ?? checked)
+    }, [defaultChecked, checked])
 
     // Handle implicit deselection of radio in collection
     useEffect(() => {
@@ -75,7 +85,7 @@ export const CheckboxRadioBase = React.forwardRef<
     }
 
     const handleOnChange = (e: React.FormEvent<HTMLInputElement>) => {
-      setIsChecked(!isChecked)
+      setIsChecked(e.currentTarget.checked)
 
       if (type === 'radio') {
         const event = new CustomEvent('_mod-uk-ds:radio:selected', {
@@ -89,8 +99,6 @@ export const CheckboxRadioBase = React.forwardRef<
         onChange(e)
       }
     }
-
-    const hasContainer = variant !== CHECKBOX_RADIO_VARIANT.NO_CONTAINER
 
     return (
       <StyledWrapper>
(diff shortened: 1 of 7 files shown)


jpveooys (approved):
Was there anything broken when `Checkbox` supported both `checked` and `defaultChecked`? I think it would be good to allow both possibilities like a native `<input>` if possible (but not critical).

Also seems like it was an existing problem but also noticed that the [prepoluated native form story](https://5e25c277526d380020b5e418-omlfrwpjmj.chromatic.com/?path=/story/forms-usage-native--prepopulated) isn't prepopulating the checkboxes and radios correctly – not sure if that's an easy fix to add in here.

m7kvqbe1:
There is a [seperate ticket](https://github.com/defencedigital/mod-uk-design-system/issues/3219) for the `TextInput` and `TextArea` pre-populated form example issues.

packages/react-component-library/src/components/CheckboxRadioBase/CheckboxRadioBase.tsx
```diff
@@ -124,6 +131,7 @@ export const CheckboxRadioBase = React.forwardRef<
                 onBlur={onBlur}
                 disabled={isDisabled}
                 data-testid={`${type}-input`}
+                checked={defaultChecked ? undefined : isChecked}
```
m7kvqbe1: `defaultChecked` and `checked` are mutually exclusive.

packages/react-component-library/src/components/CheckboxRadioBase/CheckboxRadioBase.tsx
```diff
@@ -36,7 +37,13 @@ export const CheckboxRadioBase = React.forwardRef<
   ) => {
     const id = useExternalId(`${type}-input`, externalId)
     const localRef = useRef<HTMLInputElement>(null)
-    const [isChecked, setIsChecked] = useState(defaultChecked)
+    const [isChecked, setIsChecked] = useState<boolean>(
+      defaultChecked || checked
```
jpveooys: `??` might make more sense here (i.e. still use `defaultChecked` if it's false rather than undefined):

```suggestion
      defaultChecked ?? checked
```

packages/react-component-library/src/components/CheckboxRadioBase/CheckboxRadioBase.tsx
```diff
@@ -36,7 +37,13 @@ export const CheckboxRadioBase = React.forwardRef<
...
+    )
+
+    useEffect(() => {
+      setIsChecked(checked)
+    }, [checked])
```
jpveooys: If `defaultChecked` is used then the background ends up in the wrong state which seems to be due to the useEffect running when the component is mounted:

Screencast_04-04-22_11:50:39

So I think this useEffect would need a guard in the body to not do anything in that case.

Alternatively, another approach may be to rejig things to try to do this (should get rid of the need for a useEffect I think):

- if the component is controlled, just use `checked` (and don't worry about maintaining `isChecked`)
- if the component is uncontrolled, then try to passively monitor the state and maintain `isChecked`
- (make the stories use `defaultChecked` again)

packages/react-component-library/src/components/CheckboxRadioBase/CheckboxRadioBase.tsx
jpveooys: Would be good to explicitly test against `undefined` I think (as earlier comment – I think `false` should be respected):

```suggestion
                checked={defaultChecked !== undefined ? undefined : isChecked}
```

(It might even better to define e.g. an `const isControlled = ...` above and use that in both places.)

jpveooys (commented):
Nice work! Think there's a problem with `defaultChecked` though (see middle comment).

jpveooys (approved):

PR #3230, opened 2022-04-11
Add toggle button focus to `Autocomplete`

### LINKED ISSUE #3015: Fix `Autocomplete` enter key and focus
> if you select a value using the Enter key, nothing is focused afterwards and it's a bit tricky to change your answer using the keyboard

thyhjwb6:
## Related issue
Closes #3015 

## Overview
Focuses the toggle button when an item is selected.

## Link to preview
https://autocomplete-demo.netlify.app/?path=/story/autocomplete--default

## Reason
User needs to be able to change the option.

## Work carried out
- [x] Tidy code
- [x] Add focus of toggle button

## Developer notes
Tests have been added using Cypress as it's difficult to get the coverage with Jest.


packages/react-component-library/src/components/Select/Select.tsx
```diff
@@ -8,6 +8,7 @@ import {
   SelectChildWithStringType,
   SelectLayout,
 } from '../SelectBase'
+import { useMenuVisibility } from '../SelectBase/hooks/useMenuVisibility'
 import { useExternalId } from '../../hooks/useExternalId'
```
jpveooys: Are the imports in the right order here?

jpveooys (approved):
Looks good :+1: 

I noticed that Select also seems to have a problem with Shift-Tab – might be good to fix that as well if it's easy.

FYI the Chromatic UI Tests should run after this is rebased.

jpveooys:
@thyhjwb6 There are some build failures – looks like `useSelectMenu.ts` is in the wrong location.

packages/react-component-library/src/components/Select/hooks/useSelectMenu.ts
```diff
@@ -0,0 +1,22 @@
...
+  inputRef: React.RefObject<HTMLInputElement>,
+  closeMenu: () => void
+): {
+  onMenuKeyDownHandler: (e: React.KeyboardEvent<HTMLUListElement>) => void
+} {
+  const onMenuKeyDownHandler = useCallback(
+    (e: React.KeyboardEvent<HTMLUListElement>) => {
+      if (e.key === 'Tab') {
+        inputRef.current?.focus()
+        closeMenu()
```
jpveooys: Small thing but I was wondering if the closeMenu() is necessary – was there a problem without it?

jpveooys (approved):

PR #3245, opened 2022-04-19
fix(Forms): Stop long input labels from wrapping

### LINKED ISSUE #3217: Date Picker overflow is not handled when the component is small
Hi we are using this component in a drawer component where the space is compact
image

jpveooys:
## Related issue

Resolves #3217

## Overview

This stops long labels wrapping in `TextInput`, `TextArea` and various other components.

Instead, an ellipsis is shown when a label doesn't fit in the available space.

## Link to preview

https://5e25c277526d380020b5e418-yzdzruclmq.chromatic.com/

## Reason

To avoid visual glitches.

## Work carried out

- [x] Rework label styles
- [x] Add stories
- [x] Update Babel config

#### Before

Screencast_19-04-22_14:27:52

#### After

Screencast_19-04-22_14:37:47

#### Before

Screencast_19-04-22_14:29:13

#### After

Screencast_19-04-22_14:38:41

## Developer notes

The styles have been updated to use a transition on `font-size` and `padding` instead of `transform`, as it wasn't possible to fix this while still using a `transform`.

Full list of affected components:

- Autocomplete
- DatePicker
- Select
- NumberInput
- TextArea
- TextInput

### DIFF
--- a/packages/react-component-library/babel.config.js
+++ b/packages/react-component-library/babel.config.js
@@ -45,7 +45,13 @@ module.exports = {
       ],
     },
     test: {
-      plugins: ['@babel/plugin-transform-modules-commonjs'],
+      plugins: [
+        'babel-plugin-styled-components',
+        '@babel/plugin-transform-modules-commonjs',
+      ],
+    },
+    'storybook-test': {
+      plugins: ['babel-plugin-styled-components'],
     },
   },
 }
--- a/packages/react-component-library/cypress/specs/Autocomplete/index.spec.ts
+++ b/packages/react-component-library/cypress/specs/Autocomplete/index.spec.ts
@@ -4,7 +4,9 @@ import {
   ColorDanger800,
   ColorNeutral100,
   ColorNeutral200,
+  TypographyS,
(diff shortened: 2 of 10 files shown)


packages/react-component-library/babel.config.js
```diff
@@ -45,7 +45,13 @@ module.exports = {
...
     },
     test: {
-      plugins: ['@babel/plugin-transform-modules-commonjs'],
+      plugins: [
+        'babel-plugin-styled-components',
+        '@babel/plugin-transform-modules-commonjs',
+      ],
+    },
+    'storybook-test': {
+      plugins: ['babel-plugin-styled-components'],
```
jpveooys, line 54: Added a new BABEL_ENV with `babel-plugin-styled-components` to get the css prop working for `yarn storybook:test`. Also no longer seems to cause a performance regression having it in `test` so added there too for consistency.

m7kvqbe1: Is it worth breaking this change out into a seperate commit?

m7kvqbe1 (approved):
LGTM

PR #3258, opened 2022-04-27
fix(FloatingBox): Resolve StrictMode warning

### LINKED ISSUE #3250: findDomNode is deprecated warning in React Strict Mode
Transition from 'react-transition-group' using findDomNode could we use a nodeRef to stop this warning.

https://github.com/reactjs/react-transition-group/issues/779

I believe adding `nodeRef={floatingElementRef}`  to `<Transition>` should solve this.

https://github.com/defencedigital/mod-uk-design-system/blob/25db9e94ad6257e11287ec329f3df45f7411ad88/packages/react-component-library/src/primitives/FloatingBox/FloatingBox.tsx#L92

m7kvqbe1:
## Related issue

Closes #3250 

## Overview

Resolve `React.StrictMode` warning about deprecated `findDOMNode` usage.

## Reason

>Transition from 'react-transition-group' using findDomNode could we use a nodeRef to stop this warning.

## Work carried out

- [x] Enable `React.StrictMode` for Storybook development builds
- [x] Use a ref to avoid the `findDOMNode` usage by `Transition`

## Developer notes

Note that `React.StrictMode` is disabled in the deployed production build of Storybook.

### DIFF
--- a/packages/react-component-library/.storybook/preview.js
+++ b/packages/react-component-library/.storybook/preview.js
@@ -1,3 +1,4 @@
+import React from 'react'
 import '@defencedigital/fonts'
 import 'iframe-resizer/js/iframeResizer.contentWindow'
 import { ResizeObserver } from '@juggle/resize-observer'
@@ -58,8 +59,10 @@ export const parameters = {
 
 export const decorators = [
   (Story) => (
-    <GlobalStyleProvider>
-      <Story />
-    </GlobalStyleProvider>
+    <React.StrictMode>
+      <GlobalStyleProvider>
+        <Story />
+      </GlobalStyleProvider>
+    </React.StrictMode>
   ),
 ]
--- a/packages/react-component-library/.storybook/preview.js
+++ b/packages/react-component-library/.storybook/preview.js
@@ -58,11 +58,6 @@ export const parameters = {
 }
 
 export const decorators = [
-  (Story) => (
-    <React.StrictMode>
-      <GlobalStyleProvider>
-        <Story />
-      </GlobalStyleProvider>
-    </React.StrictMode>
-  ),
+  // https://github.com/storybookjs/storybook/issues/15223#issuecomment-1092837912
+  (Story) => <GlobalStyleProvider>{Story()}</GlobalStyleProvider>,
 ]
--- a/packages/react-component-library/src/primitives/FloatingBox/FloatingBox.tsx
+++ b/packages/react-component-library/src/primitives/FloatingBox/FloatingBox.tsx
@@ -1,6 +1,7 @@
-import React from 'react'
+import React, { useRef } from 'react'
 import { Placement } from '@popperjs/core'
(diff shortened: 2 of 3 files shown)


packages/react-component-library/src/primitives/FloatingBox/FloatingBox.tsx
```diff
@@ -89,11 +91,11 @@ export const FloatingBox: React.FC<FloatingBoxProps> = ({
           {renderTarget}
         </StyledTarget>
       )}
-      <Transition in={isVisible} timeout={0} unmountOnExit>
+      <Transition nodeRef={nodeRef} in={isVisible} timeout={0} unmountOnExit>
         {(state) => (
           <StyledFloatingBox
             style={{ ...styles.popper, ...TRANSITION_STYLES[state] }}
-            ref={floatingElementRef}
+            ref={mergeRefs([nodeRef, floatingElementRef])}
```
m7kvqbe1, line 98: Local `nodeRef` merged because `Transition` doesn't like the special stateful refs `useFloatingBox` utilises.

packages/react-component-library/.storybook/preview.js
```diff
@@ -58,8 +59,10 @@ export const parameters = {
 
 export const decorators = [
   (Story) => (
-    <GlobalStyleProvider>
-      <Story />
-    </GlobalStyleProvider>
+    <React.StrictMode>
```
jpveooys: Does the strictMode option mentioned here work? https://storybook.js.org/blog/storybook-6-1/

(BTW I noticed that these decorators runs multiple times on the Docs tab, which is causing some duplicate styles from `GlobalStyleProvider`.)

m7kvqbe1: Hmm i'll investigate I was under the impression each story was nested within an iframe - I guess not within docs.

jpveooys: Doing it the using the option I get a lot of "Rendered more hooks than during the previous render." messages which is due to https://github.com/storybookjs/storybook/issues/15223 it seems...

So it may be this way is better for now.

packages/react-component-library/.storybook/preview.js
```diff
@@ -3,6 +3,7 @@ import 'iframe-resizer/js/iframeResizer.contentWindow'
 import { ResizeObserver } from '@juggle/resize-observer'
 
 import { GlobalStyleProvider } from '../src/styled-components/GlobalStyle'
+import React from 'react'
```
jpveooys: Would be good to move the React import to the first group.

jpveooys (approved):

packages/react-component-library/.storybook/preview.js
m7kvqbe1: Created ticket to look further into enforcing singleton behaviour of provider: https://github.com/defencedigital/mod-uk-design-system/issues/3259

PR #3269, opened 2022-05-06
feat(Storybook): Add performance addon

### LINKED ISSUE #2601: Get the Storybook Performance Addon working with the react-component-library Storybook implementation
Created from #2495

The [add on](https://github.com/atlassian-labs/storybook-addon-performance) will give us useful data.

Suspected issue is Webpack related.

m7kvqbe1:
## Related issue

Closes #2601

## Overview

Add Storybook add on that profiles components.

https://github.com/atlassian-labs/storybook-addon-performance

## Work carried out

- [x] Register performance add on and apply decorator to stories
- [x] Add `devDependency` for`@testing-library/dom` to `react-component-library` package
- [x] Skip `ContextMenu` Storybook docs tab Cypress test

## Screenshot

Screenshot 2022-05-06 at 10 21 31


jpveooys:
Looks like the resolutions have been undone – `yarn install` from the root should fix it. Not sure if that's related to the test failures though :thinking:

m7kvqbe1:
Added explicit `devDependency` to `react-component-library` package for:

```
"@testing-library/dom": "^8.0.0"
```

The Storybook addon was installing a previous major that caused tests related to external state updates to fail.

m7kvqbe1:
Bumped `@storybook/*` dependencies in order to invalidate the GHA `node_modules/` cache.

There is currently no way to manually invalidate the cache: https://github.com/actions/cache/issues/2

m7kvqbe1:
Investigating issues with IE11 (package not transpiled upstream).

m7kvqbe1:
Looks like i've already been here 2 years ago...

https://github.com/atlassian-labs/storybook-addon-performance/issues/53

packages/react-component-library/.storybook/main.js
```diff
@@ -11,6 +11,7 @@ module.exports = {
     '@storybook/addon-a11y',
     '@storybook/addon-controls',
     '@storybook/addon-actions',
+    'storybook-addon-performance/dist/cjs/register',
```
m7kvqbe1, line 14: Using transpiled CJS build for IE11: https://github.com/atlassian-labs/storybook-addon-performance/issues/53

packages/react-component-library/cypress/specs/ContextMenu/index.spec.ts
```diff
@@ -8,9 +8,11 @@ describe('ContextMenu', () => {
   describe('Storybook: Docs Mode', () => {
```
jpveooys: It looks like this was an explicit test for the Docs tab (I guess we had a problem with the context menu being obscured there) – would be good to make the name consistent with the test in any case.

jpveooys (approved):

packages/react-component-library/cypress/specs/ContextMenu/index.spec.ts
m7kvqbe1: Hmm yeah i'm not sure how that being docs tab specific would have been any different. I'll update the naming.

m7kvqbe1: I've created an issue upstream re Storybook `ArgsTable` Emotion warnings: https://github.com/storybookjs/storybook/issues/18232

In the meantime i'm going to temporarily skip that Cypress suite. It is after all testing Storybook itself and not the component implementation.

PR #3271, opened 2022-05-09
Fix various incorrect CSS rules

### LINKED ISSUE #3229: Masthead search bar obscured by navigation tabs
Screenshot 2022-04-08 at 14 59 54

jpveooys:
## Related issue

Resolves #3229

## Overview

This fixes various incorrect CSS, including a z-order problem with the Masthead search box.

## Link to preview

https://5e25c277526d380020b5e418-vdhbcknonq.chromatic.com/

## Reason

To make sure all CSS is correct.

## Work carried out

- [x] Fix CSS errors identified with linting tool

#### Before

image

#### After

image

## Developer notes

I got the CSS-in-JS linting working here: https://github.com/defencedigital/mod-uk-design-system/commit/f9c863133bacad90892d2cf84e637a41a22ce202

However, [it depends on a package that is being deprecated](https://github.com/stylelint/postcss-css-in-js/issues/225), and it isn't the smoothest of experiences either. So I've left it out; things should be better further down the line when a new PostCSS syntax for Styled Components has been written.

### DIFF
--- a/packages/react-component-library/src/components/Dropdown/Dropdown.tsx
+++ b/packages/react-component-library/src/components/Dropdown/Dropdown.tsx
@@ -9,7 +9,7 @@ import { DropdownLabel } from './DropdownLabel'
 import { DropdownOption } from './DropdownOption'
 import { DropdownPlaceholder } from './DropdownPlaceholder'
 
-const { color, spacing } = selectors
+const { color, shadow, spacing } = selectors
 
 export interface DropdownProps {
   /**
@@ -71,7 +71,7 @@ const StyledSelect = styled(Select)`
   }
 
   .rn-dropdown__menu {
-    @include m.shadow('1');
+    box-shadow: 0 4px 16px rgba(0, 0, 0, 0.1), 0 1px 2px rgba(0, 0, 0, 0.1);
     top: 102%;
     margin: ${spacing('6')} 0 0 5%;
     width: 95%;
--- a/packages/react-component-library/src/components/TextInput/partials/StyledInput.tsx
+++ b/packages/react-component-library/src/components/TextInput/partials/StyledInput.tsx
@@ -21,7 +21,7 @@ function removeAutoFillBackground() {
     &:-webkit-autofill:hover,
     &:-webkit-autofill:focus,
     &:-webkit-autofill:active {
-      -webkit-transition: background-color 9999s ease-out;
+      transition: background-color 9999s ease-out;
     }
 
     filter: none; // Firefox
--- a/packages/react-component-library/src/components/Toast/partials/StyledToastLabel.tsx
+++ b/packages/react-component-library/src/components/Toast/partials/StyledToastLabel.tsx
@@ -4,12 +4,10 @@ import { selectors } from '@defencedigital/design-tokens'
 const { color, fontSize, spacing } = selectors
 
 export const StyledToastLabel = styled.span`
-  display: inline-flex;
-  align-items: center;
+  display: block;
   color: ${color('neutral', '500')};
   font-size: ${fontSize('m')};
   font-weight: 600;
-  display: block;
   padding: 0;
 
   > svg {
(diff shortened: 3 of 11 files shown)


packages/react-component-library/src/components/Dropdown/Dropdown.tsx
```diff
@@ -71,7 +71,7 @@ const StyledSelect = styled(Select)`
   }
 
   .rn-dropdown__menu {
-    @include m.shadow('1');
+    box-shadow: ${shadow('1')};
```
jpveooys, line 74: This one is a bit debatable; `shadow('1')` is obviously what was intended at some point, but it's barely noticeable and the default react-select shadow has been there for a while.

packages/react-component-library/src/components/TextInput/partials/StyledInput.tsx
```diff
@@ -21,7 +21,7 @@ function removeAutoFillBackground() {
     &:-webkit-autofill:hover,
     &:-webkit-autofill:focus,
     &:-webkit-autofill:active {
-      -webkit-transition: background-color 9999s ease-out;
+      transition: background-color 9999s ease-out;
```
jpveooys, line 24: `transition` is autoprefixed so the vendor prefix isn't needed here. (`autofill` doesn't get autoprefixed yet, unfortunately.)

packages/react-component-library/src/components/Toast/partials/StyledToastLabel.tsx
```diff
@@ -4,12 +4,10 @@ import { selectors } from '@defencedigital/design-tokens'
 const { color, fontSize, spacing } = selectors
 
 export const StyledToastLabel = styled.span`
-  display: inline-flex;
```
jpveooys, line 7: `display: inline-flex;` was being overridden with `display: block;`, so I've removed the flexbox rules (the alternative, removing `display: block;`, broke the positioning of the icon).

packages/react-component-library/src/components/TopLevelNavigation/SearchBar/partials/StyledSearchBar.tsx
```diff
@@ -1,21 +1,23 @@
...
+  ${StyledInput} {
     font-size: ${fontSize('l')};
   }
 
   ${mq({ gte: 'xs' })`
-    .rn-textinput__input {
+    ${StyledInput} {
       font-size: ${fontSize('base')};
     }
   `}
```
jpveooys: I've fixed the font-size rules, but I find this media query a bit weird if I'm being honest (especially as it's the only media query in the Masthead).

packages/react-component-library/src/components/Dropdown/Dropdown.tsx
```diff
@@ -71,7 +71,7 @@ const StyledSelect = styled(Select)`
   }
 
   .rn-dropdown__menu {
-    @include m.shadow('1');
+    box-shadow: 0 4px 16px rgba(0, 0, 0, 0.1), 0 1px 2px rgba(0, 0, 0, 0.1);
```
jpveooys, line 74: Currently, the default shadow from react-select is being used. 

The broken value, `shadow('1')`, is almost invisible and doesn't match what's in Axure and the previous Sketch library – I've copied the definition from Sketch instead.

packages/react-component-library/src/components/TopLevelNavigation/SearchBar/partials/StyledSearchBar.tsx
m7kvqbe1: I think we should remove it.

m7kvqbe1 (approved):
LGTM

packages/react-component-library/src/components/TopLevelNavigation/SearchBar/partials/StyledSearchBar.tsx
jpveooys: Yep, cool. I'll just delete these then and leave it at the default 16px (which is how it's been for ages).

PR #3292, opened 2022-05-24
refactor(TabSet): Remove ResizeObserver usage

### LINKED ISSUE #3291: Scrollable TabSet e2e tests intermittently failing in Firefox
The scrollable TabSet e2e tests have suddenly started failing in Firefox. The behaviour appears to be inconsistent.

jpveooys:
## Related issue

Resolves #3291

## Overview

This removes ResizeObserver usage in scrollable `TabSet`s, and instead checks for changes to the element size lazily.

## Link to preview

https://5e25c277526d380020b5e418-dokkqurffe.chromatic.com/?path=/story/tab-set--scrollable

## Reason

The scrollable TabSet e2e tests became unreliable in Firefox 100 due to the ResizeObserver callback being called with a longer delay.

Checking for resizes to the element lazily avoids that class of problem, as the operations always happen in the right order.

## Work carried out

- [x] Replace ResizeObserver usage
- [x] Remove `use-resize-observer` and `@juggle/resize-observer` dependencies

## Developer notes

The resize check is mainly relevant when the tabs are scrolled all the way right and the viewport is resized.

### DIFF
--- a/packages/cra-template-defencedigital/template.json
+++ b/packages/cra-template-defencedigital/template.json
@@ -16,7 +16,6 @@
       "@graphql-codegen/typescript": "^2.4.1",
       "@graphql-codegen/typescript-operations": "^2.2.1",
       "@graphql-codegen/typescript-react-apollo": "^3.2.2",
-      "@juggle/resize-observer": "^3.3.1",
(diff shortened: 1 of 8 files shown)


thyhjwb6 (approved):
👍

PR #3293, opened 2022-05-26
fix(Autocomplete): Allow cursor position

### LINKED ISSUE #3010: Click on `AutocompleteE` cursor position
> Clicking on a Select that already has some user input always places the cursor at the end of the user input, rather than at the point where the user clicked on (which would be consistent with how our other inputs work - eg. DatePicker or Text Input). If easy to change I would, but if not this is ok too.

thyhjwb6:
## Related issue
Closes #3010 

## Overview
Allows the cursor to be positioned where the input is clicked.

## Reason
The cursor was being positioned at the end which is not good from a user perspective as you would expect the cursor to be positioned where the input is clicked as with other text inputs.

## Work carried out
- [x] Position cursor where input is clicked

## Developer notes
A test has been added with `skip` because of an [open issue](https://github.com/cypress-io/cypress/issues/5721) with Cypress.

### DIFF
--- a/packages/react-component-library/cypress/specs/Select/index.spec.ts
+++ b/packages/react-component-library/cypress/specs/Select/index.spec.ts
@@ -17,10 +17,9 @@ describe('Select', () => {
       cy.get(selectors.select.option).should('have.length', 4)
     })
 
-    describe('and `th` is typed', () => {
+    describe('and `t` is typed', () => {
       before(() => {
-        cy.get(selectors.select.outerWrapper).type('Th{enter}')
-        cy.get('body').click()
+        cy.get(selectors.select.listBox).type('t{enter}')
       })
 
       it('sets the value as the item', () => {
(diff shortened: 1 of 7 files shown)


packages/react-component-library/cypress/specs/Select/index.spec.ts
```diff
@@ -17,9 +17,9 @@ describe('Select', () => {
       cy.get(selectors.select.option).should('have.length', 4)
     })
 
-    describe('and `th` is typed', () => {
+    describe('and `t` is typed', () => {
       before(() => {
-        cy.get(selectors.select.outerWrapper).type('Th{enter}')
+        cy.get(selectors.select.listBox).type('t{enter}')
         cy.get('body').click()
```
jpveooys: Maybe this click can go as well (since it's pressing Enter).

jpveooys (approved):
Looks good! Nice one :+1:

m7kvqbe1 (approved):
LGTM

PR #3294, opened 2022-05-26
fix(NumberInput): Display initial value of zero correctly

### LINKED ISSUE #3284: A NumberInput with an initial value of 0 renders a blank input
If `value={0}` is passed to a NumberInput, it renders as blank and doesn't display the value of `0`.

The component appears to not be differentiating between `null` and `0`.

Discussed in https://github.com/defencedigital/mod-uk-design-system/discussions/3277

jpveooys:
## Related issue

Resolves #3284

## Overview

This makes an initial value of `0` display as `'0'`, rather than an empty string.

## Reason

`0` and null should be treated differently; `null` meaning no value, and `0` being the number 0.

## Work carried out

- [x] Correct handling of an initial value of `0`

### DIFF
--- a/packages/react-component-library/src/components/NumberInput/NumberInput.test.tsx
+++ b/packages/react-component-library/src/components/NumberInput/NumberInput.test.tsx
@@ -679,6 +679,17 @@ describe('NumberInput', () => {
     })
   })
 
+  describe.each([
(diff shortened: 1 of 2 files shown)


m7kvqbe1 (approved):
👍

PR #3295, opened 2022-05-27
feat(NumberInput): Add ability to increment decrement with keyboard

### LINKED ISSUE #2976: NumberInputE: Add ability to increment or decrement value using up and down keys
### Description

From the spec:

> Pressing the ↑key will increase the current value by an amount equal to the preset step.
> - If there is no current value, the input should automatically start either at 0 or the valid value nearest to 0.
>
> Pressing the ↓key will decrease the current value by an amount equal to the preset step.
> - If there is no current value, the input should automatically start either at 0 or the valid value nearest to 0.

m7kvqbe1:
## Related issue

Closes #2976

## Overview

Add ability to increment and decrement the committed value via the `ArrowUp` and `ArrowDown` keys.

## Link to preview

https://6290970b73698a03f02ddfb6--vocal-melomakarona-4997f7.netlify.app

## Reason

>Pressing the ↑key will increase the current value by an amount equal to the preset step.
>Pressing the ↓key will decrease the current value by an amount equal to the preset step.

## Work carried out

- [x] Supplement existing automated tests
- [x] Implement keyboard behaviour when input is focussed

### DIFF
--- a/packages/react-component-library/src/components/NumberInput/Input.tsx
+++ b/packages/react-component-library/src/components/NumberInput/Input.tsx
@@ -1,5 +1,6 @@
-import React, { KeyboardEventHandler } from 'react'
+import React, { KeyboardEventHandler, useCallback } from 'react'
 import { isNil } from 'lodash'
+import { Decimal } from 'decimal.js'
 
 import { ComponentSizeType } from '../Forms'
 import { StyledInput } from '../TextInput/partials/StyledInput'
@@ -23,6 +24,7 @@ export interface InputProps {
   onPaste: (event: React.ClipboardEvent<HTMLInputElement>) => void
   placeholder?: string
   size: ComponentSizeType
+  step: number
   value?: string | null
 }
 
@@ -32,6 +34,7 @@ export const Input: React.FC<InputProps> = ({
   id,
   label,
   size,
+  step,
   value,
   onChange,
   ...rest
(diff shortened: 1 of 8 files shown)


packages/react-component-library/src/components/NumberInput/Input.tsx
```diff
@@ -53,6 +88,8 @@ export const Input: React.FC<InputProps> = ({
         disabled={isDisabled}
         id={id}
         value={value || ''}
+        onKeyDown={handleKeyDown}
```
m7kvqbe1, line 65: `onKeyDown` is used to prevent glitchy cursor positioning issue.

packages/react-component-library/src/components/NumberInput/Input.tsx
```diff
@@ -29,10 +33,41 @@ export const Input: React.FC<InputProps> = ({
...
   ...rest
 }) => {
   const hasLabel = !!(label && label.length)
 
+  const KEY_ARROW_UP = 'ArrowUp'
+  const KEY_ARROW_DOWN = 'ArrowDown'
+
+  const handleKeyDown: KeyboardEventHandler<HTMLInputElement> = (event) => {
+    const {
+      code,
```
jpveooys: Would try using `key` here – it's used elsewhere I think and should work in in IE11 etc.

packages/react-component-library/src/components/NumberInput/Input.tsx
```diff
@@ -29,10 +33,41 @@ export const Input: React.FC<InputProps> = ({
...
+        ...event,
+        currentTarget: {
+          ...event.currentTarget,
+          value: String(
+            code === KEY_ARROW_UP
+              ? Number(eventValue) + 1
+              : Number(eventValue) - 1
+          ),
+        },
+      }
```
jpveooys: Hmm I'm not too sure about this as it'll get passed to the parent component (via `onChange`) but at least `currentTarget` won't have the same 'real' type/prototype and seems like methods and most properties doesn't come through:

image

jpveooys (commented):
Seems to work well :+1:

But a couple of comments on the implementation though.

packages/react-component-library/src/components/NumberInput/Input.tsx
```diff
@@ -29,10 +33,41 @@ export const Input: React.FC<InputProps> = ({
...
   value,
+  onChange,
   ...rest
 }) => {
   const hasLabel = !!(label && label.length)
 
+  const KEY_ARROW_UP = 'ArrowUp'
+  const KEY_ARROW_DOWN = 'ArrowDown'
+
+  const handleKeyDown: KeyboardEventHandler<HTMLInputElement> = (event) => {
```
thyhjwb6: An option to consider is whether this could live inside its own hook such as `useInputKeys`.

thyhjwb6 (commented):
Looks good 👍

packages/react-component-library/src/components/NumberInput/Input.tsx
m7kvqbe1: Hmm good spot - i'm now using some lodash utilities to deep clone the event object and mutate the copy.

jpveooys (approved):

thyhjwb6 (approved):
👍

PR #3307, opened 2022-06-01
feat(GlobalStyleProvider): Handle multiple instances gracefully

### LINKED ISSUE #3259: Add GlobalStyleProvider singleton check to prevent use more than once within a consuming app (parallel and nested)
Storybook is a good example of where this is a problem.

Individual stories are decorated with the provider. 

In the docs tab these are rendered inline within a single iframe. 

The result is duplication of global styles.

---

Find a way of ensuring only a single instance of the `GlobalStyleProvider` can exist at any one time within a React tree. 

Possibly also add a warning if it's not the root node.

---

See this ticket for relevant comments: https://github.com/defencedigital/mod-uk-design-system/pull/3258

jpveooys:
## Related issue

Resolves #3259

## Overview

This adds logic to `GlobalStyleProvider` to stop it inserting duplicate styles when it is inserted into a page multiple times.

## Link to preview

https://5e25c277526d380020b5e418-okdsdrhdoc.chromatic.com/

## Reason

Avoid duplicate styles e.g. in Storybook on the Docs tab.

## Work carried out

- [x] Add first instance check to `GlobalStyleProvider`

### Before

image

### After

image

## Developer notes

Note that there are many edge cases in trying to do this, in particular relating to the order different instances of `GlobalStyleProvider` are mounted and unmounted. Hence, I'm not convinced trying to do this is a good idea – though I've managed to come up with something that handles those edge cases.

#3308 is suggested as a simpler fix just for Storybook (where the problem originated).

### DIFF
--- a/packages/react-component-library/src/hooks/useIsFirstInstance.ts
+++ b/packages/react-component-library/src/hooks/useIsFirstInstance.ts
@@ -1,0 +1,37 @@
+import { useEffect, useRef, useState } from 'react'
+import { v4 as uuidv4 } from 'uuid'
+
+const registries: {
+  [name: string]: { instances: Set<string>; callbacks: Set<() => void> }
+} = {}
+
+export function useIsFirstInstance(registryName: string) {
+  registries[registryName] ??= { instances: new Set(), callbacks: new Set() }
+  const registry = registries[registryName]
+
+  const [isFirstInstance, setIsFirstInstance] = useState(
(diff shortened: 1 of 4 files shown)


jpveooys:
Discussed this and we'll go with fixing this in Storybook (https://github.com/defencedigital/mod-uk-design-system/pull/3308) and making sure the documentation is clear instead.

PR #3308, opened 2022-06-01
docs(Storybook): Fix duplicated global styles on Docs tab

### LINKED ISSUE #3259: Add GlobalStyleProvider singleton check to prevent use more than once within a consuming app (parallel and nested)

jpveooys:
## Related issue

Resolves #3259 

## Overview

This resolves a problem where GlobalStyleProvider was being rendered multiple times on the Docs tab.

## Link to preview

https://5e25c277526d380020b5e418-uyvvyyskvv.chromatic.com/

## Reason

This was causing duplicate global styles and was slowing down some pages.

## Work carried out

- [x] Use only one GlobalStyleProvider on Storybook docs tab
- [x] Add tests to make sure global style rules are loaded once in Storybook

### Before

image

### After

image

## Developer notes

I amended the Cypress console error check to ignore the `:first-child` console errors were currently getting from Storybook, as they were being triggered in the new tests.

### DIFF
--- a/packages/react-component-library/.storybook/preview.js
+++ b/packages/react-component-library/.storybook/preview.js
@@ -1,6 +1,8 @@
+import { DocsPage } from '@storybook/addon-docs'
 import React from 'react'
 import '@defencedigital/fonts'
 import 'iframe-resizer/js/iframeResizer.contentWindow'
+import 'url-search-params-polyfill'
 import { withPerformance } from 'storybook-addon-performance/dist/cjs'
 
 import { GlobalStyleProvider } from '../src/styled-components/GlobalStyle'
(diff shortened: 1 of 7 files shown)


m7kvqbe1 (approved):
👍

thyhjwb6 (approved):
👍

PR #3422, opened 2022-09-01
fix(Autocomplete): Handle changes to options list after initial render

### LINKED ISSUE #3420: Autocomplete component does not re-render when it receives new children
### Steps to reproduce
Go to 
[Code Sandbox](https://codesandbox.io/s/eloquent-yalow-dwlps3?file=/src/App.js)

or 
[https://dwlps3.csb.app/](https://dwlps3.csb.app/)

The demo is mocking an async loading of the children. 

Once the loading text has gone click the drop down and nothing appears, click again and you will see the children

I think the culprit is here.
https://github.com/defencedigital/mod-uk-design-system/blob/5de10fbdb5502f82844248bd15211556ce14589b/packages/react-component-library/src/components/Autocomplete/hooks/useAutocomplete.ts#L23

I think the onIsOpenChange is setting it after.

### Minimal reproduction

https://codesandbox.io/s/eloquent-yalow-dwlps3?file=/src/App.js

jpveooys:
## Related issue

Resolves #3420

## Overview

This resolves a problem where Autocomplete options did not update if they changed after the initial render.

## Link to preview

https://red-playground.netlify.app/?path=/docs/autocomplete--default

## Reason

This is expected to work e.g. for loading options asynchronously.

## Work carried out

- [x] Remove options from state to fix the bug

## Developer notes

Changes to the option list after the initial render (e.g. loading options asynchronously) previously didn't work, as a copy of the options was kept in the state. As the options are made up of ReactElements (which also have ReactNodes within their props), it is also difficult to keep track of changes to them (simply adding a `useEffect` caused an infinite loop).

This changes how things work slightly so the filter state is tracked in the state, and not the items themselves, which fixes the problem.

### DIFF
--- a/packages/react-component-library/src/components/Autocomplete/hooks/useAutocomplete.ts
+++ b/packages/react-component-library/src/components/Autocomplete/hooks/useAutocomplete.ts
@@ -20,26 +20,39 @@ export function useAutocomplete(
   onIsOpenChange: (
     changes: UseComboboxStateChange<SelectChildWithStringType>
   ) => void
+  onSelectedItemChange: (
+    changes: UseComboboxStateChange<SelectChildWithStringType>
+  ) => void
 } {
   const [hasError, setHasError] = useState<boolean>(isInvalid)
   const inputRef = useRef<HTMLInputElement>(null)
-  const [items, setItems] = useState<SelectChildWithStringType[]>(children)
+  const [filterValue, setFilterValue] = useState<string>('')
+
+  const items = React.Children.toArray(children).filter((child) => {
+    if (!React.isValidElement<SelectBaseOptionAsStringProps>(child)) {
+      return false
+    }
+
+    if (!filterValue) {
+      return true
+    }
+
+    return (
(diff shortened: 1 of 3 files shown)


packages/react-component-library/src/components/Autocomplete/hooks/useAutocomplete.ts
```diff
@@ -20,26 +20,39 @@ export function useAutocomplete(
...
+  onSelectedItemChange: (
+    changes: UseComboboxStateChange<SelectChildWithStringType>
+  ) => void
 } {
   const [hasError, setHasError] = useState<boolean>(isInvalid)
   const inputRef = useRef<HTMLInputElement>(null)
-  const [items, setItems] = useState<SelectChildWithStringType[]>(children)
+  const [filterValue, setFilterValue] = useState<string>('')
+
+  const items = React.Children.toArray(children).filter((child) => {
```
c9qsm62x, line 31: I agree with this approach store the text value in the state and filter in render

PR #3444, opened 2022-09-13
fix(Autocomplete): Don't clear external error state automatically

### LINKED ISSUE #3438: AutoComplete Control Error State [Red Border] gets lost after loosing the focus of the control
### Steps to reproduce
1. Create a Form with AutoComplete control which is Mandatory
2. Try submitting a form which clearly says that "You Must select xxxx" And also a red border around it. 
3. But, when we move the focus away from the auto complete control the Red Border gets lost in spite of the value not being selected. But, the error message stays on

### Minimal reproduction

Please provide a reference to a repository that includes a minimal reproduction of the issue.

We recommend spawning a new application using the [Defence Digital CRA Template](https://www.npmjs.com/package/cra-template-defencedigital):

```bash
// npx
npx create-react-app my-app --template defencedigital

// yarn create
yarn create react-app my-app --template defencedigital
```

### Screenshot
*Add a screenshot*
image

jpveooys:
## Related issue

Resolves #3438

## Overview

This stops an external error state being automatically cleared (e.g. on blur) in the Autocomplete component.

## Link to preview

https://5e25c277526d380020b5e418-hrvzlnpkuh.chromatic.com/?path=/docs/autocomplete--with-error

## Reason

The component doesn't know when the error has been fixed; only the parent component can indicate this by unsetting the prop.

## Work carried out

- [x] Update handling of `isInvalid` prop

### Before

Screencast_13_09_22_09:29:31

### After

Screencast_13_09_22_09:30:44

### DIFF
--- a/packages/react-component-library/src/components/Autocomplete/Autocomplete.test.tsx
+++ b/packages/react-component-library/src/components/Autocomplete/Autocomplete.test.tsx
@@ -1,10 +1,16 @@
-import { TypographyS } from '@defencedigital/design-tokens'
+import { ColorDanger800, TypographyS } from '@defencedigital/design-tokens'
 import React from 'react'
 import '@testing-library/jest-dom/extend-expect'
 import { render, RenderResult } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
 
+import { COMPONENT_SIZE } from '../Forms'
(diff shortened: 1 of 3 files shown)


thyhjwb6 (approved):
👍

PR #3507, opened 2022-11-01
build(Tooling): Update to Node.js 18 LTS

jpveooys:
## Overview

This updates to Node.js version 18 now it is in LTS. It also fixes Playwright artefacts from the shards overwriting each other on GitHub Actions.

## Reason

Node.js is the latest LTS version.

## Work carried out

- [x] Update `.nvmrc` and `package.json`
- [x] Update Playwright
- [x] Update GitHub Actions workflows
- [x] Configure Node.js to prefer IPv4 when resolving localhost on GitHub Actions
- [x] Preserve all Playwright shard artefacts on GitHub Actions

## Developer notes

[Node.js now prefers IPv6 during DNS resolutions](https://github.com/nodejs/node/pull/39987), and `http-server` doesn't support IPv6. This meant the end-to-end and accessibility tests didn't work on GitHub Actions.

This has been worked around by configured Node.js to prefer IPv4 again on GitHub Actions.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -149,7 +149,7 @@ jobs:
         shard: [1/3, 2/3, 3/3]
     steps:
       - name: Git clone repository
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: Set up environment
         uses: ./.github/actions/prepare-environment
@@ -157,7 +157,7 @@ jobs:
           unpack-prebuilt-libraries: true
 
       - name: Cache Playwright browsers
-        uses: actions/cache@v2
+        uses: actions/cache@v3
         with:
           path: ~/.cache/ms-playwright
           key: playwright-${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}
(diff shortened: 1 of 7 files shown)


jpveooys:
`CI=1 yarn --cwd packages/react-component-library test:e2e --shard 1/3` works locally, so it's not clear why it's failing here. Seems similar to https://github.com/microsoft/playwright/issues/18467.

.github/workflows/build_and_test.yml
```diff
@@ -177,12 +179,14 @@ jobs:
         run: tar -xzf /tmp/storybook-static-e2e.tar.gz
 
       - name: Run e2e tests
-        run: yarn --cwd packages/react-component-library test:e2e --shard ${{ matrix.shard }}
+        env:
+          NODE_OPTIONS: --dns-result-order=ipv4first
```
jpveooys, line 183: Scoped `NODE_OPTIONS` to the specific commands and not the whole workflow, as it started to affect actions used in other steps otherwise (as they use Node.js as well – some were on old versions of Node.js which don't support `--dns-result-order`).

adamledwards (approved):

PR #3509, opened 2022-11-04
fix(Select): Allow options to be disabled

### LINKED ISSUE #3508: Ability required to disable options in `Select`
## Description
In v2 of the `Select` it was possible to disable options because an `isDisabled` flag was being passed to `react-select`. v3 of the `Select` needs the ability to pass an `isDisabled` prop to an option. Early investigation seems to be that this can be done through Downshift's `getItemProps`, in particular there is a `disabled` but there are potentially knock on effects with styling.

Note that this is required for feature parity between v2 and v3.

jpveooys:
## Related issue

Resolves #3508 

## Overview

This exposes missing functionality in Select to allow options to be disabled.

## Link to preview

https://5e25c277526d380020b5e418-ftekgwmned.chromatic.com/?path=/story/select--default

## Reason

> In v2 of the `Select` it was possible to disable options because an `isDisabled` flag was being passed to `react-select`. v3 of the `Select` needs the ability to pass an `isDisabled` prop to an option. Early investigation seems to be that this can be done through Downshift's `getItemProps`, in particular there is a `disabled` but there are potentially knock on effects with styling.
> 
> Note that this is required for feature parity between v2 and v3.

## Work carried out

- [x] Add `isDisabled` prop to `SelectOption`
- [x] Update Downshift
- [x] Transpile Downshift in Storybook and CJS bundles

## Screenshot

image

## Notes

There was no lighter colour in the neutral palette available with sufficient contrast, so the text of non-disabled options was made darker.

A newer versions of Downshift was needed as it contains a bug fix for disabled items. However, newer versions of Downshift are no longer transpiled to ES5. This was worked around by transpiling it in our own bundles.

### DIFF
--- a/packages/react-component-library/.storybook/main.js
+++ b/packages/react-component-library/.storybook/main.js
@@ -20,6 +20,28 @@ module.exports = {
     ${head}
     ${process.env.NETLIFY ? newRelic.script : ''}
   `,
+  webpackFinal: (config) => {
+    // Transpile Downshift for IE11 compatibility
+    return {
+      ...config,
+      module: {
+        ...config.module,
+        rules: [
+          ...config.module.rules,
+          {
+            test: /\.m?js$/,
+            include: /node_modules\/downshift\//,
+            use: {
+              loader: 'babel-loader',
+              options: {
+                presets: ['@babel/preset-env'],
+              },
+            },
+          },
+        ],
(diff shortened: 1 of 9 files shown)


thyhjwb6 (approved):
👍

PR #3551, opened 2022-12-16
fix(Timeline): Use visually hidden text for header rows

jpveooys:
## Related issue

Split from #3522

## Overview

This makes Timeline header rows use visually hidden text rather than an `aria-label`.

## Link to preview

https://5e25c277526d380020b5e418-yfnaldinxs.chromatic.com/?path=/docs/compound-timeline--default

## Reason

This was being flagged by axe 4.5. More changes are required to get everything passing with axe 4.5, so this part was split out.

## Work carried out

- [x] Replace aria-labels with visually hidden text in Timeline header rows

### DIFF
--- a/packages/react-component-library/src/components/Timeline/TimelineRow.tsx
+++ b/packages/react-component-library/src/components/Timeline/TimelineRow.tsx
@@ -7,6 +7,7 @@ import { StyledNoEvents } from './partials/StyledNoEvents'
 import { StyledRow } from './partials/StyledRow'
 import { StyledRowContent } from './partials/StyledRowContent'
 import { StyledRowHeader } from './partials/StyledRowHeader'
+import { StyledVisuallyHiddenText } from '../../styled-components/partials/StyledVisuallyHiddenText'
 import { SubcomponentProps } from '../../common/SubcomponentProps'
 import { TimelineContext, TimelineEventsProps } from '.'
 import { useTimelineRowContent } from './hooks/useTimelineRowContent'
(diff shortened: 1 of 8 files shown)


packages/react-component-library/src/components/Timeline/TimelineRow.tsx
```diff
@@ -81,10 +82,14 @@ export const TimelineRow: React.FC<TimelineRowProps> = ({
...
-              aria-label={ariaLabel || name}
+              aria-label={ariaLabel}
               {...restHeaderProps}
             >
-              {render ? render({ name }) : name}
+              {render && render({ name })}
+              {!render && isHeader && (
+                <StyledVisuallyHiddenText>{name}</StyledVisuallyHiddenText>
+              )}
+              {!render && !isHeader && name}
```
jpveooys, line 92: Not particularly elegant – open to suggestions for alternatives...

packages/react-component-library/src/components/Timeline/__tests__/Timeline.test.tsx
```diff
@@ -386,12 +386,13 @@ describe('Timeline', () => {
...
+      it('adds text to the row headers', () => {
         const rowHeaders = wrapper.getAllByTestId('timeline-row-header')
 
-        expect(rowHeaders[0]).toHaveAttribute('aria-label', 'Months')
-        expect(rowHeaders[1]).toHaveAttribute('aria-label', 'Weeks')
-        expect(rowHeaders[2]).toHaveAttribute('aria-label', 'Days')
+        // Text is visually hidden
+        expect(rowHeaders[0]).toHaveTextContent('Months')
+        expect(rowHeaders[1]).toHaveTextContent('Weeks')
+        expect(rowHeaders[2]).toHaveTextContent('Days')
```
jpveooys, line 395: Not really a nice way to check that the text is actually visually hidden here (there are the styles but they're an implementation detail) – the visual regression tests should cover the visibility of the text in any case.

adamledwards (approved):

PR #3573, opened 2023-01-11
fix(Autocomplete): Fix <strong/> on empty string

### LINKED ISSUE #3572: `Autocomplete` adds empty `<strong />` elements when `value` is `""`
### Steps to reproduce
1. Set `value` to empty string
2. View source for options

### Minimal reproduction
```
  describe('strong reproduction', () => {
    beforeEach(() => {
      wrapper = render(
        <Autocomplete label="Label" value="">
          <AutocompleteOption value="one">One</AutocompleteOption>
          <AutocompleteOption value="two">Two</AutocompleteOption>
          <AutocompleteOption value="three">Three</AutocompleteOption>
        </Autocomplete>
      )

      return userEvent.click(wrapper.getByTestId('select-input'))
    })

    it('shows the items', () => {
      const options = wrapper.queryAllByTestId('select-option')
      expect(options[0].children[0].querySelectorAll('strong')).toHaveLength(0)
      expect(options[1].children[0].querySelectorAll('strong')).toHaveLength(0)
      expect(options[2].children[0].querySelectorAll('strong')).toHaveLength(0)
    })
  })
```

### Screenshot
Screenshot 2023-01-11 at 10 30 40

adamledwards:
## Related issue

Resolves #3572

## Overview

This PR adds tests for the filter and the expected HTML content. 
If the string is empty it will pass undefined so no replacements are made to the string

## Link to preview

[If you're on a project which auto-deploys branches, link to the branches preview link here]

## Reason
The package used to perform the replacement was matching an empty string
image

## Work carried out

- [x] Check for an empty string
- [x]  Add additional test case

### DIFF
--- a/packages/react-component-library/src/components/Autocomplete/AutocompleteOption.tsx
+++ b/packages/react-component-library/src/components/Autocomplete/AutocompleteOption.tsx
@@ -3,7 +3,6 @@ import reactStringReplace from 'react-string-replace'
 
 import { SelectBaseOption, SelectBaseOptionAsStringProps } from '../SelectBase'
 
-// eslint-disable-next-line @typescript-eslint/no-empty-interface
 export interface AutocompleteOptionProps extends SelectBaseOptionAsStringProps {
   inputValue?: string
 }
@@ -14,9 +13,11 @@ export const AutocompleteOption = React.forwardRef<
 >(({ children, inputValue, ...rest }, ref) => {
   return (
     <SelectBaseOption ref={ref} {...rest}>
-      {reactStringReplace(children, inputValue, (match, i) => (
-        <strong key={i}>{match}</strong>
-      ))}
+      {inputValue
+        ? reactStringReplace(children, inputValue, (match, i) => (
+            <strong key={i}>{match}</strong>
+          ))
+        : children}
     </SelectBaseOption>
   )
 })
--- a/packages/react-component-library/src/components/SelectBase/SelectBaseOption.tsx
+++ b/packages/react-component-library/src/components/SelectBase/SelectBaseOption.tsx
@@ -6,7 +6,7 @@ import { StyledOptionBadge } from './partials/StyledOptionBadge'
 import { StyledOptionText } from './partials/StyledOptionText'
 import { ComponentWithClass } from '../../common/ComponentWithClass'
 
-export interface SelectBaseOptionBaseProps extends ComponentWithClass {
+export interface SelectBaseOptionProps extends ComponentWithClass {
   badge?: string | number
   icon?: React.ReactNode
   isHighlighted?: boolean
@@ -14,19 +14,9 @@ export interface SelectBaseOptionBaseProps extends ComponentWithClass {
   title?: string
 }
 
-export interface SelectBaseOptionAsArrayProps
-  extends SelectBaseOptionBaseProps {
-  children: React.ReactNodeArray
-}
-
-export interface SelectBaseOptionAsStringProps
-  extends SelectBaseOptionBaseProps {
+export interface SelectBaseOptionAsStringProps extends SelectBaseOptionProps {
   children: string
 }
-
-export type SelectBaseOptionProps =
-  | SelectBaseOptionAsArrayProps
-  | SelectBaseOptionAsStringProps
 
 export const SelectBaseOption = React.forwardRef<
   HTMLLIElement,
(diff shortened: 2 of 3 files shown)


packages/react-component-library/src/components/Autocomplete/AutocompleteOption.tsx
```diff
@@ -14,7 +14,7 @@ export const AutocompleteOption = React.forwardRef<
 >(({ children, inputValue, ...rest }, ref) => {
   return (
     <SelectBaseOption ref={ref} {...rest}>
-      {reactStringReplace(children, inputValue, (match, i) => (
+      {reactStringReplace(children, inputValue || undefined, (match, i) => (
```
jpveooys: It seems to match options containing the text 'Undefined' like this unfortunately.

image

I think we probably need to skip `reactStringReplace` completely if `inputValue` is falsey.

packages/react-component-library/src/components/SelectBase/SelectBaseOption.tsx
```diff
@@ -14,20 +14,10 @@ export interface SelectBaseOptionBaseProps extends ComponentWithClass {
   title?: string
 }
 
-export interface SelectBaseOptionAsArrayProps
-  extends SelectBaseOptionBaseProps {
-  children: React.ReactNodeArray
+export interface SelectBaseOptionProps extends SelectBaseOptionBaseProps {
+  children: React.ReactNode
```
adamledwards, line 18: The union type of not needed here for children. The type ReactNode contains `string` and `Iterable<React.ReactNode>`
```ReactNode == string | number | boolean | {} | React.ReactElement<any, string | React.JSXElementConstructor<any>> | Iterable<React.ReactNode> | React.ReactPortal | null | undefined```

File: packages/react-component-library/src/components/Autocomplete/AutocompleteOption.tsx [hunk 3]
```diff
@@ -1,22 +1,25 @@
...
+export interface AutocompleteOptionProps extends SelectBaseOptionProps {
   inputValue?: string
 }
 
 export const AutocompleteOption = React.forwardRef<
   HTMLLIElement,
   AutocompleteOptionProps
 >(({ children, inputValue, ...rest }, ref) => {
+  const optionElement =
+    inputValue && typeof children === 'string'
```
adamledwards: I had to add type string to appease ts 
I was getting `Type 'null' is not assignable to type 'string | ReactNodeArray | undefined'.`

packages/react-component-library/src/components/Autocomplete/AutocompleteOption.tsx
adamledwards: 👍

packages/react-component-library/src/components/Autocomplete/AutocompleteOption.tsx
```diff
@@ -1,22 +1,25 @@
 import React from 'react'
 import reactStringReplace from 'react-string-replace'
 
-import { SelectBaseOption, SelectBaseOptionAsStringProps } from '../SelectBase'
+import { SelectBaseOption, SelectBaseOptionProps } from '../SelectBase'
 
-// eslint-disable-next-line @typescript-eslint/no-empty-interface
-export interface AutocompleteOptionProps extends SelectBaseOptionAsStringProps {
+export interface AutocompleteOptionProps extends SelectBaseOptionProps {
```
jpveooys: Is there any need to change the external type here? (Assume it was restricted to a string for children on purpose.)

packages/react-component-library/src/components/SelectBase/SelectBaseOption.tsx
```diff
@@ -6,28 +6,17 @@ import { StyledOptionBadge } from './partials/StyledOptionBadge'
...
 }
-
-export interface SelectBaseOptionAsArrayProps
-  extends SelectBaseOptionBaseProps {
-  children: React.ReactNodeArray
-}
-
-export interface SelectBaseOptionAsStringProps
-  extends SelectBaseOptionBaseProps {
+export interface SelectBaseOptionAsStringProps extends SelectBaseOptionProps {
```
jpveooys, line 17: Seems like the blank line between the interfaces has been lost here – would be good to add it back in

packages/react-component-library/src/components/Autocomplete/AutocompleteOption.tsx
jpveooys: Should be able to omit it if `AutocompleteOptionProps` is made to extend from `SelectBaseOptionAsStringProps` again.

jpveooys (commented):
Nice work! Just that comment about AutocompleteOptionProps mainly.

packages/react-component-library/src/components/Autocomplete/AutocompleteOption.tsx
adamledwards: 👍

jpveooys (approved):
👍

Issue #3618, opened 2023-03-23
Updating `endDate` from outside the `Timeline` produces unexpected results


thyhjwb6:
### Description
First load appears to be correct but when clicking the buttons in the example, the range is not updated correctly.

### Storybook example
```
export const Repro = () => {
  const TimelineRepro: React.FC = () => {
    const [startDate, setStartDate] = useState<Date>(new Date(2023, 0, 1))
    const [endDate, setEndDate] = useState<Date>(new Date(2023, 5, 1))

    return (
      <>
        <Button
          onClick={() => {
            setStartDate(addMonths(startDate, -6))
            setEndDate(addMonths(endDate, -6))
          }}
        >
          Prev
        </Button>
        <Button
          onClick={() => {
            setStartDate(addMonths(startDate, 6))
            setEndDate(addMonths(endDate, 6))
          }}
        >
          Next
        </Button>
        <Timeline startDate={startDate} endDate={endDate}>
          <TimelineTodayMarker />
          <TimelineMonths />
          <TimelineDays />
          <TimelineRows>{}</TimelineRows>
        </Timeline>
      </>
    )
  }
  return <TimelineRepro />
}
Repro.storyName = 'Repro'
```

### Unit test
```
  describe('`endDate` bug', () => {
    const TimelineRepro: React.FC = () => {
      const [startDate, setStartDate] = useState<Date>(new Date(2023, 0, 1))
      const [endDate, setEndDate] = useState<Date>(new Date(2023, 5, 1))

      return (
        <>
          <Button
            onClick={() => {
              setStartDate(addMonths(startDate, -6))
              setEndDate(addMonths(endDate, -6))
            }}
          >
            Prev
          </Button>
          <Button
            onClick={() => {
              setStartDate(addMonths(startDate, 6))
              setEndDate(addMonths(endDate, 6))
            }}
          >
            Next
          </Button>
          <Timeline startDate={startDate} endDate={endDate}>
            <TimelineTodayMarker />
            <TimelineMonths />
            <TimelineDays />
            <TimelineRows>{}</TimelineRows>
          </Timeline>
        </>
      )
    }

    it('sets the months correctly', () => {
      const { getByText, getAllByTestId } = render(<TimelineRepro />)

      const originalMonths = getAllByTestId('timeline-month')
      expect(originalMonths[0]).toHaveTextContent('Jan2023')
      expect(originalMonths[1]).toHaveTextContent('Feb2023')
[...]

PR #3619, opened 2023-03-24
fix(Timeline): Fix issue with the timeline when updating the `endDate`

adamledwards:
## Related issue

#3618 

## Overview
The endDate prop incorrectly updates the Timline, this PR fixes that to enable the Timeline to be driven by the props.
I do have some concerns about the timeline being unpredictable because the UI state is being driven from external props and the internal state. 
consider the following code 

```tsx
import { Timeline } from '@defencedigital/react-component-library'
import useDataFetcher from './some-service'

function App() {
  const data = useDataFetcher(...)
  const startDate = new Date(data.startDateIsoString)
  const endDate = new Date(data.endDateIsoString)

  return (
    <Timeline startDate={startDate} endDate={endDate}></Timeline>
  )
}
```

If a user clicked the next or previous button after there was a rerender from a parent component the timeline would revert to the original position, this is because the equality check is done on a Date object and not on the value of the date.
I would recommend using the `startDate` and `endDate`  props with `hideToolbar=true`. So the Timeline can be controlled from the props only.

**Example.**

- [x] A bit of work I completed
- [ ] A bit of work I did not complete

## Screenshot

[If the work is UI related then paste a screenshot of the update here.]

## Developer notes

[Sometimes, extra notes are needed to add clarity to a PR, add them here]

### DIFF
--- a/packages/react-component-library/src/components/Timeline/__tests__/Timeline.test.tsx
+++ b/packages/react-component-library/src/components/Timeline/__tests__/Timeline.test.tsx
@@ -49,7 +49,7 @@ const TimelineDates: React.FC<TimelineDaysProps> = () => {
 
 describe('Timeline', () => {
   let wrapper: RenderResult
-  let user: ReturnType<typeof userEvent['setup']>
+  let user: ReturnType<(typeof userEvent)['setup']>
 
   beforeEach(() => {
     jest.useFakeTimers()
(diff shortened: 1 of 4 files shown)


packages/react-component-library/src/components/Timeline/__tests__/Timeline.test.tsx
```diff
@@ -49,7 +49,7 @@ const TimelineDates: React.FC<TimelineDaysProps> = () => {
 
 describe('Timeline', () => {
   let wrapper: RenderResult
-  let user: ReturnType<typeof userEvent['setup']>
+  let user: ReturnType<(typeof userEvent)['setup']>
```
adamledwards, line 52: I'll check why my IDE decided to do this

packages/react-component-library/src/components/Timeline/context/types.ts
```diff
@@ -58,7 +58,7 @@ export type TimelineState = {
 }
 
 export const TIMELINE_ACTIONS = {
-  CHANGE_START_DATE: 'CHANGE_START_DATE',
```
jpveooys, line 61: FYI the actions are public, so we'll probably want to keep `CHANGE_START_DATE` as an alias of `CHANGE_DATE`.

packages/react-component-library/src/components/Timeline/__tests__/Timeline.test.tsx
```diff
@@ -2374,6 +2374,121 @@ describe('Timeline', () => {
     })
   })
 
+  describe('when `startDate`  and `endDate` is updated externally', () => {
```
jpveooys: Minor: extra space here. 
```suggestion
  describe('when `startDate` and `endDate` is updated externally', () => {
```

jpveooys (commented):
Approach looks reasonable to me :+1:  (Did some manual testing in Storybook as well and all seemed well.)

jpveooys (approved):

Issue #3630, opened 2023-04-18
Allow `Dialog` arbitrary `description` JSX


m7kvqbe1:
### Description

Currently any content passed to a `Dialog` component via the `description` prop is wrapped in a `p` element and it only accepts strings.

This flags React warnings downstream when we supply rich content (design requirement). 

Enhance to be more flexible.

PR #3699, opened 2023-12-05
feat(Dialog): Enhance to support rich description content

m7kvqbe1:
## Related issue

#3630

## Overview

Enhance to support rich description content (arbitrary JSX).

## Reason

>Currently any content passed to a Dialog component via the description prop is wrapped in a p element and it only accepts strings. This flags React warnings downstream when we supply rich content (design requirement).

## Work carried out

- [x] Loosen the typings
- [x] Update wrapping element
- [x] Update tests and add new story

## Screenshot

Screenshot 2023-05-25 at 11 31 27


ghost (approved):
LGTM

PR #3706, opened 2023-12-15
fix(Popover): Respect existing callbacks

### LINKED ISSUE #3705: Popover is not respecting existing callbacks on render targer
### Description
If the render target has `onMouseEnter` and `onMouseLeave` callbacks then they are overwritten for rendering a popover. The existing callbacks need to be respected.

thyhjwb6:
## Related issue
Fixes #3705 

## Overview
This change makes it so that existing callbacks are still called.

## Reason
Callbacks were being overridden.

## Work carried out
- [x] Fix `onMouseEnter` and `onMouseLeave` not respected

### DIFF
--- a/packages/react-component-library/src/components/Popover/Popover.test.tsx
+++ b/packages/react-component-library/src/components/Popover/Popover.test.tsx
@@ -249,4 +249,27 @@ describe('Popover', () => {
       )
     })
   })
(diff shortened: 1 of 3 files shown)


m7kvqbe1 (approved):
LGTM

PR #3717, opened 2024-01-23
fix(Popover): Prevent pending timer being cleared early

m7kvqbe1:
## Overview

Prevent pending `hideElement` timer from being wrongly cleared.

## Reason

Invocation of `onMouseEnter` handler exposed by `useHideShow` hook in quick succession has the potential to wrongly clear a pending timer set by `hideElement`. This is particularly apparent when multiple Popover instances are used adjacent to one another with small target elements. The end result would be Popovers left open when they should have closed.

## Work carried out

- [x] Add automated test
- [x] Add logic to return early

## Developer notes

This has already been tested downstream in a consuming application.

### DIFF
--- a/packages/react-component-library/src/components/Popover/Popover.test.tsx
+++ b/packages/react-component-library/src/components/Popover/Popover.test.tsx
@@ -6,7 +6,9 @@ import {
   render,
   RenderResult,
   waitFor,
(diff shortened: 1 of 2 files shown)


thyhjwb6 (approved):
👍

PR #3718, opened 2024-02-02
feat(Table): Add onRowClick event

ghost:
## Overview
Add an `onRowClick` event to the Table component

## Reason
Allows component consumers to react to clicks on items in the table

## Work carried out
- [x] New optional prop for `onRowClick` added to `TableProps`
- [x] Update tests

### DIFF
--- a/packages/react-component-library/src/components/Table/Table.tsx
+++ b/packages/react-component-library/src/components/Table/Table.tsx
@@ -26,6 +26,11 @@ export interface TableProps extends ComponentWithClass {
    * Optional text caption to display above the component.
    */
   caption?: string
+  /**
+   * Optional, handler invoked when a row is clicked.
+   * @param id - the id of the row
+   */
+  onRowClick?: (id: string) => void
 }
 
 function getKey(prefix: string, id: string) {
@@ -37,6 +42,7 @@ export const Table: React.FC<TableProps> = ({
   caption,
   children,
   className,
+  onRowClick,
   ...rest
 }) => {
   const { tableData, sortTableData, sortField, sortOrder } = useTableData(data)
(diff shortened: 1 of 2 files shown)


packages/react-component-library/src/components/Table/Table.tsx
```diff
@@ -26,6 +26,11 @@ export interface TableProps extends ComponentWithClass {
    * Optional text caption to display above the component.
    */
   caption?: string
+  /**
+   * Optional, callback when a row is clicked
```
m7kvqbe1:
```suggestion
   * Optional handler invoked when a row is clicked.
```

m7kvqbe1: Sorry I know this is really nitpicky. 

Just trying to align it with style of other prop interface comments (tone of voice / full stop).

packages/react-component-library/src/components/Table/Table.tsx
```diff
@@ -61,6 +75,7 @@ export const Table: React.FC<TableProps> = ({
         <tbody>
           {tableData.map((row: RowProps) => (
             <StyledTableRow
+              onClick={() => handleRowClick(row.id)}
```
m7kvqbe1, line 78: Might be able to do this inline with optional chaining `onRowClick?(id)`.

Fine as is though.

m7kvqbe1 (commented):
LGTM just change to the interface comment. 

Only nitpicky about these because they end up in the Storybook docs.

packages/react-component-library/src/components/Table/Table.tsx
ghost: 👍 Done. Totally agree about consistency and tone.

m7kvqbe1 (approved):

PR #3719, opened 2024-02-05
fix(Popover): Avoid timeout usage for a 0 closeDelay

m7kvqbe1:
## Overview

Avoid redundant `setTimeout` usage for a `closeDelay` of 0.

## Reason

This was causing issues downstream where a large number of Popovers were used with small target elements.

In some instances Popovers would persist and no longer close.

## Work carried out

- [x] Resolve bug 
- [x] Fix wrongly implemented Masthead automated test


ghost (approved):
👍 LGTM

PR #3735, opened 2024-03-21
fix(SectionDivider): Apply supplied className 

m7kvqbe1:
- Resolve issue with Masthead styling
- Resolve issue with SectionDivider styling
- Add resolutions for audit flagged deps


markhigham (changes requested):
Can you remove the boilerplate from the PR description.
Otherwise LGTM

PR #3737, opened 2024-03-25
feat(DataGrid): Create base implementation

### LINKED ISSUE #2657: Table with selectable rows and also checkbox selection
### Description

- Make the table component more interactive by providing selectable rows and hover styles when rows are selectable. 
- Allow for checkbox selection on rows

I ended up building this out of the partial components of the current table, for a link to the repo contact me via NDS slack

### LINKED ISSUE #3730: DataGrid component
### Description

As recently discussed, there is a common need for an advanced table component that is specifically designed for data.

For our project, the following would be useful (comparing with the Table component):

- **Checkboxes for each row, as per #2657** 

  This includes a select all checkbox in the header, and shading of selected rows. It should also be possible to disable some rows/checkboxes. See also my comment on that issue.

- **Review of font size and use of capitals in the header**

   These both seem inconsistent with other components in the design system. On the font size, it becomes noticeable when comparing with other things on the page, and when you embed other components in a table cell.

- **Review of background colour**

  It currently doesn't have one, so it ends up being transparent. Perhaps it should have a default background colour, and people can override it using Styled Components if they want something different.

- **Perhaps easier control over column widths**

  Currently Table uses `table-layout: fixed`, which results in each column having an equal width by default (not so useful when the contents of some columns are wider than others).

  `table-layout: auto` works okay for us for now. But I can imagine scenarios where you would want `table-layout: fixed` with percentages for some columns (e.g. if you have pagination, you probably don't want the column widths to change when navigating between pages). This could be probably be currently done with the `css` prop on `TableColumn`, but perhaps it could be nicer.

- **Right-aligned columns**

  There may be cases where right-aligned columns (including the heading) are needed e.g. for numbers.

If I can think of anything else will post back here.

m7kvqbe1:
## Related issue

Closes #3730, Closes #2657

## Overview

Create baseline DataGrid that has feature parity with existing Table component. 

Enhance to include some other additional baseline features.

## Reason

Table is lacking the features required to interrogate and manipulate large data sets.

## Work carried out

- [x] Feature parity with Table component (Column sorting, Arbitrary cell content, Caption)
- [x] Add ability to select rows with Checkbox and All Rows (header) + callbacks
- [x] Hover and active state styling for individual rows + callbacks
- [x] Solid background color as opposed to transparent
- [x] Consumer controlled column sizing (https://tanstack.com/table/latest/docs/guide/column-sizing) 
- [x] Enable custom alignment for columns

## Screenshot

Screenshot 2024-04-17 at 08 51 23
Screenshot 2024-04-17 at 08 51 30
Screenshot 2024-04-17 at 08 51 42

## Developer notes

https://github.com/Royal-Navy/design-system/issues/3730

https://github.com/Royal-Navy/design-system/issues/2657


m7kvqbe1:
@jpveooys @thyhjwb6

packages/react-component-library/src/components/DataGrid/DataGrid.stories.tsx
```diff
@@ -0,0 +1,228 @@
+import React from 'react'
+import { fn } from '@storybook/test'
+import { faker } from '@faker-js/faker'
+import { StoryFn, Meta } from '@storybook/react'
+import { ColumnDef } from '@tanstack/react-table'
+
+import { DataGrid } from '.'
+import { Badge } from '../Badge'
+
+export type Order = {
```
markhigham: does this need to be exported?

packages/react-component-library/src/components/DataGrid/DataGrid.test.tsx
```diff
@@ -0,0 +1,439 @@
...
+    expect(checkboxes[0]).not.toBeChecked()
+    expect(checkboxes[1]).not.toBeChecked()
+    expect(checkboxes[2]).not.toBeChecked()
+    expect(checkboxes[3]).not.toBeChecked()
+
+    userEvent.click(checkboxes[1], {
+      pointerEventsCheck: PointerEventsCheckLevel.Never,
+    })
+
+    await new Promise((resolve) => setTimeout(resolve, 100))
```
markhigham: :)

packages/react-component-library/src/components/DataGrid/DataGrid.test.tsx
```diff
@@ -0,0 +1,439 @@
...
+        })
+      })
+    })
+
+    describe('when the first cell header is clicked once and then the third cell header is clicked', () => {
+      beforeEach(async () => {
+        userEvent.click(screen.getByText('First'))
+        await new Promise((resolve) => setTimeout(resolve, 100))
+        userEvent.click(screen.getByText('Third'))
+        await new Promise((resolve) => setTimeout(resolve, 100))
```
markhigham: We use this hack frequently. How about creating a single function that does the same thing. 
I'm thinking that we can then annotate that function with a note to future maintainers explaining why we arrived at this hack

`await hackyWaitForHook()`

```
async function hackyWaitForHook() {
    // Apologies for the hack but we needed to to do this because ...
    return new Promise((resolve) => setTimeout(resolve, 100))
}
```

packages/react-component-library/src/components/DataGrid/DataGrid.tsx
```diff
@@ -0,0 +1,306 @@
...
+  RowSelectionState,
+  TableOptions,
+} from '@tanstack/react-table'
+import {
+  IconSortAscending,
+  IconSortDescending,
+  IconSortUnsorted,
+} from '@royalnavy/icon-library'
+
+import { StyledDataGrid } from './partials/StyledDataGrid'
```
markhigham: maybe add an `index.ts` to `./partials` to simplify these imports

markhigham (approved):
LGTM: We should add a follow up tickets once #3746 is merged and `mdx` is enabled.
- Add a description mdx file for each of the new storybook folders. 
- Document why Table is deprecated in the mdx
- Add some text to the deprecated stories with links to the docs

Issue #3734, opened 2024-03-18
Allow arbitrary content in Alert


jpveooys:
### Description

Currently `children` in Alert is typed as `string`:

https://github.com/Royal-Navy/design-system/blob/ab175e02beaa4e7e6fea7992766f5dbab620cbd4/packages/react-component-library/src/components/Alert/Alert.tsx#L40

If there isn't a reason for that, it'd be helpful to allow arbitrary content. (The design for our app currently has alerts with more complex content.)

Note also that `children` is currently placed inside a `p` element. A `p` element only allows phrasing content as children, so some elements (`div`, `ul`, `dl` etc) are not valid children of `p`.

PR #3740, opened 2024-03-29
feat(Alert): Allow arbitrary content in Alert

markhigham:
## Related issue

#3734

## Overview

Changed `children` props of Alert component from string to `ReactNode` and changed container element from `<p>` to `<div>` 

## Reason
This allows Design System users to include arbitrary html in their alerts.

## Work carried out

- [x] Update props for Alert
- [x] Added unit tests
- [x] Added storybook example

## Screenshot

image

### DIFF
--- a/packages/react-component-library/src/components/Alert/Alert.stories.tsx
+++ b/packages/react-component-library/src/components/Alert/Alert.stories.tsx
@@ -33,3 +33,25 @@ WithoutTitle.args = {
 }
 
 WithoutTitle.storyName = 'Without title'
+
+export const ArbitraryContent: StoryFn<typeof Alert> = ({
+  title,
+  children,
+  variant,
+}) => (
+  <Alert variant={variant} title={title}>
+    {children}
+  </Alert>
+)
+
+ArbitraryContent.args = {
+  title: 'Using arbitrary JSX as content',
+  children: (
+    <div>
+      <p>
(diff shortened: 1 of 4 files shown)


packages/react-component-library/src/components/Alert/Alert.stories.tsx
```diff
@@ -33,3 +36,23 @@ WithoutTitle.args = {
...
+  background-size: contain;
+  height: 3rem;
+`
+
+ArbitraryContent.args = {
+  title: 'Using arbitrary JSX as content',
+  children: <div><StyledLogoContainer />
```
m7kvqbe1: Although it's now loosened and possible to do this i'd rather us include a more simplistic example.

Maybe just `<strong>Hello, World!</strong>` or something?

My thinking around it is I don't want people to think it's ok to put images etc in an Alert.

packages/react-component-library/src/components/Alert/Alert.stories.tsx
```diff
@@ -1,8 +1,11 @@
 import React from 'react'
 import { StoryFn, Meta } from '@storybook/react'
+import styled from 'styled-components'
 
 import { Alert, ALERT_VARIANT } from './index'
 
+
```
m7kvqbe1: Extra newline?

m7kvqbe1 (approved):

Issue #3682, opened 2023-11-14
Allow `ListItem` to accept `ReactNode` as a child


oliverrees:
### Steps to reproduce
1. Create a `<List>` with a `<ListItem>`
2. Try to add a ReactNode as a child of the `<ListItem>`
3. You will receive a type error

Accoring to the [ListItem documentation](https://storybook.design-system.digital.mod.uk/?path=/docs/list--default) a `ListItem` should accept `(string | string[]) & ReactNode` as children. However, in the [Type definition](https://github.com/defencedigital/mod-uk-design-system/blob/9eb0eb2597b4cbcba88d0e1c2a158ec0e3a9c3ba/packages/react-component-library/src/components/List/ListItem.tsx) only `string | string[]` are allowed, dramatically reducing the customisation possible in a `ListItem`.

PR #3741, opened 2024-04-03
fix(List): Allow arbitrary contents in List Item

markhigham:
## Related issue

#3682

## Overview

Change to `ListItem` props to allow arbitrary contents.

## Reason

Taken from the Issue
> Accoring to the [ListItem documentation](https://storybook.design-system.digital.mod.uk/?path=/docs/list--default) a ListItem should accept (string | string[]) & ReactNode as children. However, in the [Type definition](https://github.com/defencedigital/mod-uk-design-system/blob/9eb0eb2597b4cbcba88d0e1c2a158ec0e3a9c3ba/packages/react-component-library/src/components/List/ListItem.tsx) only string | string[] are allowed, dramatically reducing the customisation possible in a ListItem

## Work carried out
- [x] Change `ListItem` props to accept `ReactNode`
- [x] Unit tests
- [x] Update to storybook 

## Screenshot
image

### DIFF
--- a/packages/react-component-library/src/components/List/ListItem.tsx
+++ b/packages/react-component-library/src/components/List/ListItem.tsx
@@ -11,9 +11,9 @@ import { useExternalId } from '../../hooks/useExternalId'
 
 export interface ListItemProps extends ComponentWithClass {
   /**
-   * Description text to display for an individual item.
+   * Description to display for an individual item.
    */
-  children: string | string[]
+  children: React.ReactNode
   /**
    * Toggles whether the item is in an active state.
    */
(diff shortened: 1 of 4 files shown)


packages/react-component-library/src/components/List/ListItem.tsx
```diff
@@ -13,7 +13,7 @@ export interface ListItemProps extends ComponentWithClass {
   /**
    * Description text to display for an individual item.
    */
-  children: string | string[]
+  children: string | string[] | React.ReactNode
```
m7kvqbe1: I think `string | string[]` is redundant because it's implicitly included as part of the `ReactNode` type.

m7kvqbe1 (approved):

PR #3754, opened 2024-04-16
fix(PhaseBanner): Wrap text to the right of the phase badge

### LINKED ISSUE #3733: Phase banner doesn't wrap well on small viewports
### Steps to reproduce

1. Go to https://storybook.design-system.navy.digital.mod.uk/iframe.html?args=&id=phase-banner--default&viewMode=story
2. Make the viewport narrow

The main text wraps below the badge and is indented:

image

I think it would make sense if the main text stayed to the right of the badge (and wrapped there). As a comparison, the GOV.UK Design System phase banner wraps like this:

image

markhigham:
## Related issue

Closes #3733 

## Overview

Wrap the banner text away from the badge

## Reason
Better design treatment and consistency with GDS

## Work carried out

- [x] Add CSS to handle the wrapping
- [x] Add new story with narrow example 
- [x] Add unit tests

## Screenshot
image

### DIFF
--- a/packages/react-component-library/src/components/PhaseBanner/PhaseBanner.stories.tsx
+++ b/packages/react-component-library/src/components/PhaseBanner/PhaseBanner.stories.tsx
@@ -1,5 +1,6 @@
 import React from 'react'
 import { StoryFn, Meta } from '@storybook/react'
+import styled from 'styled-components'
 
 import { PhaseBanner } from '.'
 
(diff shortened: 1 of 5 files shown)


m7kvqbe1:
When it wraps do you think the Alpha badge should be vertically centered?

m7kvqbe1 (commented):
When it wraps do you think the Alpha badge should be vertically centered?

markhigham:
> When it wraps do you think the Alpha badge should be vertically centered?

I used the same layout as the GDS component 
https://design-system.service.gov.uk/components/phase-banner/default/

m7kvqbe1 (approved):

Issue #3756, opened 2024-04-17
Make Masthead responsive


jpveooys:
### Description

Currently the Masthead doesn't seem to be responsive.

Standard Masthead with a small viewport:

![image](https://github.com/Royal-Navy/design-system/assets/66470099/7857a4ad-90f3-467b-a654-0a47b1144244)

Inline nav Masthead:

![image](https://github.com/Royal-Navy/design-system/assets/66470099/913aff16-23e3-4de6-9201-03fbe3035fb0)

It would be good if it was responsive.

m7kvqbe1:
Some design work has now been done to make the Masthead responsive. We'll expand upon this soon.

PR #3764, opened 2024-04-28
chore(Masthead): Make masthead nav handle smaller viewports

markhigham:
## Related issue

#3756 

## Overview

This is the first step towards making Masthead responsive. We've changed the overflow treatment for `MastheadNav` and individual `StyledTabNav` to prevent wrapping

## Reason

Downstream consumers of the Design system need to support smaller viewports

## Work carried out
- [x] Make Masthead Inline nav handle smaller viewports by preventing overflow

## Screenshot

See how this doesn't _fix_ responsive layout, but is better than previous treatments.
image


m7kvqbe1 (approved):
LGTM - should the commit message be a `fix` type?

PR #3772, opened 2024-04-30
fix(ProgressIndicator): Avoid downstream runtime warning from Next.js

### LINKED ISSUE #3714: Incorrect transform-origin prop in IconLoader
### Steps to reproduce

1. Try to use the `ProgressIndicator` component in a Next.js 13.5 project
2. Check console output (server or browser)

There is a warning (accompanied by a stack trace): 

```
Invalid DOM property `transform-origin`. Did you mean `transformOrigin`?
```

This appears to be happening in the build process for the icon library. It is most easily seen by building the icon library locally and observing that the incorrect prop is in `packages/icon-library/src/icons/Loader.tsx`.

It may be an upstream bug in SVGR, or SVGO going wrong. (It's a style rather than an attribute in the original SVG: https://github.com/Royal-Navy/design-system/blob/b03e891d339bc678d0b2516206571c8145058393/packages/icon-library/src/assets/icons/standards/loader.svg#L23)

### Minimal reproduction

Let me know if those details aren't sufficient.

m7kvqbe1:
## Related issue

Closes #3714

## Overview

Apply via css property in keyframe definition to avoid inline `transform-origin` upon on transpilation.

## Reason

Next.js is wrongly throwing a runtime warning about the `transform-origin` SVG attribute.

[This is a valid HTML SVG attribute.](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/transform-origin)

## Work carried out

- [x] Upgrade @svgr/cli to latest
- [x] Avoid inline `transform-origin` usage
- [x] Fix `loader.svg` code formatting

## Screenshot

Screenshot 2024-04-30 at 11 57 14

## Developer notes

https://github.com/vercel/next.js/issues/53342#issuecomment-2084923492

https://github.com/facebook/react/pull/26130/files#diff-67b481b96d526280780f010dfd192761f2d1f6dc812e3015020a7986501d2ebaR524

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,8 @@
   "resolutions": {
     "lerna/axios": "^1.6.0",
     "**/tar": "^6.2.1",
-    "**/semver": "^7.5.2"
+    "**/semver": "^7.5.2",
+    "**/ejs": "^3.1.10"
   },
   "prettier": {
     "bracketSpacing": true,
--- a/packages/icon-library/package.json
+++ b/packages/icon-library/package.json
@@ -43,7 +43,7 @@
     "@babel/preset-react": "^7.8.3",
     "@babel/preset-typescript": "^7.8.3",
     "@royalnavy/eslint-config-react": "^4.5.2",
-    "@svgr/cli": "^6.1.2",
+    "@svgr/cli": "^8.1.0",
     "@types/react": "^17.0.38",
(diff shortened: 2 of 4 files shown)


markhigham (approved):
Some merge conflicts. Changes LGTM

PR #3785, opened 2024-05-07
fix(Modal): Apply focus trap and expose imperative handle

### LINKED ISSUE #3757: Add focus trap to Modal
### Description

Currently the Modal doesn't have a focus trap. This means when a Modal is active, you can still press Tab on the keyboard to focus and then interact with things outside the modal.

IIRC this can be resolved by using the [`<dialog>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog) element in modal mode i.e. using showModal (if support for it is considered good enough now).

Alternatively, focus-trap-react could be used (as it is in DatePicker).

m7kvqbe1:
## Related issue

Closes #3757

## Overview

Apply focus trap to Modal and Dialog components (use native Dialog).

Expose show and close methods via forwardedRef and imperative handle.

## Reason

Modal and Dialog were not correctly applying a focus trap.

## Work carried out

- [x] Fix styling issues with Close button (use SVG Icon)
- [x] Refactor to utilise native Dialog element
- [x] Expose show and hide methods via imperative handle and isOpen state
- [x] Forward supplied ref from Dialog component to underlying Modal
- [x] Fix styling inconsistency with Drawer close button (now matches Modal)

## Developer notes

When a dialog opens, the browser, by default, gives focus to the first element that can be focused within the dialog (always the close button if it is specified).

The Escape key can also now be used to close the modal.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog


packages/react-component-library/.storybook/hooks/useIsInDocs.test.ts
```diff
@@ -0,0 +1,42 @@
+import { renderHook, act } from '@testing-library/react-hooks'
+
+import { useIsInDocs } from './useIsInDocs'
+
+describe('useIsInDocs', () => {
+  const toggleSbDocsClass = (present) => {
```
markhigham: pedantic but if you pass a boolean then the function isn't _really_ a toggle

markhigham (approved):
Maybe some duplication between Modal and Dialog. Perhaps add a ticket to refactor

m7kvqbe1:
> Maybe some duplication between Modal and Dialog. Perhaps add a ticket to refactor

Dialog is just a wrapper around Modal really - it's being prescriptive about how Modal is composed. 

It exposes an isDanger prop, onCancel and onConfirm handler.

PR #3800, opened 2024-05-15
fix(TextInput): Focus on input when adornments are clicked

### LINKED ISSUE #3787: When clicking the icon the input should have focus on `TextInputE`
The icon is blocking focus - probably needs `pointer-events: none`.

markhigham:
## Related issue

Closes #3787

## Overview

Set focus to the text input when either of the adornments are clicked

## Reason

Better UX

## Work carried out

- [x] Added an onclick handler to the start and end adornment components

### DIFF
--- a/packages/react-component-library/src/components/TextInput/TextInput.test.tsx
+++ b/packages/react-component-library/src/components/TextInput/TextInput.test.tsx
@@ -9,6 +9,37 @@ import { TextInput } from '.'
 describe('TextInput', () => {
   let wrapper: RenderResult
 
+  describe('adornment', () => {
+    let onBlurSpy: (event: React.FormEvent) => void
+    let onChangeSpy: (event: React.ChangeEvent<HTMLInputElement>) => void
+    const adornmentId = 'adornment-test-id'
+    beforeEach(() => {
+      onBlurSpy = jest.fn()
+      onChangeSpy = jest.fn()
+
+      wrapper = render(
+        <TextInput
+          autoFocus
+          className="custom-class"
+          data-arbitrary="arbitrary"
+          endAdornment={<span>end</span>}
+          value="value"
+          name="name"
+          onBlur={onBlurSpy}
+          onChange={onChangeSpy}
+          id="id"
+          label="label"
+          startAdornment={<span data-testid={adornmentId}>start</span>}
+          type="text"
+        />
+      )
+    })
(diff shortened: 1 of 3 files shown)


packages/react-component-library/src/components/TextInput/TextInput.tsx
```diff
@@ -98,6 +98,14 @@ const Component = React.forwardRef<HTMLInputElement, TextInputProps>(
   ) => {
     const { hasFocus, onLocalBlur, onLocalFocus } = useFocus(onBlur)
 
+    const inputRef = useRef<HTMLInputElement>(null)
+
+    useImperativeHandle(ref, () => inputRef.current!)
```
m7kvqbe1: Another technique used elsewhere is to use mergeRefs:

https://github.com/Royal-Navy/design-system/blob/eb9b42d198c8ac604528dd2b1139cfb2e62f90c6/packages/react-component-library/src/components/CheckboxRadioBase/CheckboxRadioBase.tsx#L128

This way we don't need to decorate the optional forwarded ref.

packages/react-component-library/src/components/TextInput/TextInput.test.tsx
```diff
@@ -9,6 +9,40 @@ import { TextInput } from '.'
...
+            value="value"
+            name="name"
+            onBlur={onBlurSpy}
+            onChange={onChangeSpy}
+            id="id"
+            label="label"
+            startAdornment={<span data-testid={adornmentId}>start</span>}
+            type="text"
+          />
+          <input type="text" data-testid="next-input" />
```
m7kvqbe1: Is this second input redundant?

m7kvqbe1 (commented):
Looks good I wonder if mergeRefs is more appropriate in this instance?

packages/react-component-library/package.json
```diff
@@ -170,6 +170,7 @@
     "focus-trap": "^6.9.0",
     "focus-trap-react": "^8.11.0",
     "lodash": "^4.17.21",
+    "merge-refs": "^1.3.0",
```
m7kvqbe1: This already exists I think - see `react-merge-refs` in deps.

We should just use the one already there or replace everywhere if this ones better.

m7kvqbe1 (approved):

PR #3801, opened 2024-05-16
feat(DataGrid): Update styling to latest

### LINKED ISSUE #3789: Update DataGrid styling to latest
https://ryk4ho.axshare.com/?id=d4vlwh&p=dash_links_and_culmative_sprints&sc=3

Screenshot 2024-05-15 at 13 05 32

m7kvqbe1:
## Related issue

Closes #3789 

## Overview

Update DataGrid styling in line with new designs and fix Kitchen Sink story.

## Reason

DataGrid is evolving to become a more fully fledged component.

## Work carried out

- [x] Fix Kitchen Sink story (missing sorting)
- [x] Update styling in line with new designs


packages/react-component-library/src/components/DataGrid/DataGrid.stories.tsx
```diff
@@ -19,6 +22,7 @@ const generateRandomData = (length: number): Order[] => {
       id: i + 1,
       productName: faker.commerce.productName(),
       quantity: faker.number.int({ min: 1, max: 100 }),
+      price: `£${faker.commerce.price()}`,
```
markhigham: can we seed the random number generator so that it always returns the same "random" numbers. Then we won't fail visual regression every time?

markhigham (approved):

packages/react-component-library/src/components/DataGrid/DataGrid.stories.tsx
m7kvqbe1: Ah yes, I forgot about this. I will have to strip faker out I think. All good i'll just take a static copy.

PR #3803, opened 2024-05-17
chore(Tokens): Export individual tokens

### LINKED ISSUE #3797: Export selector functions directly in @royalnavy/design-tokens
### Description

Currently to use the selectors in the Design Tokens library you have to import `selectors` and destructure it:

```ts
import { selectors } from '@royalnavy/design-tokens'

const { color } = selectors
```

I don't know the background to why it was done in this way, but it'd be helpful in general for it to be a bit more ergonomic to use the tokens.

One change towards that would be to make it possible to import the selector functions directly:

```ts
import { color } from '@royalnavy/design-tokens'
```

(Having to destructure `selectors` is a bit tedious, and also stops IDEs from being able to suggest missing imports.)

markhigham:
## Related issue

Closes #3797

## Overview

Export individual token members from design-tokens

## Link to preview

Usage before
```
import { selectors } from '@royalnavy/design-tokens'

const { color } = selectors
```

Usage after
```
import { color } from '@royalnavy/design-tokens'
```

## Work carried out

- [x] Destructured and exported individual tokens
- [x] Update to tokens stories using the new syntax

### DIFF
--- a/packages/design-tokens/src/index.ts
+++ b/packages/design-tokens/src/index.ts
@@ -1,9 +1,35 @@
+import selectors from './selectors'
+
 export * from './types'
 export * from './flat'
 
 export { getTheme } from './getters'
 
-export { default as selectors } from './selectors'
+export { selectors }
 
 export { default as lightTheme } from './themes/light'
 export { default as darkTheme } from './themes/dark'
+
+const {
+  animation,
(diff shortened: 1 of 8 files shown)


m7kvqbe1 (approved):
LGTM

PR #3809, opened 2024-05-20
docs(Button): Add single page documenting all button variants

markhigham:
## Overview

Add a single mdx page with all variants of buttons in place. It makes it easier to compare future changes to button styling

## Work carried out

**Example.**

- [x] New layout page for buttons 
- [x] Custom loading component

## Screenshot
image

### DIFF
--- a/packages/react-component-library/src/components/Button/LoadingButtons.tsx
+++ b/packages/react-component-library/src/components/Button/LoadingButtons.tsx
@@ -1,0 +1,73 @@
+import React, { useEffect, useState } from 'react'
+
+import { Button } from './Button'
+import { BUTTON_VARIANT } from './constants'
+import { COMPONENT_SIZE, ComponentSizeType } from '../Forms'
+
+const allVariants = [
+  BUTTON_VARIANT.PRIMARY,
+  BUTTON_VARIANT.SECONDARY,
+  BUTTON_VARIANT.TERTIARY,
+  BUTTON_VARIANT.DANGER,
+]
+
+export const LoadingButtons = ({
+  size = COMPONENT_SIZE.FORMS,
+}: {
+  size: ComponentSizeType
+}) => {
+  const [isLoading, setIsLoading] = useState(allVariants)
+
+  useEffect(() => {
+    const handleInterval = () => {
+      if (!isLoading.length) {
+        setIsLoading(allVariants)
+        return
+      }
+
+      setIsLoading((current) => {
+        return current.slice(1)
+      })
+    }
+
+    const interval = window.setInterval(handleInterval, 1000)
+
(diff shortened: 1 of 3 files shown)


packages/react-component-library/src/components/Button/Button.mdx
```diff
@@ -0,0 +1,122 @@
+import { Meta } from '@storybook/blocks'
```
m7kvqbe1: I love this change in principle but wonder if there is a way to do this without losing the 'props table' and 'show code' features from the dynamically generated docs page.

Maybe we could replace the existing Stories with these? 

Not only are these more succinct, they cover more permutations.

packages/react-component-library/src/components/Button/LoadingButtons.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { BUTTON_VARIANT } from './constants'
+
+import { COMPONENT_SIZE, ComponentSizeType } from '../Forms'
+
+export const LoadingButtons = ({
+  size = COMPONENT_SIZE.FORMS,
+}: {
+  size: ComponentSizeType
+}) => {
+  const allVariants = useMemo(
```
m7kvqbe1: Maybe just define as a const[] outside of the LoadingButtons declaration.

Then there is no need to memoize (which comes with its own overhead).

packages/react-component-library/src/components/Button/LoadingButtons.tsx
```diff
@@ -0,0 +1,76 @@
+import React, { useEffect, useMemo, useState } from 'react'
+
+import { Button } from './Button'
+import { BUTTON_VARIANT } from './constants'
+
```
m7kvqbe1: Redundant line break.

packages/react-component-library/src/components/Button/LoadingButtons.tsx
```diff
@@ -0,0 +1,76 @@
...
+      BUTTON_VARIANT.PRIMARY,
+      BUTTON_VARIANT.SECONDARY,
+      BUTTON_VARIANT.TERTIARY,
+      BUTTON_VARIANT.DANGER,
+    ],
+    []
+  )
+
+  const [isLoading, setIsLoading] = useState(allVariants)
+  useEffect(() => {
```
m7kvqbe1, line 21: Line break.

packages/react-component-library/src/components/Button/Button.mdx
markhigham: Good point. I've renamed the mdx and it's Meta data to make it appear separately.

m7kvqbe1 (approved):

Issue #3805, opened 2024-05-18
Masthead search box appears below Sidebar


markhigham:
### Steps to reproduce

Add the following components to a page.
- Masthead
- SideNav

Enable the search bar for masthead `onSearch={...}`

### Screenshot
![2024-05-18 08 16 10](https://github.com/Royal-Navy/design-system/assets/2064710/b9752271-258d-4592-9be8-b20344ef8120)

PR #3815, opened 2024-05-22
fix(SearchBar): Fixed searchbar obscuring text

will-lamb:
## Related issue

[#3805](https://github.com/Royal-Navy/design-system/issues/3805)

## Overview

Changed the Z-index value of the masthead in the StyledSearchBar component so that the search bar will be infront of the sidebar

## Reason

Currently, when opening the search bar, the sidebar is obscuring the text being typed into the search bar. 

## Work carried out

**Example.**

- [x] Changed z-index value in the StyledSearchBar component to `z-index: ${zIndex('masthead', 1)};` 

## Screenshot

image

## Developer notes

I have left the story in that I used to view both the masthead and sidebar at the same time but this will need to be removed before merging.

### DIFF
--- a/packages/react-component-library/src/components/TopLevelNavigation/SearchBar/partials/StyledSearchBar.tsx
+++ b/packages/react-component-library/src/components/TopLevelNavigation/SearchBar/partials/StyledSearchBar.tsx
@@ -6,9 +6,9 @@ interface StyledSearchBarProps {
 }
 
 export const StyledSearchBar = styled.div<StyledSearchBarProps>`
-  z-index: ${zIndex('dropdown', 1)};
+  z-index: ${zIndex('masthead', 1)};
   display: block;
   margin-top: -1px;
-  position: relative;
+  position: absolute;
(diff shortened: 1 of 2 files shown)


packages/react-component-library/src/components/TopLevelNavigation/SearchBar/partials/StyledSearchBar.tsx
```diff
@@ -9,6 +9,6 @@ export const StyledSearchBar = styled.div<StyledSearchBarProps>`
   z-index: ${zIndex('dropdown', 1)};
   display: block;
   margin-top: -1px;
-  position: absolute;
+  position: relative;
```
m7kvqbe1: I think the fix for this is possibly actually a z-index change:

`z-index: ${zIndex('masthead', 1)};`

Lets have a call to check.

m7kvqbe1 (approved):
LGTM just need to squash these commits on / before merge :)

PR #3820, opened 2024-05-29
docs(Button): Update story with example of preventing text wrapping

markhigham:
## Overview

The default behaviour for `DataGrid` is to wrap long text on to multiple lines. This PR includes an example story to keep text on a single line 

## Work carried out

[A list of work you have done, use markdown checklist format, if you leave any boxes unchecked, be sure to set this PR as a 

- [x] Update the Arbitrary content story with new styling to prevent text wrapping

## Screenshot
Before
image

After
image

### DIFF
--- a/packages/react-component-library/src/components/DataGrid/DataGrid.stories.tsx
+++ b/packages/react-component-library/src/components/DataGrid/DataGrid.stories.tsx
@@ -2,7 +2,7 @@ import React from 'react'
 import styled from 'styled-components'
 import { spacing } from '@royalnavy/design-tokens'
 import { fn } from '@storybook/test'
-import { StoryFn, Meta } from '@storybook/react'
+import { Meta, StoryFn } from '@storybook/react'
 import { ColumnDef } from '@tanstack/react-table'
 
 import { storyAccessibilityConfig } from '../../a11y/storyAccessibilityConfig'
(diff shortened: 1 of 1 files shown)


packages/react-component-library/src/components/DataGrid/DataGrid.stories.tsx
```diff
@@ -247,13 +246,30 @@ Sorting.parameters = {
...
+`
+
 const columnsWithArbitraryContent: ColumnDef<object>[] = columns.map(
   (item) => ({
     ...item,
     cell: (info) => {
       const value = info.getValue() as string
 
-      return <Badge>{value}</Badge>
+      return <StyledCell>{value}</StyledCell>
```
m7kvqbe1, line 282: This has removed the Badge so it's no longer visually apparent there is arbitrary cell content:

Screenshot 2024-05-30 at 08 33 25

Should this styling maybe be applied via a boolean column meta prop (`noWrap`?) and have a distinct story?

See example here used for alignment:

https://github.com/Royal-Navy/design-system/blob/9d5292ed941b3d8671f4d58d6b9dc724c14831d4/packages/react-component-library/src/components/DataGrid/DataGrid.d.ts#L6

m7kvqbe1 (approved):

PR #3837, opened 2024-07-04
fix(Sidebar): Sidebar now overlays page content

will-lamb:
## Related issue

#[3826](https://github.com/Royal-Navy/design-system/issues/3826)

## Overview

I have changed the display property of the sidebar to absolute when it is open so that it overlays the page content rather than pushing it to one side. 

## Reason

Currently, there is a layout shift happening when the sidebar is opened as it pushes the content to the right. 

## Work carried ou

**Example.**

- [x] changed the display property of the sidebar to absolute when it is open

## Screenshot

Before:
image

After:
image


m7kvqbe1:
Looks like there may be a bit more to this as it's causing a layout shift. It might be the way the example is setup.

m7kvqbe1 (approved):

PR #3877, opened 2024-09-16
fix(Alert): Prevent constained text with hideDismiss

m7kvqbe1:
## Related issue

#[issueid]

## Overview

Prevent text from being constrained when Dismiss button is disabled.

Also refactor responsive styling.

## Reason

The Alert styles were super old and weren't using flex box properly.

## Work carried out

- [x] Refactor styles
- [x] Fix constrained text


markhigham (approved):
Are you sure you want the icon above the title at mobile breakpoint?

m7kvqbe1:
> Are you sure you want the icon above the title at mobile breakpoint?

I figured it was better as the horizontal viewport is what's constrained as opposed to vertical.

PR #3882, opened 2024-09-17
feat(Masthead): Add slot to display client content in the masthead

markhigham:
## Related issue

#[issueid]

## Overview

Add an optional prop to masthead which allows the insertion of custom client jsx. 

## Reason

Vertical screen space is at a premium, especially on MODNET laptops. Sometimes it is a small as 570 available pixels.
This slot was added so that things like DatePickers can be inlined in the masthead to save space.

## Work carried out

- [x] New prop and slot in the masthead
- [x] Updated stories and tests

### DIFF
--- a/packages/react-component-library/src/components/TopLevelNavigation/Masthead/Masthead.test.tsx
+++ b/packages/react-component-library/src/components/TopLevelNavigation/Masthead/Masthead.test.tsx
@@ -129,6 +129,15 @@ describe('Masthead', () => {
     })
   })
 
+  describe('custom client component', () => {
+    it('should render custom client component', () => {
+      wrapper = render(
+        <Masthead {...props} customClientComponent={<div>Hello world</div>} />
+      )
+      expect(wrapper.getByText('Hello world')).toBeInTheDocument()
+    })
+  })
+
   describe('inline nav', () => {
     it('should render nav inline', () => {
       wrapper = render(<Masthead {...props} hasInlineNav />)
--- a/packages/react-component-library/src/components/TopLevelNavigation/Masthead/Masthead.tsx
+++ b/packages/react-component-library/src/components/TopLevelNavigation/Masthead/Masthead.tsx
@@ -76,6 +76,10 @@ export interface MastheadProps {
    * Optional jsx to render the classification bar above the masthead.
    */
   classificationBar?: React.ReactElement<ClassificationProps>
+  /**
+   * Optional jsx to insert after the before the icons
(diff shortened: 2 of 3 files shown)


packages/react-component-library/src/components/TopLevelNavigation/Masthead/Masthead.test.tsx
```diff
@@ -129,6 +129,15 @@ describe('Masthead', () => {
     })
   })
 
+  describe('custom client component', () => {
+    it('should render custom client component', () => {
+      wrapper = render(
+        <Masthead {...props} customClientComponent={<div>Hello world</div>} />
```
m7kvqbe1, line 135: Can we change the name of this prop to `rightSlot`?

packages/react-component-library/src/components/TopLevelNavigation/Masthead/Masthead.test.tsx
```diff
@@ -129,6 +129,15 @@ describe('Masthead', () => {
...
+      )
+      expect(wrapper.getByText('Hello world')).toBeInTheDocument()
```
m7kvqbe1, line 137: 😉 

```suggestion
      expect(wrapper.getByText('Hello, World!')).toBeInTheDocument()
```

packages/react-component-library/src/components/TopLevelNavigation/Masthead/Masthead.tsx
```diff
@@ -150,7 +155,7 @@ export const Masthead = ({
             {nav}
           </StyledInlineNav>
         ) : null}
-
+        {!!customClientComponent && customClientComponent}
```
m7kvqbe1, line 158:
```suggestion
        {customClientComponent}
```

m7kvqbe1 (commented):
Does it need some styling to stop it sitting flush?

Screenshot 2024-09-17 at 14 04 49

packages/react-component-library/src/components/TopLevelNavigation/Masthead/Masthead.test.tsx
markhigham: I don't want to surround the slot with styling. It's an "advanced" feature so the end user should take responsibility for how it works?
There may also be an edge case where they want it flush

markhigham:
Closing this PR as it is replaced by https://github.com/Royal-Navy/design-system/pull/3888

PR #3888, opened 2024-09-17
feat(Masthead): Add slot for custom client jsx

markhigham:
## Related issue

#[issueid]

## Overview

Add an optional prop to masthead which allows the insertion of custom client jsx. 

## Reason

Vertical screen space is at a premium, especially on MODNET laptops. Sometimes it is a small as 570 available pixels.
This slot was added so that things like DatePickers can be inlined in the masthead to save space.

## Work carried out

- [x] New prop and slot in the masthead
- [x] Updated stories and tests

### DIFF
--- a/packages/react-component-library/src/components/TopLevelNavigation/Masthead/Masthead.stories.tsx
+++ b/packages/react-component-library/src/components/TopLevelNavigation/Masthead/Masthead.stories.tsx
@@ -8,6 +8,7 @@ import {
   IconSettings,
   IconHome,
 } from '@royalnavy/icon-library'
+import { spacing } from '@royalnavy/design-tokens'
 
 import { Link } from '../../Link'
 import {
@@ -20,6 +21,8 @@ import {
 import { Notification, Notifications } from '../NotificationPanel'
 import { MASTHEAD_SUBCOMPONENT } from './constants'
 import { ClassificationBar } from '../../ClassificationBar'
+
+import { TextE } from '../../Text'
 
 const StyledContainer = styled.div`
   min-height: 10rem;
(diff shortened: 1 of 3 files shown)


packages/react-component-library/src/components/TopLevelNavigation/Masthead/Masthead.tsx
```diff
@@ -76,6 +76,10 @@ export interface MastheadProps {
    * Optional jsx to render the classification bar above the masthead.
    */
   classificationBar?: React.ReactElement<ClassificationProps>
+  /**
+   * Optional jsx to insert after the before the icons
```
m7kvqbe1: Is this description a typo?

packages/react-component-library/src/components/TopLevelNavigation/Masthead/Masthead.stories.tsx
```diff
@@ -322,3 +325,21 @@ WithClassificationBar.args = {
...
+
+export const WithCustomClientComponent = Default.bind({})
+WithCustomClientComponent.args = {
+  ...Default.args,
+  rightSlot: (
+    <StyledClientComponent>
+      <TextE>Arbitrary text</TextE>
+    </StyledClientComponent>
+  ),
+}
```
m7kvqbe1, line 346: Change to `export const RightSlot` and add the following so the story title case is correct:

```
RightSlot.storyName = 'Right slot'
```

packages/react-component-library/src/components/TopLevelNavigation/Masthead/Masthead.tsx
markhigham: makes no sense at all

m7kvqbe1 (approved):

PR #3892, opened 2024-09-24
feat(Dialog): Handle async action by disabling buttons and showing spinner

m7kvqbe1:
## Overview
Makes the `Dialog` handle async actions so that the buttons are disabled and a spinner shows if a promise has not yet resolved.

## Reason
Required in downstream application so that the user knows background work is in progress and doesn't double submit the action.

## Work carried out
- [x] Handle async actions
- [x] Refactor `Dialog` tests
- [x] Address `TextInput` `autoComplete` tech debt (downstream app is using `@ts-expect-error`)

## Test instructions
Browse to http://localhost:6006/?path=/story/components-dialog--async-action

### DIFF
--- a/packages/react-component-library/src/components/Dialog/Dialog.stories.tsx
+++ b/packages/react-component-library/src/components/Dialog/Dialog.stories.tsx
@@ -65,10 +65,10 @@ const Example = (props: DialogProps) => {
     <StyledWrapper>
       <Button onClick={() => ref?.current?.open()}>Open Dialog</Button>
       <Dialog
-        {...props}
         ref={ref}
         onCancel={dismissDialog}
         onConfirm={dismissDialog}
+        {...props}
       />
     </StyledWrapper>
   )
@@ -98,4 +98,15 @@ RichDescription.args = {
       Support Arbitrary JSX for <strong>rich</strong> description text.
     </div>
   ),
+}
+
+export const AsyncAction = Template.bind({})
+AsyncAction.storyName = 'Async loading'
+AsyncAction.args = {
+  title: 'Example Title',
+  description: 'Example description',
+  onConfirm: () =>
+    new Promise((resolve) => {
+      setTimeout(resolve, 2000)
+    }),
 }
--- a/packages/react-component-library/src/components/Dialog/Dialog.test.tsx
+++ b/packages/react-component-library/src/components/Dialog/Dialog.test.tsx
@@ -1,6 +1,11 @@
 import React from 'react'
 
-import { render, RenderResult, fireEvent } from '@testing-library/react'
+import {
+  render,
+  RenderResult,
+  fireEvent,
+  waitFor,
+} from '@testing-library/react'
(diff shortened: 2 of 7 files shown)


markhigham (approved):

packages/react-component-library/src/components/Dialog/Dialog.tsx
```diff
@@ -53,16 +55,24 @@ export const Dialog = React.forwardRef<ModalImperativeHandle, DialogProps>(
...
     ref
   ) => {
+    const [isLoading, setIsLoading] = useState(false)
+
     const confirmButton: ButtonProps = {
-      onClick: onConfirm,
+      onClick: (event) => {
+        setIsLoading(true)
+
+        Promise.resolve(onConfirm?.(event)).finally(() => setIsLoading(false))
```
m7kvqbe1: I am curious what made you go for this over `async` `await`?

packages/react-component-library/src/components/Dialog/Dialog.test.tsx
```diff
@@ -154,4 +159,53 @@ describe('Dialog', () => {
...
+      fireEvent(
+        wrapper.getByTestId('modal-primary'),
+        new MouseEvent('click', {
+          bubbles: true,
+          cancelable: true,
+        })
+      )
+    })
+
+    it('renders the buttons', async () => {
```
m7kvqbe1: Maybe a slightly more descriptive name for test?

packages/react-component-library/src/components/Dialog/Dialog.stories.tsx
```diff
@@ -99,3 +99,14 @@ RichDescription.args = {
     </div>
   ),
 }
+
+export const AsyncAction = Template.bind({})
+AsyncAction.storyName = 'Async action'
```
m7kvqbe1: Maybe something a bit more descriptive like 'Async loading'?

packages/react-component-library/src/components/Dialog/Dialog.test.tsx
```diff
@@ -1,211 +1,142 @@
...
-  let onConfirm: (event: React.FormEvent<HTMLButtonElement>) => void
-  let onCancel: (event: React.FormEvent<HTMLButtonElement>) => void
-
-  beforeEach(() => {
-    title = 'Dialog Title'
-    description = <span>Dialog description.</span>
-    onConfirm = jest.fn()
-    onCancel = jest.fn()
-  })
+function setup({
```
m7kvqbe1, line 9: So much better :tada:

m7kvqbe1 (commented):
Looks great, just a couple of comments.

It says I can't approve my own PR.

packages/react-component-library/src/components/Dialog/Dialog.tsx
thyhjwb6: It was a response to Sonarcloud but it's clear now it can be solved in a better way.

image

PR #3893, opened 2024-09-26
fix(Drawer): Handle scrolling content

markhigham:
## Related issue

#[issueid]

When you have a large list within the drawer the content will not scroll as expected. This is because the outer container doesn't have a 100% height.

## Reason

Scrolling drawer content *may* be needed for the NLIMS dashboard.

## Work carried out

- [x] Update the styled component to have 100% height
- [x] Add a new story with lorem ipsum text to demonstrate scrolling


m7kvqbe1 (commented):
Can we get an automated test around this?

PR #3906, opened 2024-10-04
fix(Toast): Use unique IDs when multiple toasts shown

jpveooys:
## Overview

This fixes a bug where duplicate IDs were used for `aria-labelledby` and `aria-describedby` (and the related elements) when multiple toasts were on screen.

It also removes the conditional setting of `aria-labelledby` based on whether there was a message, as that didn't make much sense (there is always some text in the label element, and it doesn't depend on whether there's a message).

## Link to preview

https://deploy-preview-3906--storybook-navy-digital-mod-uk.netlify.app/

## Reason

Duplicate IDs were causing toasts to have the wrong accessible names etc when multiple toasts were visible.

## Work carried out

- [x] Append toast ID to element IDs
- [x] Remove conditional setting of `aria-labelledby`
- [x] Update tests

## Screenshot

Previous duplicate IDs can be seem on https://storybook.design-system.navy.digital.mod.uk/?path=/docs/components-toast--docs by creating multiple toasts:

image

With the changes in, the IDs are unique:

image

### DIFF
--- a/packages/react-component-library/src/components/Toast/Toast.test.tsx
+++ b/packages/react-component-library/src/components/Toast/Toast.test.tsx
@@ -94,3 +94,30 @@ it('sets the message when the message is JSX', async () => {
   expect(within(lastToast).getByText(expectedNewLabel)).toBeInTheDocument()
   expect(within(lastToast).getAllByRole('paragraph')).toHaveLength(2)
 })
+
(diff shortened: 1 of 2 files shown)


m7kvqbe1 (approved):
Nice fix! I have a feeling this will sort out an issue @thyhjwb6 was having with RTL matchers finding multiples of the same Toast recently.

PR #3945, opened 2024-10-14
fix(TabNav): Make full TabNav tab clickable

### LINKED ISSUE #3891: Masthead/TabNav: Tab click area does not cover whole tab
### Steps to reproduce

1. Go to https://storybook.design-system.navy.digital.mod.uk/?path=/story/components-masthead--default
2. Notice that clicking on the text in a tab navigates to the link as expected, but clicking on the rest of the tab does not. This is despite the same mouse cursor/pointer being used for the whole tab.

### Demo

The difference can easily be seen in the browser link preview at the bottom of the browser window (plus the hover effect difference as well):

https://github.com/user-attachments/assets/6b2a4755-0123-4b02-af47-b1375268d066

### Additional info

On another project I think `&::after` was used to extend the click area in a similar scenario, but I'm not sure how easily that can be applied in this context.

m7kvqbe1:
## Related issue

Closes #3891

## Overview

Ensure that the entire tab area within TabNav is a clickable anchor.

## Reason

Previously a user would have to click on the anchor text to register a click.

## Work carried out

- [x] Ensure anchor takes full space of Tab
- [x] Supplement TabNav with additional automated tests
- [x] Enable skipped Playwright test within Masthead suite
- [x] Improve display of Masthead stories within docs tab

## Screenshot

Screenshot 2024-10-14 at 13 08 16

## Developer notes

There is a trade off here - _I think_ we've improved the a11y however i've also had to disable a rule 😬.

Let me know what you all think - alternative is to strip out the ARIA attr and `tablist, tab` roles.

### DIFF
--- a/packages/react-component-library/e2e/Masthead/default.spec.ts
+++ b/packages/react-component-library/e2e/Masthead/default.spec.ts
@@ -15,8 +15,7 @@ test.describe('Masthead', () => {
     await expect(page.locator(selectors.banner)).toBeVisible()
   })
 
-  // eslint-disable-next-line playwright/no-skipped-test
-  test.skip('shows the buttons', async ({ page }) => {
+  test('shows the buttons', async ({ page }) => {
     await expect(page.locator(selectors.buttons.search)).toBeVisible()
     await expect(page.locator(selectors.buttons.notifications)).toBeVisible()
     await expect(page.locator(selectors.buttons.user)).toBeVisible()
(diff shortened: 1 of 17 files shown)


markhigham (approved):
It's a dilemma about the a11y changes from from a UX / Fitt's law pov this is great

jpveooys:
Yep not sold on `tablist`, as it sounds like there should be `tabpanel`s as well then: https://w3c.github.io/aria/#tablist

The `<ol>`, `<li>` and `<a>` elements were all inside a `<nav>` so I think the roles were okay before.

PR #3946, opened 2024-10-14
fix(Select): Ensure Select menu listbox toggles correctly

### LINKED ISSUE #3905: Select stops expanding after repeated clicks in label area (not the button)
### Steps to reproduce

1. Go to https://storybook.design-system.navy.digital.mod.uk/?path=/story/components-select--default
2. Click three times in either the empty white space in the middle, or on the label text

The Select doesn't expand on the third click and any subsequent clicks in the same area

#### No value selected

https://github.com/user-attachments/assets/fd3159bd-1861-4f9c-8dfe-7d7ae3b4855e

#### Value selected

https://github.com/user-attachments/assets/98d3139c-67b4-4a6c-a8e7-4026ccd3a26e

m7kvqbe1:
## Related issue

Closes #3905

## Overview

Ensure Select menu listbox toggles open and closed correctly upon interaction.

## Reason

It was previously only possible to toggle the listbox open/closed once via click.

You would have to trigger a blur and refocus to be able to open it again.

## Work carried out

- [x] Add noop handler with `preventDefault` for `onMouseDown`
- [x] Supplement with additional automated tests

## Developer notes

This was a nightmare to fix.

### DIFF
--- a/packages/react-component-library/src/components/Select/Select.tsx
+++ b/packages/react-component-library/src/components/Select/Select.tsx
@@ -72,8 +72,16 @@ export const Select = ({
       hasSelectedItem={!isNil(selectedItem)}
       id={id}
       inputProps={{
-        onFocus: onInputFocusHandler,
+        ...getToggleButtonProps({
+          onFocus: onInputFocusHandler,
+          onMouseDown: (event) => {
+            // This is crucial for correct focus and click behavior
+            // Without this, Downshift internal state gets mixed up
+            event.preventDefault()
+          },
+        }),
         ref: inputRef,
+        'aria-expanded': undefined,
       }}
       isOpen={isOpen}
       menuProps={getMenuProps({
--- a/packages/react-component-library/src/components/Select/Select.test.tsx
+++ b/packages/react-component-library/src/components/Select/Select.test.tsx
@@ -1,6 +1,6 @@
 import React from 'react'
 
-import { render, RenderResult } from '@testing-library/react'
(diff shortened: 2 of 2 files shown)


packages/react-component-library/src/components/Select/Select.tsx
```diff
@@ -72,8 +72,16 @@ export const Select = ({
...
+        ...getToggleButtonProps({
+          onFocus: onInputFocusHandler,
+          onMouseDown: (event) => {
+            // This is crucial for correct focus and click behavior
+            // Without this, Downshift internal state gets mixed up
+            event.preventDefault()
+          },
+        }),
         ref: inputRef,
+        'aria-expanded': undefined,
```
m7kvqbe1, line 84: This gets applied by `getToggleButtonProps` so we need to unset it.

Screenshot 2024-10-14 at 16 48 32

jpveooys: Hmm yeah `getToggleButtonProps()` is being used on two elements, maybe that's not right but hunting down the Downshift v6 docs isn't straightforward :smile: (Plus the WAI-ARIA spec has changed since v6...)

jpveooys (approved):
Works well :+1:

packages/react-component-library/src/components/Select/Select.tsx
m7kvqbe1: Another one we need to upgrade at some point - both this and DatePicker are crying out for it.

PR #3955, opened 2024-11-04
feat(DatePicker): Upgrade react-day-picker to v8

### LINKED ISSUE #2871: DatePickerE: Focus date when picker opened

### LINKED ISSUE #2872: DatePickerE: Tabbing away from day in calendar should clear range highlight

### LINKED ISSUE #3528: DatePicker - Warning: findDOMNode is deprecated in StrictMode
### Steps to reproduce
1. Create a new application using the CRA template
2. Add `DatePicker` inside `React.StrictMode`
3. Start application
4. Open `DatePicker` sheet and view the console

### Screenshot
Screenshot 2022-11-25 at 16 05 05

m7kvqbe1:
## Related issue

Closes https://github.com/Royal-Navy/design-system/issues/2871
Closes https://github.com/Royal-Navy/design-system/issues/2872

## Overview

Upgrades react-day-picker to v8.

## Reason

We are reliant on an old and unsupported major version of the underlying library.

## Work carried out

- [x] Upgrade to `8.3.1`
- [x] Upgrade to `8.10.1`
- [x] Add ability to toggle 'Jump to today' feature
- [x] Fixes keyboard range selection blur issue
- [x] Focus selected date or current date on open
- [x] Backfill some tests

## Developer notes

- https://github.com/Royal-Navy/design-system/pull/3956

- Note the `aria-label` associated with picker day buttons was removed, leaving them without a long form accessible name e.g. `5th December (Thursday)`. I imagine they had a good reason for this and it works well with screen readers.

- Not sure why the lockfile change is so huge. I checked to see if I nuked it by accident but it seems not.

- Once this and the companion PR has been merged the ambition is to upgrade to `9.2.1`.

- This hasn't yet been resolved: https://github.com/Royal-Navy/design-system/issues/3528


packages/react-component-library/src/components/DatePicker/DatePicker.stories.tsx
```diff
@@ -86,3 +86,9 @@ RangeWithInitialValue.args = {
   initialStartDate: parseISO('2021-12-05'),
   initialEndDate: parseISO('2021-12-15'),
 }
+
+export const JumpToToday = Template.bind({})
```
m7kvqbe1, line 90: This feature previously existed but there was no dedicated story or prop to toggle it.

packages/react-component-library/src/components/DatePicker/utils/isDateDisabled.ts
```diff
@@ -5,11 +5,11 @@ import {
   isWithinInterval,
   startOfDay,
 } from 'date-fns'
-import { type Modifier } from 'react-day-picker'
+import { type Matcher } from 'react-day-picker'
 
 export const isDateDisabled = (
   date: Date,
-  disabledDays: Modifier | Modifier[]
+  disabledDays: Matcher | Matcher[] | undefined
```
jpveooys: `daysOfWeek` is now `dayOfWeek` in the matchers so the function (and the tests) should be reflecting that...

However, I'm not sure this function is needed at all now – there is [useActiveModifiers](https://daypicker.dev/v8/api/functions/useActiveModifiers) and also [isMatch](https://daypicker.dev/v8/api/functions/isMatch) in react-day-picker v8. ~(Should hopefully be able to use `useActiveModifiers` to work out if `today` is disabled.)~ Ah would probably have to be `isMatch` since the Jump to today button isn't inside the day picker context.

packages/react-component-library/package.json
```diff
@@ -178,8 +178,8 @@
     "lodash": "^4.17.21",
     "polished": "^4.0.3",
     "react-compound-slider": "^3.3.1",
-    "react-day-picker": "^7.4.8",
     "react-hot-toast": "^2.4.1",
+    "react-day-picker": "^8.10.1",
```
jpveooys, line 181: Looks like this has moved line by accident.

packages/react-component-library/src/components/DatePicker/DatePicker.tsx
```diff
@@ -339,45 +345,58 @@ export const DatePicker = ({
...
-              if (isDateDisabled(day, disabledDays)) {
-                return
-              }
+          <div>
+            {jumpToToday && (
+              <StyledJumpToToday
+                variant={BUTTON_VARIANT.TERTIARY}
+                size={COMPONENT_SIZE.SMALL}
+                aria-label="Jump to today"
+                aria-disabled="false"
```
jpveooys: Are these two props needed? (The label matches the button text, and `aria-disabled` value is hard-coded to false.)

jpveooys (commented):
Nice! I haven't reviewed the tests, but a few small comments on the implementation.

I also noticed that there's a regression with the 'Jump to today' feature – if a different month is currently active in the picker, clicking 'Jump to today' doesn't switch the view to the today month.

packages/react-component-library/src/components/DatePicker/DatePicker.tsx
```diff
@@ -245,15 +231,28 @@ export const DatePicker = ({
   )
 
   const modifiers = {
-    start: replaceInvalidDate(startDate),
-    end: replaceInvalidDate(endDate),
-    ...(today ? { today } : {}),
+    start: replaceInvalidDate(startDate) || false,
+    end: replaceInvalidDate(endDate) || false,
+  }
+  const modifiersClassNames = {
```
thyhjwb6: I wonder if this could be declared as a constant at the top just to remove some noise from this function.

packages/react-component-library/src/components/DatePicker/DatePicker.tsx
```diff
@@ -320,7 +319,9 @@ export const DatePicker = ({
               aria-owns={floatingBoxId}
               data-testid="datepicker-input-button"
               isDisabled={isDisabled}
-              onClick={toggleIsOpen}
```
thyhjwb6: I don't see the problem with this - is there a reason for the change?

m7kvqbe1:
>I also noticed that there's a regression with the 'Jump to today' feature – if a different month is currently active in the picker, clicking 'Jump to today' doesn't switch the view to the today month.

@jpveooys ah yeah ok - that regression is actually sorted by the next PR. I had to work backwards from there in the end.

Update: Resolved now within this PR.

packages/react-component-library/src/components/DatePicker/DatePicker.tsx
```diff
@@ -339,45 +340,56 @@ export const DatePicker = ({
...
         <FocusTrap focusTrapOptions={focusTrapOptions}>
-          <StyledDayPicker
-            firstDayOfWeek={1}
-            todayButton="Jump to today"
-            onTodayButtonClick={(day) => {
-              if (isDateDisabled(day, disabledDays)) {
-                return
-              }
+          <div>
+            <StyledJumpToToday
```
thyhjwb6: I think ultimately a lot of this needs separating out as there is quite a bit going on in `DatePicker.tsx`.

thyhjwb6 (commented):
Looks good

packages/react-component-library/src/components/DatePicker/DatePicker.tsx
m7kvqbe1: Ok cool I agree - we'll get the two that are in the queue through then I can open a separate PR for that.

packages/react-component-library/src/components/DatePicker/DatePicker.tsx
m7kvqbe1: Hmm no I don't think so, they were in NSS for some reason and I missed them. I'll remove.

packages/react-component-library/src/components/DatePicker/DatePicker.tsx
m7kvqbe1: No idea why that changed wasn't one of mine.

packages/react-component-library/src/components/DatePicker/utils/isDateDisabled.ts
m7kvqbe1: WIP adding test to supplement this.

m7kvqbe1: I've implemented a `isDateDisabled` function that uses `isMatch` and backfilled tests accordingly.

The `DayPickerProps['disabled']` type actually supports a few other things beyond `Matcher[]`.

packages/react-component-library/e2e/DatePicker/default.spec.ts
```diff
@@ -44,3 +44,46 @@ test.describe('DatePicker, default', () => {
     })
   })
 })
+
+test.describe('DatePicker, jump to today', () => {
```
m7kvqbe1, line 48: This was difficult to test with RTL for some reason.

packages/react-component-library/src/components/DatePicker/DatePicker.tsx
```diff
@@ -154,6 +154,25 @@ export interface DatePickerProps
...
+  date: Date,
+  disabledDays?: DayPickerProps['disabled']
+) => {
+  if (!disabledDays) {
+    return false
+  }
+
+  if (typeof disabledDays === 'function') {
+    return disabledDays(date)
+  }
```
jpveooys: `Matcher` already supports functions (and the `disabled` prop is typed as `Matcher | Matcher[] | undefined;`). 

E.g.:

```ts
> isMatch(new Date(), [() => true])
true
> isMatch(new Date(), [() => false])
false
```

m7kvqbe1: Oops yeah you're right my mistake.

m7kvqbe1: Looks like there was already an implementation of this elsewhere: `isDateInMatcher`.

[Refactored to utilise that.](https://github.com/Royal-Navy/design-system/pull/3955/commits/4ffe4a0191c2b226477752ec5001b2e937c9ee32)

packages/react-component-library/src/components/DatePicker/DatePicker.tsx
```diff
@@ -378,7 +359,7 @@ export const DatePicker = ({
                 variant={BUTTON_VARIANT.TERTIARY}
                 size={COMPONENT_SIZE.SMALL}
                 onClick={() => {
-                  if (isDateDisabled(today, disabledDays)) {
+                  if (isDateInMatcher(today, disabledDays)) {
```
jpveooys: Hmm might make sense to do `calculateDateValidity(today, disabledDays) !== DATE_VALIDITY.VALID` instead (not that it's likely that `today` is an invalid date, but keeps `calculateDateValidity` as a single way of working out if a date is valid).

packages/react-component-library/e2e/DatePicker/default.spec.ts
```diff
@@ -44,3 +44,46 @@ test.describe('DatePicker, default', () => {
...
+
+    const prevMonthButton = page.getByRole('button', {
+      name: /go to previous month/i,
+    })
+    await prevMonthButton.click()
+    await prevMonthButton.click()
+    await prevMonthButton.click()
+
+    const initialMonth = await page
+      .locator(selectors.monthHeading)
```
jpveooys, line 66: BTW (if not aware) Playwright [no longer encourages use of the locator method](https://playwright.dev/docs/locators) (originally it was all they had...)

jpveooys (approved):
:+1:

thyhjwb6 (approved):
👍

PR #3959, opened 2024-11-07
refactor(DatePicker): Break up, extract handlers, consolidate state

m7kvqbe1:
## Overview

Refactor DatePicker.

This is a boy scout step change, I'm not aiming for perfection (other things to move on to).

I'm also not fully convinced this is better. I feel like i've added more code and indirection.

## Reason

https://github.com/Royal-Navy/design-system/pull/3955#discussion_r1827789321

## Work carried out

- [x] Break up the `DatePicker` into `Input` and `FloatingCalendar` sub components
- [x] Extract handlers from JSX
- [x] Consolidate state within the reducer (`calendarTableVariant`, `isOpen`)
- [x] Expose reducer via context
- [x] Memoize where appropriate (note impact on speed of test suites)

## Developer notes

I'm using context to expose the reducer, which is naughty with mutable state but performance impact is limited:

1. State updates are naturally bounded to user interactions
2. Most state changes (like paging the calendar) require child components to re-render anyway
3. Context scope is limited to the relatively shallow DatePicker component tree only
4. Refactoring all of the state and introducing Zustand felt like overkill given the above

### DIFF
--- a/packages/react-component-library/src/components/DatePicker/FloatingCalendar.tsx
+++ b/packages/react-component-library/src/components/DatePicker/FloatingCalendar.tsx
@@ -1,0 +1,272 @@
+import {
+  addMonths,
+  setMonth as changeMonth,
+  setYear as changeYear,
+} from 'date-fns/fp'
+import { enGB } from 'date-fns/locale'
+import { format } from 'date-fns'
+import FocusTrap from 'focus-trap-react'
+import React, { memo, useCallback } from 'react'
+
+import { BUTTON_VARIANT } from '../Button'
+import {
+  CALENDAR_TABLE_VARIANT,
+  CalendarTableVariant,
+  DATE_VALIDITY,
+} from './constants'
+import { CalendarNavigation } from './CalendarNavigation'
+import { CalendarTable } from './CalendarTable'
+import { COMPONENT_SIZE } from '../Forms'
+import { DATEPICKER_ACTION } from './types'
+import { DatePickerProps } from './DatePicker'
+import { replaceInvalidDate } from './utils'
+import { StyledDayPicker } from './partials/StyledDayPicker'
+import { StyledFloatingBox } from './partials/StyledFloatingBox'
+import { StyledJumpToToday } from './partials/StyledJumpToToday'
+import { useDatePickerContext } from './useDatePickerContext'
+import { useFocusTrapOptions } from './useFocusTrapOptions'
+import { useProcessDayClick, calculateDateValidity } from './useProcessDayClick'
+import { useRangeHoverOrFocusDate } from './useRangeHoverOrFocusDate'
+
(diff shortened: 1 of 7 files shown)


packages/react-component-library/src/components/DatePicker/useDatePickerContext.tsx
```diff
@@ -116,10 +154,44 @@ export function useDatePickerReducer(
           endDate: endDate ?? null,
           currentMonth,
           datePickerFormat,
+          isOpen: initialIsOpen,
```
jpveooys: IIRC the resetting thing was mainly about when it's a controlled component, and the start date or end date (or date format) change externally (and so the new values need to be brought into the state).

So strictly speaking, I think it should preserve the current `isOpen` value.

packages/react-component-library/src/components/DatePicker/Input.tsx
```diff
@@ -0,0 +1,197 @@
+import { IconEvent } from '@royalnavy/icon-library'
+import React, { memo, useCallback } from 'react'
+
+import { CALENDAR_TABLE_VARIANT } from './constants'
+import { DATE_FORMAT } from '../../constants'
+import { DATEPICKER_ACTION } from './types'
+import { DatePickerProps } from './DatePicker'
+import { hasClass } from '../../helpers'
+import { InlineButton } from '../InlineButtons/InlineButton'
+import { isDateValid } from './utils/isDateValid'
```
jpveooys: Import can be shortened here:
```suggestion
import { isDateValid } from './utils'
```

packages/react-component-library/src/components/DatePicker/FloatingCalendar.tsx
```diff
@@ -0,0 +1,272 @@
...
+  CALENDAR_TABLE_VARIANT,
+  CalendarTableVariant,
+  DATE_VALIDITY,
+} from './constants'
+import { CalendarNavigation } from './CalendarNavigation'
+import { CalendarTable } from './CalendarTable'
+import { COMPONENT_SIZE } from '../Forms'
+import { DATEPICKER_ACTION } from './types'
+import { DatePickerProps } from './DatePicker'
+import { replaceInvalidDate } from './utils/isDateValid'
```
jpveooys: Import can be shortened here:
```suggestion
import { replaceInvalidDate } from './utils'
```

jpveooys (approved):
I think it looks okay generally. Would've been useful to split this into multiple commits though :smile: 

(Main negative I can see is probably that it's added a fair bit to the line count...)

thyhjwb6 (approved):
👍 

I think it looks good and also gets me thinking about custom hooks - whether 1-2 more could be introduced or whether the existing ones need rework/adding to, but this is far enough for now.

PR #3964, opened 2024-11-22
fix(DatePicker): Fix compatibility with strict mode

jpveooys:
## Related issue

#3528

## Overview

This:

- updates focus-trap and focus-trap-react to fix findDOMNode deprecation warnings from React
- makes DatePicker work in strict mode

## Reason

Strict mode should be supported. (Previously the day picker would not open when strict mode is on.)

## Work carried out

- [x] Update focus-trap and react-focus-trap
- [x] Use `clickOutsideDeactivates` and `escapeDeactivates` focus-trap callbacks instead of `onDeactivate`
- [x] Remove `DATEPICKER_ACTION.TOGGLE_OPEN` reducer action (as reducer functions are called twice in strict mode)

## Developer notes

See https://github.com/focus-trap/focus-trap-react?tab=readme-ov-file#%EF%B8%8F%EF%B8%8F-react-18-strict-mode-%EF%B8%8F%EF%B8%8F for the react-focus-trap node about strict mode – `onDeactivate` wasn't really the right callback as it gets called when the component is unmounted.

Strict mode is still disabled in Storybook as there were some other unrelated failures with it on (mainly in the form usage examples).


m7kvqbe1 (approved):
Looks good! Legendary stuff!

PR #3974, opened 2024-12-19
Fix `FloatingBox` merged refs being overwritten by `ref` prop in React 19

### LINKED ISSUE #3969: Incorrect DatePicker FloatingCalendar positioning with React 19
### Steps to reproduce

1. In `packages/react-component-library`, run `pnpm add react@latest react-dom@latest`

2. Start Storybook

3. Go to http://localhost:6006/iframe.html?globals=&args=&id=components-date-picker--default&viewMode=story

4. Click the DatePicker calendar button.

Observe that the calendar appears in the wrong place.

### Additional details

Next.js 15 requires the use of React 19 when using the app router.

The inline styles applied by react-popper are incorrect in this case. Simply replacing `react-popper` with `@floating-ui/react` didn't seem to fix the problem. However, the latter has a different API and a lot more functionality so switching to it may still be needed to fix the problem.

### Screenshot

image

thyhjwb6:
## Related issue
Fixes #3969

## Overview
Fixes an issue when using React 19 which adds the `ref` as a prop. This was overwriting the merged refs because it is `null`.

## Reason
The floating box for the `DatePicker` is positioned incorrectly.

## Work carried out
- [x] Align existing tests with patterns
- [x] Add a test to cover positioning
- [x] Fix the `ref` issue

## Test instructions
Follow the reproduction steps in the [original issue](https://github.com/Royal-Navy/design-system/issues/3969).

## Developer notes
A follow up to this (hopefully tomorrow) will be to replace [React Popper](https://github.com/floating-ui/react-popper) with [Floating UI](https://github.com/floating-ui/floating-ui).

### DIFF
--- a/packages/react-component-library/src/primitives/FloatingBox/FloatingBox.tsx
+++ b/packages/react-component-library/src/primitives/FloatingBox/FloatingBox.tsx
@@ -1,4 +1,4 @@
-import React, { useRef } from 'react'
+import React, { forwardRef, useRef } from 'react'
 import { BasePlacement, Placement, VariationPlacement } from '@popperjs/core'
 import { Transition } from 'react-transition-group'
 import mergeRefs from 'react-merge-refs'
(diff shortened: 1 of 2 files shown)


packages/react-component-library/src/primitives/FloatingBox/FloatingBox.tsx
```diff
@@ -52,79 +52,102 @@ export type FloatingBoxProps =
...
-            $transitionStatus={state}
-            style={styles.popper}
-            ref={mergeRefs([nodeRef, floatingElementRef])}
-            onMouseEnter={onMouseEnter}
-            onMouseLeave={onMouseLeave}
-            role={role}
-            data-testid="floating-box"
-            {...attributes.popper}
-            {...rest}
+    /*
```
m7kvqbe1: Sorry, this is really nitpicky but could we change the comment style to:
```
//
//
//

m7kvqbe1 (approved):
🎉

PR #3978, opened 2025-02-05
feat(Text): Deprecate TextE in favour of Text

markhigham:
## Related issue

DNADB-80. The creation of a new webapp repo which will leverage RNDS components

## Overview

Renamed references from TextE to Text in components, unit tests, stories, and docs
Added new TextE and TextEProps as aliased to Text / TextProps with deprecated flag

## Reason

The TextE component is in use in production systems.

## Work carried out

- [x] Update TextE to Text
- [x] Update TextProps to TextProps
- [x] Add deprecated TextE and TextEProps aliases

### DIFF
--- a/packages/react-component-library/src/components/Text/Text.stories.tsx
+++ b/packages/react-component-library/src/components/Text/Text.stories.tsx
@@ -2,12 +2,12 @@ import React from 'react'
 import styled, { css } from 'styled-components'
 
 import { Meta, StoryFn } from '@storybook/react'
-import { textAlignments, TextE } from './Text'
+import { textAlignments, Text } from './Text'
 import { allowedElements } from './textStyles'
 
 export default {
-  component: TextE,
-  title: 'Experimental/Text',
+  component: Text,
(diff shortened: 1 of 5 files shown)


packages/react-component-library/src/components/Text/Text.stories.tsx
```diff
@@ -2,12 +2,12 @@ import React from 'react'
...
 import { Meta, StoryFn } from '@storybook/react'
-import { textAlignments, TextE } from './Text'
+import { textAlignments, Text } from './Text'
 import { allowedElements } from './textStyles'
 
 export default {
-  component: TextE,
-  title: 'Experimental/Text',
+  component: Text,
+  title: 'Components/Text',
```
m7kvqbe1: Can we stick it under `Utility/Text` with `Stack` and `Group`?

m7kvqbe1 (commented):
Looks good just a comment about where it should sit in Storybook.

m7kvqbe1 (approved):

Issue #3902, opened 2024-10-01
Datagrid filters don't match when the data item is numeric 


markhigham:
Seems to be doing a string compare between strings and numbers

PR #3981, opened 2025-02-25
chore(Docs): Update columnDef filter for numeric columns

markhigham:
## Related issue

#3902 

## Overview

Add the `includesString` filterFn for the numeric Qty columnDef in stories and tests for DataGrid

## Reason

It looked like a bug with filtering and numeric columns but we just needed to let the underlying component know how to filter the data

## Work carried out

- [x] Add filterFn to Stories and Tests

### DIFF
--- a/packages/react-component-library/src/components/DataGrid/DataGrid.stories.tsx
+++ b/packages/react-component-library/src/components/DataGrid/DataGrid.stories.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useCallback } from 'react'
+import React, { useCallback, useState } from 'react'
 import styled from 'styled-components'
 import { spacing } from '@royalnavy/design-tokens'
 import { fn } from '@storybook/test'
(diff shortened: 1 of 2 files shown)


packages/react-component-library/src/components/DataGrid/DataGrid.test.tsx
```diff
@@ -1033,6 +1040,26 @@ describe('DataGrid', () => {
...
+      await hackyWaitFor()
+
+      const filterInput = screen.getByPlaceholderText('Filter Numeric')
+      userEvent.type(filterInput, '2')
+
+      await hackyWaitFor()
+
+      const rows = screen.getAllByRole('row')
+      expect(rows).toHaveLength(2)// Header row + 1 data row
+
```
m7kvqbe1, line 1060: Redundant newline.

packages/react-component-library/src/components/DataGrid/DataGrid.stories.tsx
```diff
@@ -175,6 +175,7 @@ const columns = [
     accessorKey: 'quantity',
     enableSorting: false,
     enableColumnFilter: false,
+    filterFn: 'includesString',
```
m7kvqbe1: `equalsString` - Case-insensitive string equality

m7kvqbe1 (approved):

Issue #3901, opened 2024-10-01
Pagination scrolls horizontall with the grid. It shouldn't


markhigham:
When using a scrolling container the pagination scrolls with the datagrid

It should be absolutely positioned

PR #3983, opened 2025-02-25
chore(DataGrid): Align pagination to left for DataGrid

markhigham:
## Related issue

#3901 

## Overview

Align the pagination for DataGrid to the left and avoid the issue with it being rendered off screen for full width grid

## Work carried out

- [x] Change alignment from flex-end to flex-start

## Screenshot
image

### DIFF
--- a/packages/react-component-library/src/components/DataGrid/partials/StyledPagination.tsx
+++ b/packages/react-component-library/src/components/DataGrid/partials/StyledPagination.tsx
@@ -4,6 +4,6 @@ import { spacing } from '@royalnavy/design-tokens'
 import { Pagination } from '../../Pagination'
 
 export const StyledPagination = styled(Pagination)`
-  align-self: flex-end;
+  align-self: flex-start;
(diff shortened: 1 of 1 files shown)


packages/react-component-library/src/components/DataGrid/partials/StyledPagination.tsx
```diff
@@ -4,6 +4,6 @@ import { spacing } from '@royalnavy/design-tokens'
 import { Pagination } from '../../Pagination'
 
 export const StyledPagination = styled(Pagination)`
-  align-self: flex-end;
+  align-self: flex-start;
   margin: ${spacing('8')} 0 0 ${spacing('8')};
```
m7kvqbe1: Does this margin need changing?

markhigham: I've aligned it left, with the grid

m7kvqbe1 (approved):
Can you change these commits messages to `fix(DataGrid): ...`

PR #3994, opened 2025-04-15
feat(DataGrid): Add support for manual (server-side) pagination

m7kvqbe1:
## Related issue

`DNADB-310`

## Overview

Added server-side pagination support to DataGrid component.

## Reason

Enable DataGrid to work with paginated API data using the same pattern as server-side sorting.

## Work carried out

- [x] Added `manualPagination` support
- [x] Fixed pagination calculation for server-side data
- [x] Consolidated pagination callbacks into single `onPaginationChange`
- [x] Added "ManualPagination" story and tests
- [x] Maintained backward compatibility
- [x] Add `isLoading` prop to display async loading state

## Developer notes

The component now uses a single callback pattern for pagination. 

- Set `manualPagination: true`
- Provide pagination state and `onPaginationChange` handler to fetch new data
- Specify `pageCount` for total pages


markhigham (approved):
LGTM

packages/react-component-library/src/components/DataGrid/DataGrid.tsx
```diff
@@ -177,12 +205,35 @@ export const DataGrid = <T extends object>(props: DataGridProps<T>) => {
...
   const handlePagination = useCallback(
     (...args) => {
-      const [_, currentPage] = args
-      table.setPageIndex(currentPage - 1)
-      // @ts-ignore
-      onPageChange?.(...args)
+      const [event, currentPage, totalPages] = args
+      const pageIndex = currentPage - 1
+
+      // For external pagination, use onPaginationChange
```
thyhjwb6: I sometimes wonder if we need comments then is the prop name right. Is `manualPagination` the right name?

packages/react-component-library/src/components/DataGrid/DataGrid.tsx
```diff
@@ -78,9 +82,29 @@ export interface DataGridPropsWithInternalSorting<T extends object>
   sorting?: never
 }
 
+export interface DataGridPropsWithExternalPagination<T extends object>
+  extends DataGridBaseProps<T>,
+    Pick<TableOptions<T>, 'onPaginationChange'> {
+  manualPagination: true
```
thyhjwb6, line 94: Do we need `manualPagination` or can we just know that the pagination is externally controlled if `onPaginationChange` is defined?

packages/react-component-library/src/components/DataGrid/Pagination.tsx
```diff
@@ -23,7 +23,7 @@ export const Pagination = ({
       name="datagrid-pagination"
       initialPage={pagination.pageIndex + 1}
       pageSize={pagination.pageSize}
-      total={dataLength || pageCount! * pageSize!}
+      total={dataLength !== undefined ? dataLength : pageCount * pageSize}
```
thyhjwb6: I think this can be simplified to `total={dataLength ?? pageCount * pageSize}`.

thyhjwb6 (commented):
Looks good, just a few comments mostly around the API, making sure we are happy.

packages/react-component-library/src/components/DataGrid/DataGrid.tsx
m7kvqbe1: Yeah so that prop is actually typed by the underlying library (we also have `manualSorting`).

I agree the comments about the logic are probably unnecessary - I'll remove them.

Screenshot 2025-04-16 at 08 39 23

m7kvqbe1: I'll probably just leave this one:

```
// Legacy support for onPageChange
```

packages/react-component-library/src/components/DataGrid/DataGrid.tsx
m7kvqbe1: I've changed it slightly so `onPaginationChange` gets invoked regardless of whether `manualPagination` is set.

This way even with internal pagination state there is still a handler to provide feedback.

This better aligns with the original stateful behaviour.

thyhjwb6 (approved):
Looks great 🎉

PR #3997, opened 2025-04-24
DNADB-369: Integrate `RowsPerPage` with `DataGrid` to change how many rows are displayed

thyhjwb6:
## Related issue
DNADB-369

## Overview
Adds the `Rows per page` `Select` to the bottom of the `DataGrid`.

## Reason
A useful way to alter the number of rows being displayed.

## Work carried out
- [x] Add `RowsPerPage` component
- [x] Integrate `RowsPerPage` with `DataGrid`
- [x] Fix issue in `Pagination` due to resetting the page index

## Screenshot
screencapture-localhost-6006-2025-04-25-14_24_40

## How to test
1. `pnpm run storybook`
2. Browse to http://localhost:6006/?path=/story/components-data-grid--rows-per-page


packages/react-component-library/src/components/DataGrid/DataGrid.test.tsx
```diff
@@ -71,12 +77,6 @@ describe('DataGrid', () => {
     },
   ]
 
-  it('should set the `role` attribute to `grid`', () => {
```
thyhjwb6, line 74: This test is redundant because it's implicit in the other uses of `screen.getByRole('grid')`.

packages/react-component-library/src/components/DataGrid/DataGrid.tsx
```diff
@@ -56,12 +56,14 @@ export interface DataGridBaseProps<T extends object>
   initialRowSelection?: RowSelectionState
   isFullWidth?: boolean
   hasHover?: boolean
+  hasRowsPerPage?: boolean
```
thyhjwb6: There's an issue with the interface here because if `hasRowsPerPage` is `true` then `pageSize` is not required. Normally we would solve this through union types and `never` but there is a fair amount of complexity already with `DataGrid` types. Another option could be that we always display the `RowsPerPage` and mark the `pageSize` prop as deprecated.

packages/react-component-library/src/components/RowsPerPage/RowsPerPage.tsx
```diff
@@ -0,0 +1,55 @@
+import React from 'react'
+
+import { Select, SelectOption } from '../Select'
+
+interface RowsPerPageProps {
+  /**
```
thyhjwb6, line 6: Ideally I would have reused the `SelectBaseProps` `interface`. I could have used `Omit` as we do not want the `hideClearButton`, `initialValue` or `label` props but then the documentation does not appear in Storybook. The alternative is that I reuse `SelectBaseProps` but we don't have the documentation in Storybook.

packages/react-component-library/src/components/DataGrid/DataGrid.tsx
m7kvqbe1: I think the deprecation makes sense and we should always display it when paginating.

packages/react-component-library/src/components/RowsPerPage/RowsPerPage.tsx
m7kvqbe1: Happy for you to keep it as is, agree it's fine given alternative.

m7kvqbe1 (commented):
Great work! 

Can we add this bit too?

Screenshot 2025-04-25 at 21 57 55

packages/react-component-library/src/components/DataGrid/DataGrid.tsx
m7kvqbe1: Can we check it plays nice with the external state pagination stories?

m7kvqbe1 (approved):

PR #4001, opened 2025-04-29
feat(TabSet): Add an optional hasOverflow prop

markhigham:
## Overview

Add a `hasOverflow` prop to the TabSet which allows content to overflow the container. Defaults to false

## Reason

When the hasOverflow property is set we can handle `Select` components within the content. Previously they would be constrained

## Work carried out

- [x] Add hasOverflow prop and drill through to StyledBody component
- [x] An example in Storybook

### DIFF
--- a/packages/react-component-library/src/components/TabSet/TabSet.stories.tsx
+++ b/packages/react-component-library/src/components/TabSet/TabSet.stories.tsx
@@ -1,7 +1,12 @@
 import React from 'react'
+import styled from 'styled-components'
+
 import { StoryFn as Story, Meta } from '@storybook/react'
 
 import { ScrollableTabSetProps, TabSetItem, TabSet, TabSetProps } from '.'
+import { Select, SelectOption } from '../Select'
+import { Stack } from '../Stack'
+import { Text } from '../Text'
 
 export default {
   component: TabSet,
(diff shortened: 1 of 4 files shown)


packages/react-component-library/src/components/TabSet/TabSet.stories.tsx
```diff
@@ -175,3 +179,31 @@ export const Scrollable: Story<ScrollableTabSetProps> = (props) => (
 )
 
 Scrollable.storyName = 'Scrollable paged tabs'
+
+const StyledWrapper = styled.div`
+    height: 400px;
+  padding:1rem;
```
m7kvqbe1: Indentation here.

packages/react-component-library/src/components/TabSet/TabSet.stories.tsx
```diff
@@ -175,3 +179,31 @@ export const Scrollable: Story<ScrollableTabSetProps> = (props) => (
...
+`
+
+export const HasOverflow: Story<TabSetProps> = (props) => (
+  <StyledWrapper>
+    <TabSet {...props}>
+      <TabSetItem title="Has Overflow">
+        <>
+          <Text>Sometimes we need to allow the content to overflow</Text>
```
m7kvqbe1:
```suggestion
          <Text>Sometimes we need to allow the content to overflow:</Text>
```

packages/react-component-library/src/components/TabSet/TabSet.stories.tsx
```diff
@@ -175,3 +179,31 @@ export const Scrollable: Story<ScrollableTabSetProps> = (props) => (
...
+          <br/>
```
m7kvqbe1: Can we avoid the `<br />` and stick it in a `<Stack />` please.

m7kvqbe1 (approved):

packages/react-component-library/src/components/TabSet/TabSet.test.tsx
```diff
@@ -363,4 +363,22 @@ describe('TabSet', () => {
...
+          <TabSetItem title="Title 1">Content 1</TabSetItem>
+          <TabSetItem title="Title 2">Content 2</TabSetItem>
+          <TabSetItem title="Title 3">Content 3</TabSetItem>
+        </TabSet>
+      )
+    })
+
+    it('should have overflow visible', () => {
+      expect(wrapper.getByTestId('tabset-styledbody')).toHaveStyleRule(
+        'overflow', 'visible',
```
thyhjwb6: Maybe test for the inverse eg `should not have overflow visible when hasOverflow is false`

thyhjwb6 (approved):
👍

PR #4009, opened 2025-05-19
fix(Sidebar): Adjust user menu popover position and layout

markhigham:
## Related issue

DNADB-393

## Overview

Tweak to the layout and position of the collapsed sidebar user avatar popup.

## Reason

The popup was being rendered over the sidebar

## Work carried out

- [x] Add margin to the Sheet component which is displayed when the collapsed user avatar is clicked
- [x] Align icon and text in the user popup with `Group` component

## Screenshot
Before

After

### DIFF
--- a/packages/react-component-library/src/components/TopLevelNavigation/Sidebar/partials/StyledUserItemIcon.tsx
+++ b/packages/react-component-library/src/components/TopLevelNavigation/Sidebar/partials/StyledUserItemIcon.tsx
@@ -1,11 +1,10 @@
 import styled from 'styled-components'
-import { color, spacing } from '@royalnavy/design-tokens'
+import { color } from '@royalnavy/design-tokens'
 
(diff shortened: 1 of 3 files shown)


packages/react-component-library/src/components/TopLevelNavigation/Sidebar/partials/StyledUserItemIcon.tsx
```diff
@@ -5,7 +5,6 @@ import { StyledUserItem } from './StyledUserItem'
 
 export const StyledUserItemIcon = styled.div`
   display: inline-flex;
-  margin-right: ${spacing('4')};
```
markhigham, line 8: We don't need a margin now that the icon and text are using a Group component

m7kvqbe1 (approved):
Nice one!

PR #4012, opened 2025-05-23
feat(Select): Implement vertical positioning for Select component

markhigham:
## Related issue

DNADB-413

## Overview

Add a positioning hook to the Select component which allows the dropdowns to appear above the text input. Defaults to below.

## Work carried out

- [x] Add new use `useDropDownDirection` hook 
- [x] Adjust styles to place the Options list above when requested


m7kvqbe1:
Something like this to make it automatic and then get rid of the public prop?

https://github.com/Royal-Navy/design-system/pull/4013

m7kvqbe1 (approved):

m7kvqbe1:
Can you squash the commits please.

PR #4015, opened 2025-06-12
fix(DatePicker): Add inline clear button

markhigham:
## Related issue

DNADB-533

## Overview

Add a clear button from `Select` to the `DatePicker`

## Reason

Missing feature.

## Work carried out

- [x] Move ClearButton
- [x] Add ClearButton to the DatePicker->Input component and hook up events
- [x] Add `$hideLeftBorder` prop to StyledInlineButtons container so we can reuse


m7kvqbe1 (approved):
LGTM

PR #4016, opened 2025-06-19
Reverse sorting arrows

ollie-sutton:
## Related issue

TDB

## Overview

Reverse the sorting arrows

## Reason

To ascend is to go up hence the arrow pointing up, and to descend is to go down hence the arrow pointing down.

A quick google this does seem to be a debate in the design community, then its an arrow and not just an arrow head the approach currently taken could be seen as valid.

The change is the PR does bring the table in line with how _SharePoint_ shows the arrows when ordering though 😂

## Work carried out

- [x] Reversed the sorting arrows


markhigham (approved):
👍 LGTM

m7kvqbe1 (approved):

PR #4022, opened 2025-09-26
feat(DataGrid): Expose underlying autoResetPageIndex prop

markhigham:
## Overview

Expose the `autoResetPageIndex` prop to `DataGrid`. 
This will allow clients to determine whether they want pagination to reset following an update to the DataGrid's data. 
Resetting the index is the default behaviour to help preserve the integrity of pages. This behaviour may cause unwanted resets if you are only updating individual data rows and not mutating the entire list.

## Reason
We needed to give the component user the option to override the grid defaults.

## Work carried out

- [x] use faker js to generate DataGrid stories row data.
- [x] Expose and drill down the autoResetPageIndex prop.
- [x] Extend DataGrid stories with a Date Updates example. Showing how page indexes can be reset or not.


m7kvqbe1 (approved):
Looks good!

Can you change this commit message please to `docs(DataGrid)`:

>fix(Stories): Use seeded random data for DataGrid

PR #4028, opened 2025-11-25
feat(FloatingUI): Migrate from popperjs to FloatingUI

ollie-sutton:
## Overview

Migrate from popperjs to FloatingUI

## Reason

FloatingUI is the replacement for popperjs and it provides some useful util hooks that can be used else where in the design-system (initially the Popover)

## Work carried out

- [x] Migrate from popperjs to FloatingUI
- [ ] Update Popover to use hooks provided by FloatingUI (This will come in a follow up PR as this might require a re-work of the Floating box)

## Developer notes

Seems like the functionality is the same between the components, maybe a few things are a pixel or two difference

### DIFF
--- a/packages/react-component-library/src/hooks/useFloatingElement.ts
+++ b/packages/react-component-library/src/hooks/useFloatingElement.ts
@@ -1,51 +1,120 @@
 import React, { Dispatch, SetStateAction } from 'react'
-import { PositioningStrategy, Placement, VirtualElement } from '@popperjs/core'
-import { usePopper } from 'react-popper'
+import {
+  Strategy,
+  Placement,
+  VirtualElement,
+  useFloating,
+  arrow,
+  autoUpdate,
+  autoPlacement,
+  flip,
+  shift,
+  UseFloatingOptions,
+} from '@floating-ui/react'
 
 import { useStatefulRef } from './useStatefulRef'
 
 export const useFloatingElement = (
-  placement: Placement = 'bottom',
-  strategy: PositioningStrategy = 'fixed',
+  placement: Placement | 'auto' = 'bottom',
+  strategy: Strategy = 'fixed',
   externalTargetElement: Element | VirtualElement | null = null
 ): {
   targetElementRef: Dispatch<SetStateAction<Element | null>>
   floatingElement: HTMLElement | null
   floatingElementRef: Dispatch<SetStateAction<HTMLElement | null>>
-  forceUpdate: ReturnType<typeof usePopper>['forceUpdate']
+  forceUpdate: () => void
   arrowElementRef: Dispatch<SetStateAction<HTMLElement | null>>
   styles: { [key: string]: React.CSSProperties }
-  attributes: { [key: string]: { [key: string]: string } | undefined }
+  attributes: {
+    [key: string]: { [key: string]: string | undefined } | undefined
+  }
+  placement: Placement
 } => {
   const [targetElement, targetElementRef] = useStatefulRef()
   const [floatingElement, floatingElementRef] = useStatefulRef<HTMLElement>()
   const [arrowElement, arrowElementRef] = useStatefulRef<HTMLElement>()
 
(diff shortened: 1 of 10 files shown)


packages/react-component-library/src/hooks/useFloatingElement.ts
```diff
@@ -1,51 +1,120 @@
...
+  shift,
+  UseFloatingOptions,
+} from '@floating-ui/react'
 
 import { useStatefulRef } from './useStatefulRef'
 
 export const useFloatingElement = (
-  placement: Placement = 'bottom',
-  strategy: PositioningStrategy = 'fixed',
+  placement: Placement | 'auto' = 'bottom',
```
thyhjwb6, line 18: Technically this is a breaking change because we export `useFloatingElement` at the root of the component library. We don't actually document the use of `useFloatingElement` but we don't document other hooks either and I know they are being used in downstream apps. I wonder if we need to still reference the old types and do some mapping, so the types for the function parameters can be changed in v5 of the Design System.

packages/react-component-library/src/hooks/useFloatingElement.ts
```diff
@@ -1,51 +1,120 @@
...
-          },
-        },
-      ],
-      placement,
-      strategy,
+  // Handle 'auto' placement (legacy Popper.js feature)
+  const isAutoPlacement = placement === 'auto'
+  const actualPlacement: Placement = isAutoPlacement ? 'bottom' : placement
+
+  // Build middleware array
```
thyhjwb6, line 41: Maybe we can remove this set of comments as they are also commented below.

packages/react-component-library/src/hooks/useFloatingElement.ts
```diff
@@ -1,51 +1,120 @@
...
+      reference: externalTargetElement || targetElement,
+      floating: floatingElement,
+    },
+    middleware,
+    whileElementsMounted: autoUpdate,
+  })
+
+  const arrowData = middlewareData.arrow
+
+  // Calculate arrow position based on Floating UI's coordinate system
```
thyhjwb6, line 87: I wonder if we could abstract a `calculateArrowPosition` function which would make this more self-documenting and negate the need for the comment.

packages/react-component-library/src/primitives/FloatingBox/FloatingBox.test.tsx
```diff
@@ -110,9 +110,11 @@ it('sets the position of the floating box', async () => {
...
 
   await waitFor(() => {
-    expect(screen.getByTestId('floating-box')).toHaveStyle({
-      transform: `translate(0px, ${offsetHeight}px)`,
-    })
+    // Floating UI with flip middleware may position the box above or below
+    // depending on available space. Just verify it has positioning applied.
+    const floatingBox = screen.getByTestId('floating-box')
+    const style = floatingBox.style.transform
+    expect(style).toContain('translate')
```
thyhjwb6, line 117: Can we be more specific with the assertion? Isn't the test positioning always the same?

packages/react-component-library/src/primitives/FloatingBox/FloatingBox.test.tsx
```diff
@@ -123,8 +125,10 @@ it('sets the position of the floating box when the ref is null', async () => {
   setup()
 
   await waitFor(() => {
-    expect(screen.getByTestId('floating-box')).toHaveStyle({
-      transform: `translate(0px, ${offsetHeight}px)`,
-    })
+    // Floating UI with flip middleware may position the box above or below
```
thyhjwb6, line 128: See comment about assertion.

thyhjwb6 (commented):
Code looks good, just a few comments.. maybe we need to chat about what is technically a breaking change.

`useFloatingElement` is used by `useContextMenu` which is used by the `ContextMenu` and that looks like this:

Screenshot 2025-11-26 at 09 35 35

Maybe we just need to make sure `FloatingBox` is working as expected.

PR #4030, opened 2025-11-26
fix(Popover): Fix popovers from staying open on fast mouse movements

ollie-sutton:
## Overview

Fix popovers from staying open on fast mouse movements

## Reason

Due to the way the mouse events where re-created ever time isVisible changed it caused the popover to sometimes miss events which meant you could get them staying open when they shouldn't

## Work carried out

- [x] Update useHideShow hook to handle mouse events by ref
- [x] Add useCallback to the popover target

## Developer notes

I think if we go for this change then we can park the FloatingUI change for now as per Lees comment on them currently being breaking.

I've added a Multi story for testing, happy to remove this before merging and it doesn't provide anything other than being able to reproduce the bug

### DIFF
--- a/packages/react-component-library/src/components/Popover/Popover.stories.tsx
+++ b/packages/react-component-library/src/components/Popover/Popover.stories.tsx
@@ -51,3 +51,48 @@ export const Open = Template.bind({})
 Open.args = {
   isVisible: true,
 }
+
+export const Multi: StoryFn<typeof Popover> = (args) => (
+  <div>
+    <Popover {...args}>
+      <div
+        css={`
+          display: inline-block;
+          padding: 1rem;
+          background-color: #c9c9c9;
+          margin-bottom: 10px;
+        `}
+      >
+        Hover on me
+      </div>
+    </Popover>
+    <br />
+    <Popover {...args}>
+      <div
+        css={`
+          display: inline-block;
+          padding: 1rem;
+          background-color: #c9c9c9;
+          margin-bottom: 10px;
+        `}
+      >
+        Hover on me
+      </div>
+    </Popover>
+    <br />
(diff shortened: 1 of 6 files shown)


packages/react-component-library/src/hooks/useHideShow.ts
```diff
@@ -49,23 +58,32 @@ export function useHideShow(
     }
 
     setIsVisible(true)
-  }
+  }, [])
 
   useDocumentClick(floatingBoxChildrenRef, hideElement, isClick)
 
-  function getMouseEvents() {
+  if (!mouseEventsRef.current) {
```
m7kvqbe1: I think I prefer this for readability but it's just a personal preference:

```typescript
if (!mouseEventsRef.current) {
  mouseEventsRef.current = isClick
    ? {
        onClick: () => (isVisibleRef.current ? hideElement() : showElement()),
      }
    : {
        onMouseEnter: showElement,
        onMouseLeave: hideElement,
      }
}
```

packages/react-component-library/src/components/Popover/Popover.stories.tsx
```diff
@@ -51,3 +51,48 @@ export const Open = Template.bind({})
 Open.args = {
   isVisible: true,
 }
+
+export const Multi: StoryFn<typeof Popover> = (args) => (
```
m7kvqbe1, line 55: I think lets leave the story given that this is a tricky one to get a test around.

Might be able to get a Playwright test around it but it could be flaky.

m7kvqbe1 (commented):
Looks good - I wonder if it's worth a quick stab at a Playwright test?

m7kvqbe1 (approved):
LGTM

PR #4031, opened 2025-12-11
feat(MultiSelect): Add multi select variant into the current select

ollie-sutton:
## Overview

Add a multi-select variant onto the current select component, maintaining backwards compatibility

## Reason

Required for DNADB-989

## Work carried out

- [x] Add a new isMulti prop into the Select to enable multi-select
- [x] Add new storybooks for the isMulti Select variant
- [x] Fix styling issues with the slight vertical offset on the checkbox
- [x] Decide on the order of icon -> checkbox -> text -> badge or checkbox -> icon -> text -> badge (I think option 2 is correct but isn't the current behaviour, see storybook preview)
- [x] Add unit tests for isMulti variant

### DIFF
--- a/packages/react-component-library/src/components/Select/Select.stories.tsx
+++ b/packages/react-component-library/src/components/Select/Select.stories.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useState } from 'react'
 import { Meta, StoryFn } from '@storybook/react'
 import {
   IconAgriculture,
@@ -36,6 +36,16 @@ const StyledWrapper = styled.div<{ $isDisabled?: boolean }>`
   max-width: 20rem;
 `
 
+const OPTIONS = [
+  { text: 'Apple', value: 'apple' },
+  { text: 'Banana', value: 'banana' },
+  { text: 'Cherry', value: 'cherry' },
+  { text: 'Date', value: 'date' },
+  { text: 'Elderberry', value: 'elderberry' },
+  { text: 'Fig', value: 'fig' },
+  { text: 'Grape', value: 'grape' },
+]
+
 const Template: StoryFn<typeof Select> = (args) => (
   <StyledWrapper $isDisabled={args.isDisabled}>
     <Select {...args}>
@@ -70,13 +80,13 @@ const TemplateWithIconsAndBadges: StoryFn<typeof Select> = (args) => (
         Two
       </SelectOption>
       <SelectOption badge={110} icon={<IconRemove />} value="two">
-        Two
-      </SelectOption>
-      <SelectOption badge={111} icon={<IconAgriculture />} value="three">
         Three
       </SelectOption>
-      <SelectOption badge={112} icon={<IconBrightnessAuto />} value="four">
+      <SelectOption badge={111} icon={<IconAgriculture />} value="three">
         Four
+      </SelectOption>
+      <SelectOption badge={112} icon={<IconBrightnessAuto />} value="four">
+        Five
       </SelectOption>
     </Select>
   </StyledWrapper>
@@ -124,14 +134,9 @@ const StyledBottomRightWrapper = styled.div`
 export const AutoPositioning: StoryFn<typeof Select> = (args) => (
   <StyledBottomRightWrapper>
     <Select {...args}>
-      <SelectOption value="one">Option One</SelectOption>
-      <SelectOption value="two">Option Two</SelectOption>
-      <SelectOption value="three">Option Three</SelectOption>
-      <SelectOption value="four">Option Four</SelectOption>
-      <SelectOption value="five">Option Five</SelectOption>
-      <SelectOption value="six">Option Six</SelectOption>
-      <SelectOption value="seven">Option Seven</SelectOption>
-      <SelectOption value="eight">Option Eight</SelectOption>
+      {OPTIONS.map((x) => (
+        <SelectOption value={x.value}>{x.text}</SelectOption>
+      ))}
     </Select>
   </StyledBottomRightWrapper>
 )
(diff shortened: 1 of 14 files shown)


packages/react-component-library/src/components/Select/SelectOption.tsx
```diff
@@ -1,14 +1,47 @@
...
 import { SelectBaseOption, SelectBaseOptionAsStringProps } from '../SelectBase'
+import { Checkbox } from '../Checkbox'
+import { CHECKBOX_RADIO_VARIANT } from '../CheckboxRadioBase'
+
+const StyledWrapper = styled.div`
+  display: flex;
+  align-items: center;
+  gap: 0.5rem;
+  width: 100%;
+`
```
ollie-sutton: Still have some styling issues around the check box and the text not being vertically aligned

packages/react-component-library/src/components/Select/Select.stories.tsx
```diff
@@ -147,3 +147,156 @@ AutoPositioning.parameters = {
...
+        Brightness
+      </SelectOption>
+      <SelectOption badge={112} icon={<IconAnchor />} value="navy">
+        Navy
+      </SelectOption>
+    </Select>
+  </StyledWrapper>
+)
+
+export const MultiSelectIconsAndBadges = MultiSelectWithIconsAndBadgesTemplate.bind({})
```
ollie-sutton: I think the way this shows is wrong, one to discuss

packages/react-component-library/src/components/Select/SelectOption.tsx
m7kvqbe1: Those checkboxes have always been a little bit quirky.

I think I had some styling issues with them in the DataGrid too.

packages/react-component-library/src/components/Select/Select.stories.tsx
m7kvqbe1: Yeah agreed, I think the Checkbox should probably always be furthest to the left.

packages/react-component-library/src/components/Select/Select.tsx
```diff
@@ -38,6 +53,112 @@ export const Select = ({
 
   const isControlled = value !== undefined
 
+  // Only used when isMulti is true
```
thyhjwb6: It could be good to move some of this to `useMultiSelect` or something like that. Maybe a few custom hooks would help.

packages/react-component-library/src/components/SelectBase/SelectBaseProps.ts
```diff
@@ -2,10 +2,10 @@ import { ComponentWithClass } from '../../common/ComponentWithClass'
 
 import { SelectChildrenType } from './types'
 
-export type OnChangeType = (value: string | null) => void
+export type OnChangeType<T> = (value: T) => void
 export type PopupPosition = 'above' | 'below' // | 'auto'
 
-export interface SelectBaseProps extends ComponentWithClass {
+export interface SelectBaseProps<T> extends ComponentWithClass {
```
thyhjwb6: This is technically a breaking change here as it's exported from the package.

packages/react-component-library/src/components/Select/Select.tsx
```diff
@@ -51,72 +172,151 @@ export const Select = ({
...
+                key: `select-option-${child.props.value}`,
+              }),
+              isHighlighted: highlightedIndex === index,
+              isSelected,
+              showCheckbox: isMulti,
+              title: child.props.children,
+            })
+          })}
+      </SelectLayout>
+      {/* Hidden span for measuring text width in multi-select mode */}
```
thyhjwb6: Maybe this is easiest but I am also wondering whether something like `ResizeObserver` is a cleaner solution.

packages/react-component-library/src/components/Select/Select.tsx
```diff
@@ -51,72 +172,151 @@ export const Select = ({
...
 
   const { onInputFocusHandler } = useMenuVisibility(isOpen, openMenu)
-
   const { onMenuKeyDownHandler } = useSelectMenu(inputRef)
 
-  const selectedItemText = isNil(selectedItem)
-    ? ''
-    : itemsMap[selectedItem].props.children
+  const handleClearButtonClick = useCallback(() => {
+    if (isMulti) {
```
thyhjwb6: It would be good to reduce the nesting here.

thyhjwb6 (commented):
Works well!

packages/react-component-library/src/components/Select/Select.tsx
```diff
@@ -38,6 +53,112 @@ export const Select = ({
...
+      const handler = onChange as SingleSelectProps['onChange']
+      handler?.(newValue)
+    },
+    [onChange]
+  )
+
+  const multipleSelection = useMultipleSelection({
+    selectedItems: isMulti ? currentSelectedItems : [],
+    onStateChange({ selectedItems: newSelectedItems, type }) {
+      switch (type) {
```
m7kvqbe1: How about something like this instead of these switch statements:

https://github.com/Royal-Navy/design-system/commit/c9815cb93d7820089e16e240e94186c598f1d1f0

packages/react-component-library/src/components/Select/Select.tsx
m7kvqbe1: Maybe we could extract some of this stuff into a custom hook to group the multi select related bits.

This might help to make the code more self documenting.

useMultiSelect?

packages/react-component-library/src/components/Select/Select.tsx
```diff
@@ -51,72 +172,151 @@ export const Select = ({
...
+      return
+    }
+
+    const fullText = getFullText(currentSelectedItems)
+    measureRef.current.textContent = fullText
+
+    const { offsetWidth: inputWidth } = inputRef.current
+    const { offsetWidth: textWidth } = measureRef.current
+
+    const buffer = 60 // Add some buffer for padding and clear button
```
m7kvqbe1: Maybe worth making this named const at top of file?

```typescript
const SOMETHING_MORE_DESCRIPTIVE = 60
```

packages/react-component-library/src/components/Select/Select.tsx
```diff
@@ -51,72 +172,151 @@ export const Select = ({
...
+      if (!isControlled) {
+        setSelectedItems([])
+      }
+      callMultiOnChange([])
+    } else {
+      reset()
+    }
+  }, [isMulti, isControlled, callMultiOnChange, reset])
+
+  const getFullText = useCallback(
```
m7kvqbe1: Maybe we also extract this display text stuff into a hook?

useDisplayText?

packages/react-component-library/src/components/SelectBase/partials/StyledCheckbox.tsx
```diff
@@ -0,0 +1,7 @@
+import styled from 'styled-components'
+import { Checkbox } from '../../Checkbox'
+
+export const StyledCheckbox = styled(Checkbox)`
```
m7kvqbe1, line 4: ❤️

m7kvqbe1 (commented):
Looks and behaves great! Just a few comments.

packages/react-component-library/src/components/SelectBase/SelectBaseProps.ts
ollie-sutton: https://github.com/Royal-Navy/design-system/pull/4031/changes/db284e994ea75b4729c7fa824f77b28bbbb60d68 should make it non-breaking

packages/react-component-library/src/components/Select/Select.tsx
ollie-sutton: Not sure if it is cleaner? Does mean we don't need the hidden span though

https://github.com/Royal-Navy/design-system/pull/4031/changes/fd670e53175d5954c5a322c5b19d20abaeacbff3

m7kvqbe1 (approved):

thyhjwb6 (approved):
👍

PR #4032, opened 2025-12-12
build(Release): Use OIDC trusted publishing for npm publish

m7kvqbe1:
## Overview

Updated the release workflow to use NPM's new trusted publishing with OIDC authentication instead of token-based authentication, improving security by using short-lived, automatically-managed credentials.

## Reason

NPM has introduced trusted publishing as the recommended authentication method for publishing packages from CI/CD environments. This approach is more secure than using long-lived tokens as it uses OIDC tokens that are automatically generated per workflow run and cannot be extracted or compromised.

## Work carried out

- [x] Added OIDC permissions (`id-token: write`) to the release job
- [x] Added step to update npm to latest version (required for trusted publishing support)
- [x] Removed `NODE_AUTH_TOKEN` secret usage from publish step
- [x] Configured trusted publishers on npmjs.com for all 5 packages

## Developer notes

- Trusted publishers have been configured on npmjs.com for each package
- The workflow will now authenticate using OIDC tokens generated by GitHub Actions
- Provenance attestations will be automatically generated for published packages
- The `NPM_AUTH_TOKEN` repository secret can be deleted once this is merged and verified working
- Reference: https://docs.npmjs.com/trusted-publishers

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,6 +9,9 @@ jobs:
   release:
     name: 'Release & Publish'
     runs-on: ubuntu-latest
+    permissions:
+      id-token: write  # Required for OIDC trusted publishing
+      contents: write  # Required for creating releases
     steps:
       - name: Checkout repo
         uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


ollie-sutton (approved):
LGTM

PR #4039, opened 2026-01-15
feat(IconLibrary): Add arrow upload progress icon

thyhjwb6:
## Related issue
DNADB-1000

## Overview
Adds the `arrow-upload-progress.svg` icon to the icon library.

## Reason
Required to show download progress in downstream application.

## Work carried out
- [x] Add icon

## Screensh
Screenshot 2026-01-15 at 13 32 54

## Developer notes
It's a Material Symbols icon, not classic Material Icons. The new icon uses viewBox="0 -960 960 960" which is the Material Symbols format, whereas the existing library uses classic Material Icons (viewBox="0 0 24 24"). A full migration to Material Symbols would require updating all ~1,400 existing icons and could introduce breaking changes.

Updating all icons carries increased risk. Adding a single icon avoids the risk of unintended changes to existing icons that consumers depend on, while a bulk update could alter SVG paths or remove deprecated icons.


m7kvqbe1 (dismissed):

thyhjwb6:
Not going to import the new Material Symbol as it uses a different viewbox. We will use an existing icon.

PR #4053, opened 2026-06-24
fix(Select): Fix focus ring when dropdown opens above

m7kvqbe1:
## Related issue

N/A

## Overview

Fixes the missing blue focus ring on Select (and Rows per page in DataGrid) when the dropdown opens above the trigger at the viewport edge.

## Reason

Vertical positioning for Select was added in #89eac17f6, but the above-position focus ring had a broken `::before` CSS block (missing closing brace), the trigger used below-only border-radius when open, and the DataGrid footer could paint under the dropdown menu.

## Work carried out

- [x] Fix malformed `getBoxShadow` CSS for above positioning in `StyledOptionsWrapper`
- [x] Apply position-aware border-radius on `StyledOuterWrapper` when open
- [x] Recalculate dropdown direction when the menu opens and on scroll (capture phase)
- [x] Raise DataGrid footer stacking context for full-width layouts
- [x] Allow focus ring overflow on select wrappers
- [x] Add regression tests for above-position box-shadow and border-radius

## How to test

1. Open a DataGrid with pagination (e.g. portal Sent Files / Transfer Queue)
2. Scroll so the footer sits at the bottom of the viewport
3. Open **Rows per page** — confirm the blue focus ring appears on all sides of the menu and trigger

### DIFF
--- a/packages/react-component-library/src/components/DataGrid/partials/StyledFooter.tsx
+++ b/packages/react-component-library/src/components/DataGrid/partials/StyledFooter.tsx
@@ -1,4 +1,4 @@
-import { spacing } from '@royalnavy/design-tokens'
+import { spacing, zIndex } from '@royalnavy/design-tokens'
 import styled, { css } from 'styled-components'
 
 export const StyledFooter = styled.div<{ $isFullWidth: boolean }>`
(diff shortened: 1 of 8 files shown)


ollie-sutton (approved):
LGTM

PR #4061, opened 2026-06-29
chore(Storybook): Upgrade Storybook from 8.3.4 to 10.4.6

m7kvqbe1:
## Related issue

N/A

## Overview

Upgrades Storybook from `8.3.4` to the latest `10.4.6` (two majors). Performed with the official `storybook upgrade` + `automigrate` codemods, then manual cleanup. Addons consolidated into core, deprecated standalone packages removed, and all story/mdx imports migrated. The static Storybook build passes locally on Node 24.

## Reason

Storybook 8 is two majors behind; moving to 10 keeps us on a supported line and unblocks current addon/tooling versions.

## Work carried out

- [x] Bumped all `@storybook/*` + `storybook` + `eslint-plugin-storybook` to `^10.4.6`; `@storybook/test-runner` to `^0.24.4`; `addon-webpack5-compiler-babel` to `^4`
- [x] Removed packages now built into core: `addon-actions`, `addon-essentials`, `addon-interactions`, `blocks`, `manager-api`, `theming`, `test`; added `@storybook/addon-docs`
- [x] Migrated imports across 68 story/mdx files: `@storybook/react`→`@storybook/react-webpack5`, `@storybook/blocks`→`@storybook/addon-docs/blocks`, `@storybook/addon-actions`→`storybook/actions`, `@storybook/test`→`storybook/test`
- [x] `.storybook` config: addon list trimmed, `manager.js` imports → `storybook/manager-api` + `storybook/theming`, `docs.autodocs` → `tags: ['autodocs']` in `preview.js`
- [x] Verified `storybook:static` build succeeds locally

## Developer notes

- The static build emits the same pre-existing bundle-size warnings as before (non-fatal).
- Local pre-commit lint-staged flags `.storybook/*.js` with "File ignored by default" (eslint ignores dot-dirs); this is a pre-existing quirk and **CI lint (`eslint src e2e`) is unaffected**.
- Stacks on top of the Node 24 runtime bump (already merged).


packages/react-component-library/src/intro/developer-guide.mdx
```diff
@@ -28,13 +28,13 @@ const rootElement = document.getElementById("root");
...
 
 function App() {
-  return (
-    <GlobalStyleProvider>
-      <Button variant="primary">
-        Hello, World!
-      </Button>
-    </GlobalStyleProvider>
-  )
+return (
```
thyhjwb6, line 31: Seems odd the indentation was lost here.

packages/react-component-library/src/tokens/spacing.mdx
```diff
@@ -73,22 +88,21 @@ export const ExampleComponent = () => {
   const paddingSize = spacing('8')
   const marginSize = spacing('2')
 
-  return (
-    <div
-      style={{
+return (
```
thyhjwb6, line 91: And here.

thyhjwb6 (approved):

packages/react-component-library/src/intro/developer-guide.mdx
m7kvqbe1: Good spot. This is Prettier reformatting the MDX during the upgrade rather than a hand-edit — Prettier 2.8.8 strips the indentation inside the `code={`…`}` template literal and the flattened form is what it considers canonical (re-indenting just gets reverted and fails the Prettier check). It only affects how the snippet displays, so leaving as-is for now. Worth a follow-up to either wrap these blocks in `prettier-ignore` or move the snippets into real source files and raw-import them.

packages/react-component-library/src/tokens/spacing.mdx
m7kvqbe1: Same root cause as the developer-guide thread — Prettier flattens the indentation inside the `<Source>` template literal and that's its canonical output, so leaving as-is. Captured as a follow-up.

